PR #114, opened 2014-12-22
AS6701_32x remove unavailable usb node from DTS

david56:
The other commit is used to revise 'onie_sz.b' to speed up booting into ONIE.  This patch is for all powerpc platforms except AS4600_54T, AS5600_52X, and AS5610_52X.


cbrune:
Pushed to master:

- ce910d2 Accton platforms: Revise 'onie_sz.b' to 0x500000
- 99b28aa Accton AS6701_32X: Remove unavailable usb node from DTS

PR #115, opened 2015-01-05
Accton AS7700_32X: Remove autoboot prompt from u-boot

david56:
Remove "<123ENTER>" autoboot prompt from u-boot.


cbrune:
Pushed to master:
- 89846d0 Accton AS7700_32X: Remove autoboot prompt from u-boot

PR #116, opened 2015-01-05
Accton platforms: Revise eUSB discovery method for updater

david56:
The platforms are including:
- as5712_54x
- as6712_32x
- as7702_32x


cbrune:
Pushed to master:
- b458c5a Accton platforms: Revise eUSB discovery method for updater

PR #117, opened 2015-01-06
kernel 3.14.x: enable a few Device Mapper kernel options

cbrune:
Enable the following options:

CONFIG_DM_THIN_PROVISIONING

  Provides thin provisioning and snapshots that share a data store.

CONFIG_DM_VERITY

  This device-mapper target creates a read-only device that
  transparently validates the data on one underlying device against a
  pre-generated tree of cryptographic checksums stored on a second
  device.

When the initial LVM support went in these two rather common options
were missed.


cbrune:
pushed to master:

- 5a43b7c kernel 3.14.x: enable a few Device Mapper kernel options

PR #118, opened 2015-01-12
Accton AS6710_32X: Remove autoboot prompt from u-boot

david56:
Remove "123&lt;ENTER&gt;" autoboot prompt from u-boot


cbrune:
pushed to master:

- ee708f6 Accton AS6710_32X: Remove autoboot prompt from u-boot

PR #119, opened 2015-01-15
Erase mass storage in uninstall mode

david56:


cbrune:
What problem are you fixing here?  On powerpc the block device to wipe out should be in the device tree as `/proc/device-tree/mass_storage/device`.  Is that method not working?

david56:
The onboard mass storage device detected by kernel will be either ``sda`` or ``sdb`` if an USB disk is plugged into front USB port and power on.  The ``/proc/device-tree/mass_storage/device`` specifies fixed device name.  It may cause the plugged USB disk to be erased.  So that we exploit additional script to discover the correct mass storage device.

david56:
A quick question: Erasing mass storage in ONIE uninstall mode, whether time is a concern, or full erase is a concern, or "don't write flash too many times" is a concern?

cbrune:
OK, I understand what you are trying to do now.

Regarding your concerns:  At this stage of the project I do not think any of those things are a concern.  ONIE needs a way to "erase" the mass storage device and return the system to a clean state.

It is fine if it takes some time.
Full erase is definitely desired.

Writing the flash too many times is a concern, but there is nothing ONIE can do about that.  That is really more of an issue for automated tests that burn through flash.

For users who want to quickly re-install, they can just go back to installer mode, without using uninstall.

Uninstall is really for the case where some one really wants to *remove* all the software.   For example you bought a switch, used it for a while and then decided to sell it on eBay.  Before selling it you want to completely remove all your data.

cbrune:
pushed to master:

- 6c47e68 Accton platforms: Erase onboard mass storage in uninstall mode
- a445b08 uninstall-arch(powerpc): create standalone function to erase mass storage

PR #120, opened 2015-01-16
Accton AS5710_54X: Correct USB compatible nodes in DTS

david56:


cbrune:
pushed to master:

- 84f45c3 Accton AS5710_54X: Correct USB compatible nodes in DTS

PR #122, opened 2015-01-20
Accton AS5700_96X: Add to support ONIE

david56:


cbrune:
Looks good, pushed to master:

- 0f8b6cb Accton AS5700_96X: Add to support ONIE

PR #123, opened 2015-01-20
Ensure syslinux is installed

cobracmder:
Need to ensure ```syslinux``` is installed as part of
```debian-prepare-build-host``` as it provides ```isohybrid``` and a dep
package ```syslinux-common``` provides ```isohybrid.pl``` which is used
by ```recovery-iso``` build targets.

### DIFF
--- a/build-config/Makefile
+++ b/build-config/Makefile
@@ -374,6 +374,7 @@ machine-prefix:
 # Install required build packages for a x86_64 debian based build host
 DEBIAN_BUILD_HOST_PACKAGES	= build-essential stgit u-boot-tools python-sphinx rst2pdf \
 				  gperf device-tree-compiler python-all-dev genisoimage \
+				  syslinux \
 				  autoconf automake bison flex texinfo libtool \
 				  realpath gawk libncurses5 libncurses5-dev bc
 


cbrune:
Looks good, pushed to master:

- a87fd1e Ensure syslinux is installed

mandeepsandhu:
Hey Guys, I'm late to the party...but why are we installing `syslinux` package? We're already building it from source, so I'm wondering why the package is needed?  Is it because of `isohybrid.pl`? This is already present in the `syslinux` build dir (`syslinux/syslinux-4.07/utils/isohybrid.pl`).

If I remember correctly, we decided NOT to use the packages in the repositories and have our own fixed version built from source.

cbrune:
You are right!  I remember that now.  Something else is going wrong.

I need to revert a87fd1e.

cbrune:
We probably have been skating along using the isohybrid.pl from the host (unknowingly).

When we went to building syslinux ourselves we also needed to update the makefile to use the isohybrid.pl from the syslinux source package.

cobracmder:
My bad.  Running a quick build with a patch to images.make that qualifies the path to ```isohybrid.pl``` to be from ```$(SYSLINUX_DIR)/utils```.

cbrune:
Ha, my bad also.  I was too quick to upstream it without thinking.  Thanks Mandeep!

cobracmder:
Created new PR #124 to uses the locally build ```syslinux``` vs one on the host.  closing.

mandeepsandhu:
Here's my contribution to *My bad*! I should've updated the original commit to use `isohybrid.pl` from the source build (I didn't catch it then because the `syslinux` pkg was still installed on m y machine and the kvm build was using that)!

I think we're all even now! :)

PR #130, opened 2015-02-14
discover: check both boot_file and bootfile variables

cbrune:
Ticket: 126

The problem is in ONIE and the difference between the BOOTP "file"
field and DHCP option 67 "Boot file".

For bootp reference:
https://tools.ietf.org/html/rfc951

For DHCP option 67 reference:
https://tools.ietf.org/html/rfc2132

Busybox's udhcpd translates the BOOTP "file" field into a variable
named "boot_file", which the discovery process prepends with
"onie_disco_" to finally have: onie_disco_boot_file. You can see this
variable set in the log file. Also the PCAP shows the "Boot file name"
field in the BOOTP section.

DHCP option 67 is translated into "bootfile" (without the underscore),
which gives the ONIE variable "onie_disco_bootfile".

The problem is the ONIE download code is only using
$onie_disco_bootfile. ONIE should be checking both
$onie_disco_bootfile and $onie_disco_boot_file.

In the "exact TFTP" download section this patch now loops over both
variables looking for a hit.

### DIFF
--- a/rootconf/default/bin/exec_installer
+++ b/rootconf/default/bin/exec_installer
@@ -250,21 +250,26 @@ http_download()
 
 tftp_download()
 {
-    # Try BOOTP next-server and bootfile
-    if [ -n "$onie_disco_siaddr" ] && [ -n "$onie_disco_bootfile" ] ; then
-        url_run "tftp://$onie_disco_siaddr/$onie_disco_bootfile" && return 0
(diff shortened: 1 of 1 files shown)


cbrune:
pushed to master:

- 327ecca discover: check both boot_file and bootfile variables

Issue #132, opened 2015-02-17
testing: test cases  11 & 47 (DHCP Exact - TFTP Server name + TFTP Bootfile) fail


cobracmder:
```
./test-onie.py -i 192.168.1.100/24 -I eth1 -m 7072cfaa34fa -t 47 -o onie_updater onie-updater-powerpc-accton_as4600_54t-r0.bin

2015-02-15T16:10:36 - INFO - Loaded 80 tests from config/onie-tests.json
2015-02-15T16:10:36 - INFO - Configuring for Test 47 - ONIE Update - DHCP Exact - TFTP Server name + TFTP Bootfile
Traceback (most recent call last):
  File "./test-onie.py", line 685, in <module>
    main()
  File "./test-onie.py", line 653, in main
    test_args = configure_test(args)
  File "./test-onie.py", line 472, in configure_test
    prepare_test_case(test_args)
  File "./test-onie.py", line 377, in prepare_test_case
    hostname = test_args['local_name']
KeyError: 'local_name'
```

cobracmder:
Looks like a simple typo ```local_name``` vs ```host_local_name```.

cbrune:
closed by 702c194 and 396d4b8

Issue #129, opened 2015-02-13
Testing: test cases that use <vendor> incorrectly use vendor_id vs vendor name


cobracmder:
```test-onie.py``` expects a numeric value for onie_vendor (in this case, 259), however when test-onie generates the name for the installer file, it inserts this value into the name. Therefore, we end up with a file named ```onie/contrib/autobuild/output/test-13/www-root/onie-installer-powerpc-259_accton_as4600_54t-r0```, which is not found by ONIE in discovery mode

cbrune:
closed by 702c194 and 396d4b8

PR #135, opened 2015-02-25
vendor should be string, not int & local_name should be host_local_name

ssilvestro:
vendor should be string, not int & local_name should be host_local_name

Related to issues 129 and 132:
https://github.com/opencomputeproject/onie/issues/129
https://github.com/opencomputeproject/onie/issues/132


cobracmder:
nice catch Sam.

cbrune:
pushed to master:

- 396d4b8 vendor should be string, not int & local_name should be host_local_name

PR #136, opened 2015-02-25
vendor should be string, not int

ssilvestro:
Helps resolve issue 129: https://github.com/opencomputeproject/onie/issues/129


cobracmder:
this looks good

cbrune:
push to master:

- 702c194 vendor should be string, not int

PR #143, opened 2015-03-02
ONIE updater be able to check machine image via install-platform

david56:
As discussed in the thread:
http://lists.opencompute.org/pipermail/opencompute-onie/2015-February/000414.html

Add install-platform for machine vendor to override the behaviour of checking machine image.

### DIFF
--- a/build-config/scripts/onie-mk-installer.sh
+++ b/build-config/scripts/onie-mk-installer.sh
@@ -86,6 +86,10 @@ cp $installer_dir/install.sh $tmp_installdir || exit 1
 echo -n "."
 cp -r $installer_dir/$arch/* $tmp_installdir
 
+[ -r $machine_dir/installer/install-platform ] && {
+    cp $machine_dir/installer/install-platform $tmp_installdir
+}
+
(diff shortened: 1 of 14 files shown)


cbrune:
All in all this looks really good.  I would just make the one change noted above.

david56:
> Perhaps leave the $fail check and error printing outside the function. In other words just have check_machine_image() set fail=yes (or not). Then outside the function test $fail and $force and print the error message.
>
> Looking at your other commit for the Accton specifics it looks like this would eliminate some redundant code.
>
> That way everyone can share the same error printing. Of course a custom function could also print additional error messages if necessary.

Updated.  I thought somebody would like to skip the check at some situations before.  But it is not reasonable.  ONIE should always check the target of machine.  :smiley:

cbrune:

- 167fcbf Accton platforms: be able to install on machines running ONIE v1.6
- ef94abe ONIE updater: be able to override default behaviour via install-platform
- 281cfa0 ONIE updater: move checking machine image to standalone function

PR #144, opened 2015-03-04
Added support for mmcblk devices

Karel-van-de-Plassche:
I have tested this on a device which has an eMMC, please test this on a device with a regular disk. I've checked the strings and all should be fine.

### DIFF
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -13,7 +13,7 @@ cd $(dirname $0)
 echo "Demo Installer: platform: $platform"
 
 # Install demo on same block device as ONIE
-blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' | sed -e 's/[0-9]://' | head -n 1)
+blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' |  sed -e 's/[0-9]:$//' | sed -e 's/\([0-9]\)\(p\)/\1/' | head -n 1)
 
 [ -b "$blk_dev" ] || {
     echo "Error: Unable to determine block device of ONIE install"
@@ -126,7 +126,11 @@ create_demo_msdos_partition()
 }
 
 eval $create_demo_partition $blk_dev
-demo_dev="${blk_dev}$demo_part"
+if [[${onie_dev} == sd*]]; then
+    demo_dev="${blk_dev}$demo_part"
+else
+    demo_dev="${blk_dev}p$demo_part"
+fi
 partprobe
 
 # Create filesystem on demo partition with a label
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -126,11 +126,7 @@ create_demo_msdos_partition()
 }
 
 eval $create_demo_partition $blk_dev
-if [[${onie_dev} == sd*]]; then
-    demo_dev="${blk_dev}$demo_part"
-else
-    demo_dev="${blk_dev}p$demo_part"
-fi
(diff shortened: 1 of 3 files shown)


demo/installer/x86_64/install.sh
```diff
@@ -126,7 +126,11 @@ create_demo_msdos_partition()
 }
 
 eval $create_demo_partition $blk_dev
-demo_dev="${blk_dev}$demo_part"
+if [[${onie_dev} == sd*]]; then
+    demo_dev="${blk_dev}$demo_part"
+else
+    demo_dev="${blk_dev}p$demo_part"
+fi
```
cbrune: I think this test will be better if you check explicitly for "mmcblk" in the device name.  It is the only blk device type that requires special handling.

As it is written right now this will have a problem on the virtual machine implementation, whose blk device looks like /dev/vda.

Also all the ONIE scripts are written for busybox's ASH shell, not BASH.  So please do not use BASH-isms like "[[" and "==".

ASH is very nearly identical to Debian's (and Ubuntu's) default system interpreter "dash" shell.  See the man page dash(1) -- http://manpages.ubuntu.com/manpages/lucid/man1/sh.1.html

Karel-van-de-Plassche: I changed it to be more in line with the other tests. Also, mmcblk devices are the only 'special' devices now

Karel-van-de-Plassche:
Commits are getting a bit messy. I could resend a pull request if you'd like.

cbrune:
Yes, you need to make a clean PR.

Also there are a number of shell script errors.  I tried typing one of your constructs into the ONIE shell to see and it errors out:
```
ONIE:/ # onie_dev=/dev/mmcblk0
ONIE:/ # if [${onie_dev} = mmcblk*]; then
> echo "yes"
> fi
/bin/sh: [/dev/mmcblk0: not found
```

I would stick to brain dead simple logic.  For example make a test early on to see if the blk device matches "mmcblk".  Something simple like:

```shell
blk_suffix=
echo $onie_dev | grep -q mmcblk && blk_suffix="p"
```
Then just use $blk_suffix where ever you need to.  Or perhaps you can get away with just appending $blk_suffix to $onie_dev right away:
```shell
onie_dev="${onie_dev}$blk_suffix"
```

Additionally a few shell scripting tips:

1. when using [ ]'s for tests leave a space before and after each brace.  easier to read.

2. you need to be careful when concatenating shell variables with strings.  Whenever you want to add a string to a variable be sure to use double-quotes and curly braces.  For example to append "xxx" to $old_var do this:
```shell
new_var="${old_var}xxx"
```
not this, new_var will be blank in this case:
```shell
new_var=$old_varxxx
```

example:
```shell
ONIE:/ # old_var=yes
ONIE:/ # new_var=$old_varxxx
ONIE:/ # echo $new_var

ONIE:/ # new_var="${old_var}xxx"
ONIE:/ # echo $new_var
yesxxx
```

Karel-van-de-Plassche:
Sorry, hadn't coded in ash before. I thought I tested this, but apparently not. I'll make a new pull request tomorrow. Quick question, why

```ash
onie_dev="${onie_dev}$blk_suffix"
```

and not 

```ash
onie_dev="${onie_dev}${blk_suffix}"
```
Or doesn't it matter?

cbrune:
for that either one will work.  i am just in the habit of using {} braces on the left-most variables when concatenating as the string is parsed left to right.

Thanks for sticking with it!

Karel-van-de-Plassche:
New pull request created on https://github.com/opencomputeproject/onie/pull/153. This one can be closed.

cbrune:
closing this one as #153 supersedes it.

PR #146, opened 2015-03-05
Accton AS5700_96X: enlarge diag partition size to 24Mb

david56:


cbrune:
This looks fine.  Pushed to master:

- 676b203 Accton AS5700_96X: enlarge diag partition size to 24Mb

PR #148, opened 2015-03-09
x86 uninstall: fix shell script issue in commit 2684576

david56:
In commit 2684576, the expression enclosed by single quotation
marks is always taken as a normal string.  So that test of zero
length of string never occurs.  The patch revised to use double
quotation marks to enclose the expression.


cbrune:
pushed to master:

- 3409979 x86 uninstall: fix shell script issue in commit 2684576

PR #149, opened 2015-03-09
Quanta LB9: Fixup MPC8541 SS_EN and CLK_ADJ

QuantaSwitchONIE:


machine/quanta/quanta_lb9/u-boot/Fixup-MPC8541-SS_EN-and-CLK_ADJ.patch
```diff
@@ -0,0 +1,52 @@
...
+ 	unsigned int clk_adjust;	/* Clock adjust */
+ 
+ 	clk_adjust = popts->clk_adjust;
++#if defined(CONFIG_MPC8541) || defined(CONFIG_MPC8555)
++	ddr->ddr_sdram_clk_cntl = (0x80000000 | ((clk_adjust & 0x7) << 24));
++#else
+ 	ddr->ddr_sdram_clk_cntl = (clk_adjust & 0xF) << 23;
++#endif
+ 	debug("FSLDDR: clk_cntl = 0x%08x\n", ddr->ddr_sdram_clk_cntl);
+ }
```
cbrune: This looks good, but this should be two separate patches.

The SS_EN fix is generic to all MPC8541 and MPC8555, while the below is specific to the LB9.

The SS_EN fix should be a core ONIE patch with the prefix "arch-powerpc-XXX" in this directory https://github.com/opencomputeproject/onie/tree/master/patches/u-boot.

QuantaSwitchONIE: I was wondering if SS_EN patch to a global core ONIE, impacts to all user who are using MPC8555/8541, they need to modify correct clk_adjust as 24 bits shift, otherwise DDR config may cause unknow problem.

cbrune:
The LB9 is the only machine in the ONIE repo using 8541/8555.  So it will be OK to make the global change.

PR #152, opened 2015-03-12
Dropbear 2015.67

nshopik:
Update dropbear to latest version. And fixes CVE-2013-4421, CVE-2013-4434 (memory exhaustion denial-of-service and avoids disclosing which users are valid.)

### DIFF
--- a/build-config/conf/dropbear.config.h
+++ b/build-config/conf/dropbear.config.h
@@ -2,13 +2,13 @@
  * Copyright (c) 2002,2003 Matt Johnston
  * All rights reserved. See LICENSE for the license. */
 
-#ifndef _OPTIONS_H_
-#define _OPTIONS_H_
+#ifndef DROPBEAR_OPTIONS_H_
+#define DROPBEAR_OPTIONS_H_
 
-/******************************************************************
- * Define compile-time options below - the "#ifndef DROPBEAR_XXX .... #endif"
- * parts are to allow for commandline -DDROPBEAR_XXX options etc.
- ******************************************************************/
+/* Define compile-time options below - the "#ifndef DROPBEAR_XXX .... #endif"
+ * parts are to allow for commandline -DDROPBEAR_XXX options etc. */
+
+/* IMPORTANT: Many options will require "make clean" after changes */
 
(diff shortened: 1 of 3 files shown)


cbrune:
This all seems reasonable.

I am curious about a couple of things though.

1.  I noticed the config file changes indicate the size of the executable will increase.  Do you have numbers for how much the final ONIE images changed?  Both x86 and PowerpC would be nice.  A few  older powerpc platforms are particularly sensitive to the image size.

2. How did you test it?  Things to verify for powerpc and x86 (can use VM):
a) the server starts OK
b) first time boot it generates keys and stores them
c) 2nd and beyond time boots it uses the previously generated keys
d) you can ssh to the box

3. The info you put in the PR comments about the CVE and such would be nice to put into the commit message.

nshopik:
1. I'll check binary sizes (full image difference was about 50kb but I double check), only compiled for powerpc (quanta) and tested on it.
2. I only didn't test key generation, removing env should be enough to force re-generation? every other test passed.
3. Shall I prepare another PR to fix commit message?

cbrune:
On powerpc you can zap these two env variables to regenerate keys:
- onie_dropbear_dss_host_key
- onie_dropbear_rsa_host_key

It would be nice to fix up the commit message.

nshopik:
- Difference is exactly 20480 bytes with new dropbear
- Regeneration works correctly

Currently no ECC keys are generated

nshopik:
Also we may disable option DROPBEAR_DELAY_HOSTKEY, unless we plan to enable -R option and change way we generate keys.
```
/* Generate hostkeys as-needed when the first connection using that key type occurs.
   This avoids the need to otherwise run "dropbearkey" and avoids some problems
   with badly seeded /dev/urandom when systems first boot.
   This also requires a runtime flag "-R". This adds ~4kB to binary size (or hardly 
   anything if dropbearkey is linked in a "dropbearmulti" binary) */
#define DROPBEAR_DELAY_HOSTKEY
```

cbrune:
Thanks for the info about the size.  It all looks reasonable.

PR #153, opened 2015-03-12
Added support for mmcblk devices

Karel-van-de-Plassche:
Cleaned up and improved version of pull request https://github.com/opencomputeproject/onie/pull/144
Tested for a mmcblk device. Scripts checked, but not tested for sd devices

### DIFF
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -13,7 +13,7 @@ cd $(dirname $0)
 echo "Demo Installer: platform: $platform"
 
 # Install demo on same block device as ONIE
-blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' | sed -e 's/[0-9]://' | head -n 1)
+blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' |  sed -e 's/[0-9]:$//' | sed -e 's/\([0-9]\)\(p\)/\1/' | head -n 1)
 
 [ -b "$blk_dev" ] || {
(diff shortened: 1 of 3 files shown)


demo/installer/x86_64/install.sh
```diff
@@ -13,7 +13,7 @@ cd $(dirname $0)
 echo "Demo Installer: platform: $platform"
 
 # Install demo on same block device as ONIE
-blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' | sed -e 's/[0-9]://' | head -n 1)
+blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' |  sed -e 's/[0-9]:$//' | sed -e 's/\([0-9]\)\(p\)/\1/' | head -n 1)
```
cbrune: yak shave -- I don't think you need the "$" in the first sed as the ":" already anchors the search.  no big deal though.

also you do not need to pipe sed into another sed.  instead you can specify multiple "-e" expressions, which are applied one after another in order.  again, no big deal.

do not change anything.

cbrune:
This is cool, thanks for working it out.

How did you test this change?

At a minimum I expect you tried it on your board, verifying the mmc changes.  You should also use the x86 VM to check the other case.

For the VM the device is /dev/vda, which will cover the real machines that mostly use /dev/sda.

Karel-van-de-Plassche:
I tested it on our board and today I've tested it on a virtual device. It worked on both.

cbrune:
Great!  Thanks for all your help. Pushed to master:

- cb12986 Added support for mmcblk devices

PR #156, opened 2015-03-13
Quanta LB9: Fixup MPC8541 SS_EN and CLK_ADJ

QuantaSwitchONIE:
Separate into 2 patches

### DIFF
--- a/patches/u-boot/series
+++ b/patches/u-boot/series
@@ -13,3 +13,4 @@ arch-powerpc-85xx-always-call-board_reset.patch
 driver-support-new-broadcom-phys.patch
 platform-onie-common-env.patch
 platform-common-env.patch
+arch-powerpc-fix-mpc8541-ss_en_bit.patch
--- a/machine/quanta/quanta_lb9/u-boot/correct-clk_adjust-value-for-mpc8541.patch
+++ b/machine/quanta/quanta_lb9/u-boot/correct-clk_adjust-value-for-mpc8541.patch
@@ -1,0 +1,35 @@
+From d0fcaf237405fa2d1576b8455b30c3bb680f64d7 Mon Sep 17 00:00:00 2001
+From: Luffy <luffy.cheng@quantatw.com>
+Date: Fri, 13 Mar 2015 09:53:53 +0800
+Subject: [PATCH 2/2] Correct clk_adjust value for MPC8541
+
+---
+ board/quanta/quanta_lb9/ddr.c |   11 ++++++++---
+ 1 files changed, 8 insertions(+), 3 deletions(-)
+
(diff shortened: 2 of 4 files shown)


cbrune:
OK this is closer, but it needs to be broken up into *two* commits.  Right now it is still a single commit.

Please review the ONIE developer's guide about patches:
https://opencomputeproject.github.io/onie/docs/developers/contributing.html#general-patching-philosophy

In particular the part about "Each patch should only contain one logical change."

The global SS_EN for MPC8541 should be one commit by itself.
The LB9 changes should be another commit.

patches/u-boot/series
```diff
@@ -13,3 +13,4 @@ arch-powerpc-85xx-always-call-board_reset.patch
 driver-support-new-broadcom-phys.patch
 platform-onie-common-env.patch
 platform-common-env.patch
+arch-powerpc-fix-mpc8541-ss_en_bit.patch
```
cbrune: For the global series file I try to keep it organized by putting all the arch-XXX changes together in one place, with the ONIE specific changes at the end.  For example for your change the series file should look like:
```
git-ignore.patch
feature-config-repeatable.patch
feature-dhcp-options.patch
feature-populate-serial-number.patch
feature-sys-eeprom-tlv.patch
arch-powerpc-define-85xx-sdhc-polarity-select.patch
arch-powerpc-fix-p2020-i2c-clock.patch
arch-powerpc-fix-fsl-ddram-cfg.patch
arch-powerpc-85xx-always-call-board_reset.patch
arch-powerpc-fix-mpc8541-ss_en_bit.patch    <---- new patch here
driver-support-new-broadcom-phys.patch
platform-onie-common-env.patch
platform-common-env.patch
```

nshopik:
As LB9 owner, can somebody explain me what this patch actually do? First one do something about cycles and second one about DDR timing?

What actual problem this fixes?

As we previously had problem with original patch from Quanta, just like avoid this in future.

Thanks

QuantaSwitchONIE:
The actually do for this patch is to fix the missing code in recent u-boot for MPC8541/8555 , SS_EN and CLK_ADJST bits in DDR_SDRAM_CLK_CNLT register.
for detail, Curt has already made a good description here :
http://lists.denx.de/pipermail/u-boot/2015-February/204903.html

and for lb9 ddr.c is based on clk_adjst bit shift change from 23 to 24, it should be set to 011, it is same value as original code. The DDR_SDRAM_CLK_CNTL register value is changed  from 0x0300000 to 0x83000000. 
So the actually changed is SS_EN.

cbrune:
@QuantaSwitchONIE bump.  Looking for two commits.

PR #157, opened 2015-03-13
Alphanetworks SNH60A0-320F: Add to support ONIE

willkuo103:


cbrune:
Looks good, pushed to master:

- 1e7f24c Alphanetworks SNH60A0-320F: Add to support ONIE

PR #159, opened 2015-03-13
Dell Networking S4000 switch.

lakesh-sharma:
This patch supports adding S4000 platform related changes to ONIE ver 2015.02.
S4000 would use Broadcom Trident2 as switch ASIC. It has 48x10G SFP+ and 6x40G QSFP ports.
CPU module is C2338 Atom (Rangeley) based.


lakesh-sharma:
This patch supports adding S4000 platform related changes to ONIE ver 2015.02.
S4000 would use Broadcom Trident2 as switch ASIC. It has 48x10G SFP+ and 6x40G QSFP ports.
CPU module is C2338 Atom (Rangeley) based.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
```diff
@@ -0,0 +1,255 @@
...
+     case TLV_CODE_MAC_SIZE:
+ 	sprintf(value, "%u", (tlv->value[0] << 8) | tlv->value[1]);
+ 	break;
+-    case TLV_CODE_VENDOR_EXT:
+-	value[0] = 0;
+-	for (i = 0; (i < (TLV_DECODE_VALUE_MAX_LEN/5)) && (i < tlv->length);
+-	     i++) {
+-	    sprintf(value, "%s 0x%02X", value, tlv->value[i]);
+-	}
+-	break;
```
cbrune: You can't redefine the meaning of the VENDOR_EXT TLV.  It is defined here:

https://opencomputeproject.github.io/onie/docs/design-spec/switch_hw_requirements.html#definition-of-the-tlvinfo-eeprom-format

The reason you can't re-define it's meaning is because then *every* NOS will have to do something special for your platform to decode the VENDOR_EXT.  The existing NOS's have already implemented the ONIE TLV EEPROM format decoder and would have to change *only* for your platform.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
```diff
@@ -0,0 +1,255 @@
...
++
+ /*
+  *  This macro defines the sys_eeprom command line command.
+  */
+@@ -12,10 +15,14 @@ cmd_usage()
+ {
+     static const char *usage =
+ 	"Display and program the system EEPROM data block.\n"
+-	"Usage: sys-eeprom [-h][-l] [-e] [-s <code>=<value>,...]\n"
++	"Usage: sys-eeprom [-h][-l] [-e] [-b bus] [-a address] [-s <code>=<value>,...]\n"
```
cbrune: What is this for?

Does this mean you have multiple ONIE TLV compliant EEPROMs on the board?

That in and of itself does not really present a problem.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
```diff
@@ -0,0 +1,255 @@
...
+-    const char *short_options = "hels:g:";
++    const char *short_options = "hels:g:a:b:";
+     const struct option long_options[] = {
+ 	{"help",    no_argument,          0,    'h'},
+ 	{"list",    no_argument,          0,    'l'},
+ 	{"erase",   no_argument,          0,    'e'},
+ 	{"set",     required_argument,    0,    's'},
+ 	{"get",     required_argument,    0,    'g'},
++	{"addr",    required_argument,    0,    'a'},
++	{"bus",     required_argument,    0,    'b'},
```
cbrune: Making these arguments required is a problem.  Generic NOS installers would be required to know that for *this* platform they need to supply the -b and -a arguments.  That is kind of broken.

If you must use these arguments then I suggest they be optional.  By default they have "good" values that select the primary board EEPROM.

machine/dell/dell_s4000_c2338/installer.conf
```diff
@@ -0,0 +1,51 @@
...
+                device="/dev/$(basename $d)"
+                echo $device
+                return 0
+            elif [ "$path" = "$dell_s4000_c2338_disk2" ] ; then
+                device="/dev/$(basename $d)"
+                echo $device
+                return 0
+            fi
+        fi
+    done
```
cbrune: Does this mean you plan to ship two different SKUs for the S4000?

Will the EEPROM indicate the different SKUs, i.e. a different model number?

machine/dell/dell_s4000_c2338/machine.make
```diff
@@ -0,0 +1,34 @@
...
+VENDOR_ID = 674
+
+# This platform requires the PXE_EFI64 installer
+PXE_EFI64_ENABLE = yes
+
+CONSOLE_SPEED = 115200
+CONSOLE_DEV = 0
+CONSOLE_FLAG = 0
+
+EXTRA_CMDLINE_LINUX = i2c-ismt.bus_speed=100
```
cbrune: Is this really required?

Will all NOSs also have to set the speed to 100 ?

machine/dell/dell_s4000_c2338/rootconf/sysroot-bin/onie-diag
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+
+if [ -x $diag_file ] ; then
+echo "Platform specific diag found launching....."
+. $diag_file
+else
+echo "Platform specific diag not found..."
+fi
+exit 0
```
cbrune: Cool for adding a diag! Nice!

machine/dell/dell_s4000_c2338/machine.make
lakesh-sharma: We had been using speed as 100 in almost all recent platforms, atleast on S6000, it was 80.
It is not forced unless NOS does not see any issue.

machine/dell/dell_s4000_c2338/installer.conf
lakesh-sharma: S4000 has the capability to host 2 SATAs, but the production version would have just one. We have systems inhouse, which have 2 SATAs.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: these are optional argument. If user does not provide, the default values are set to point to system eeprom. This feature was added to allow onie-syseeprom to access any eeprom on board as requested by MFG.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: yes, it allows accessing other eeprom on boards (fan/psu), which can be of onie-tlv format.

machine/dell/dell_s4000_c2338/installer.conf
cbrune: OK that sounds fine.

machine/dell/dell_s4000_c2338/machine.make
cbrune: You are the hardware vendor.  You are saying there is a technical reason to use 100 on S4000 or are you saying you are just using 100 because that is what DELL does elsewhere?  I'm trying to understand the technical reason.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
cbrune: Nice.  I think the -a and -b is a cool feature we should add to the *core* of the EEPROM code, as it could be useful for others as well.  It would be straight forward to split off this feature into it a stand alone "feature patch" and have it not be DELL S4000 specific.

Now I have a bigger problem, one that predates your participation in the project.  Please see these threads from the mailing list:

http://lists.opencompute.org/pipermail/opencompute-onie/2014-September/000103.html
http://lists.opencompute.org/pipermail/opencompute-onie/2014-September/000150.html

TL;DR -- please do not use the i2ctools version of the EEPROM code, it is deprecated.  Use the busybox version.

I will write more on the mailing list about this topic to reinforce the message for everyone.

cbrune: Sent more detailed message to the list for all to read:

machine/dell/dell_s4000_c2338/machine.make
lakesh-sharma: yes, my primary reason was that, we are using it 100 as the value, coz we used the same value in other platforms.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: there was push from MFG and folks, who work with MFG to change it to string, as they found it more user friendly.
e.g if we store it as "674-FF", then everyone know what it is after reading it back. Where as same value represented
in hex-octet would be "0x36 0x37 0x34 0x7E 0x46 0x46". By looking at octects, one can figure out, what it is.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: I agree on busy-box usage for i2ctool. But can we take up as an enhancement. Because this change would require sanity and regression at our end and would take some time. So can it be submitted later as separate commit. i.e. can we accept current commit with legacy i2ctool based implementation ?

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
cbrune: But the VENDOR_EXT TLV is not necessarily a readable string of characters.  It does contain binary data.  Quoting the spec:
```
This type code allows vendors to include extra information that is specific to the vendor
and cannot be specified using the other type codes. The format of this value field is a 
four byte IANA enterprise number, followed by a vendor defined string of bytes. The 
format of the string of bytes is entirely up to the vendor, except that it can be, at most, 
255 bytes long, including the IANA enterprise number. If more space is needed, then 
multiple TLVs with this type code can be used.
```

What type of data are you storing in this string?

Would an existing TLV type be more appropriate?

Would a *new* TLV type, generic in nature, be more appropriate?

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
cbrune: No.  I need to be firm about not accepting any more i2ctools patches.  Otherwise things just go on and on...

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: I agree, we would revert it. 
For any future new TLV, which is more appropriate, i would have to check with MFG and MFG contact folks.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: i would remove the (-a/-b) option.

machine/dell/dell_s4000_c2338/i2ctools/i2ctools-sys-eeprom-platform.patch
lakesh-sharma: I would remove the (-a/-b) option from syseeprom tool patch. A separate feature patch would be submitted for same foe eeprom code as you suggested, i.e. not part of platform patch.

Also i would check the usage of busy-box for eeprom access.

PR #164, opened 2015-03-22
Add mtree utility to ONIE (disabled by default)

cobracmder:
Change uclibc to include FTS support
Add mtree utility support (for NOS checks)


cobracmder:
For testing use:
```
ONIE_MIRROR = http://104.131.140.22/onie
CROSSTOOL_ONIE_MIRROR = http://104.131.140.22/onie/crosstool-NG
```
as the _mtree_ tarball is located there.

cobracmder:
```FTS``` is the file traversal support for ```libc``` which is why it's not a separate library. 

Which images would you like to know the size increase for PPC and x86?

cbrune:
I think David will be interested in the size of PowerPC AS4600.  In particular:

build/images/accton_as4600_54t-r0.itb
build/accton_as4600_54t-r0/sysroot.cpio

Before and after numbers would be good.

cbrune:
Generally I see no problem with this change.

However, I would split this into two commits.  One for the uClibc FTS changes and one for the rest of the mtree changes.

cobracmder:
I'm currently re-running the builds to get the ```sysroot.cpio``` sizes of the following platforms (as4600, 5712, and 6700).  Should be done shortly.

As per splitting up the commit, how would you like to see that? Same PR with 2 commits or close this PR and 1 PR per commit?

cbrune:
you can close this PR.  Then make 1 PR that has both commits in it.

PR #165, opened 2015-03-22
Fix typo in format string

cobracmder:


cbrune:
pushed to master:

- a99e224 Fix typo in format string

PR #166, opened 2015-03-22
Use consistent args for configure on various tools

cobracmder:
Ensure --prefix is $(DEV_SYSROOT)/usr vs /usr
Ensure CC is CC=$(CROSSPREFIX)gcc

### DIFF
--- a/build-config/make/dropbear.make
+++ b/build-config/make/dropbear.make
@@ -2,6 +2,7 @@
 #
 #  Copyright (C) 2013-2014 Curt Brune <curt@cumulusnetworks.com>
 #  Copyright (C) 2015 Nikolay Shopik <shopik@inblock.ru>
+#  Copyright (C) 2015 Carlos Cardenas <carlos@cumulusnetworks.com>
 #
 #  SPDX-License-Identifier:     GPL-2.0
 #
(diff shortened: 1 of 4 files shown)


cbrune:
What problem are you solving with these commits?  Nothing is broken as far as I known today.

Did this turn up when using Jessie?

The changes seem harmless.  I would avoid them if we can though.

If needed for Jessie then the changes are OK as long as they do not break Wheezy builds today.

For every tool touched in this commit you need to test that the tool still works.  The x86 virtual machine should be sufficient.  "works" can be as simple as "runs with --help" without errors.  I'm really looking to see that dynamically loaded libraries are all found, etc.

Chasing autoconf versions does not sound productive.

cobracmder:
I'm going to categorize this PR as **OCD** (not a problem, not related to jessie, etc...)

When I was in the process of adding mtree (#164), I ran into a three examples of using autotools:
1) Those that used a prefix of /usr
2) Those that defined CC of $(CROSSPREFIX)gcc
3) Those that did both

Not knowing why some tools used a one over another, I started experimenting with mtree and the 4 utils that fall into either 1) or 2).

All the tools worked *as is* and all of them work *after* the cosmetic changes.  We are just being explicit on the prefix and compiler to be used (even though the make system takes care of this for us).

From there, I went with the tools that had the most (which was (3)), thus the PR to the remaining 4 tools to be in line with the others.

I agree, this change isn't *productive* as the behavior has not been altered.

PR #169, opened 2015-03-30
As7700 32x upgrades to ONIE r1 for new hardware

david56:


cbrune:
This is quite a big patchset, but that is how it goes when adding a new CPU, like PowerPC64.

Individually I think I understand what the patches are doing.  My big question is "how did you test it" ?

You have probably already done it, but I have to ask.  Definitely need to verify that existing PowerPC-32 platforms using u-boot 2013.01.01 continue to build and function properly.  There are a few Accton machines in this category so you should be able to test that.

Following that I have some comments about the U-Boot patch reorganization.  Would it be possible to move the later u-boot patches (commits 8b47e85 and ac6df1) up in the patchset, closer to commit 0c418fd?  As it is currently it seems a bit strange to have those two commits at the end by themselves.

Also with the new common u-boot patch directory do you think it becomes hard to work with the core u-boot patches?  It seems like you really need to know what you are doing to make changes in the U-Boot patches, as patches could end up in 3 different directories (u-boot-version, u-boot-common and machine).  I understand why it is complex.

Also if you go down this round of common u-boot patches, is it possible to make other patches "common"?  For example can these also be common:

```
feature-dhcp-options.patch
feature-populate-serial-number.patch
```

Finally by last questions are about some feature patches present in 2013.01.01, but missing from fsl-sdk-v1.7 -- is there a reason why these are missing from fsl-sdk-v1.7 ?

```
feature-config-repeatable.patch
```

david56:
> Individually I think I understand what the patches are doing. My big question is "how did you test it" ?
> 
> You have probably already done it, but I have to ask. Definitely need to verify that existing PowerPC-32 platforms using u-boot 2013.01.01 continue to build and function properly. There are a few Accton machines in this category so you should be able to test that.

These patches have been built to images and running on PowerPC-32 platforms including ``AS4600_54T``, ``AS5600_52X``, ``AS5610_52X``, ``AS5710_54X`` and ``AS6700_32X``.  The ``sys_eeprom`` command in u-boot is normal as before.  Logging-in to ONIE via ``dropbear`` and performing ``update_url`` to update ONIE itself are also normal.

> Following that I have some comments about the U-Boot patch reorganization. Would it be possible to move the later u-boot patches (commits 8b47e85 and ac6df1) up in the patchset, closer to commit 0c418fd? As it is currently it seems a bit strange to have those two commits at the end by themselves.

Sure.  I'll change the patchset sequence as you said.

> Also with the new common u-boot patch directory do you think it becomes hard to work with the core u-boot patches? It seems like you really need to know what you are doing to make changes in the U-Boot patches, as patches could end up in 3 different directories (u-boot-version, u-boot-common and machine). I understand why it is complex.

It should have an additional/optional machine-common directory there.  I exploit ``cp-machine-patches`` to copy each patch located in either u-boot-version or u-boot-common directory.  It keeps the patch way same as machine's.  The complexity of patching u-boot should be close to the original one.

> Also if you go down this round of common u-boot patches, is it possible to make other patches "common"? For example can these also be common:
> 
> feature-dhcp-options.patch
> feature-populate-serial-number.patch
> feature-save-default-env-on-bad-crc.patch

The patches above can't be common because:

- fsl-sdk-v1.7 is based on u-boot v2014.07.  It has many differences then 2013.01.01.
- Each chunk does not have sufficient context for applying to fsl-sdk-v1.7 directly. (It had conflicts when I tried to apply them to fsl-sdk-v1.7 via ``patch`` command.  Eventually, I did it manually.  And it is why I just moved chunks creating new files to common patches.)
- In fsl-sdk-v1.7, it has ``CONFIG_SYS_GENERIC_BOARD`` enabled that makes u-boot do init via ``common/board_r.c`` instead of ``arch/powerpc/lib/board.c``.  And ``common/board_r.c`` never exists in 2013.01.01.

> Finally by last questions are about some feature patches present in 2013.01.01, but missing from fsl-sdk-v1.7 -- is there a reason why these are missing from fsl-sdk-v1.7 ?
> 
> feature-config-repeatable.patch
> feature-fdt-environment-size.patch

The target files have been removed from fsl-sdk-v1.7 (2014.07).  I did not know what place they had been moved.  Maybe we can complement these features in the future.

cbrune:
Alright, sounds good.  Pushed to master:

- 4f515ce Accton AS7700_32X: Enable to erase mass storage in uninstall mode
- c8ede70 Accton AS7700_32X: revise INSTALL to meet current design
- 1e53082 Accton AS7700_32X: move u-boot to fsl-sdk-v1.7
- b4eb863 Accton AS7700_32X: move linux kernel to fsl-sdk-v1.7
- 84c32ab Accton AS7700_32X: upgrade to ONIE r1
- 82c1f96 u-boot: move chunks which create new files to common patches
- b29d66d u-boot.make: add to support common patches
- 41d2eaf Add to support fsl-sdk-v1.7 u-boot
- edd0273 u-boot: be able to use different u-boot version
- 17db2db images.make: find library pathname to copy to sysroot
- 2f34a90 crosstool-ng: support building powerpc64 linux kernel
- 5ac5f2e Add to support fsl-sdk-v1.7 linux kernel

PR #173, opened 2015-04-03
Add NOS validation script

cobracmder:
Add test-nos.sh
Update README


cbrune:
Looks good.  I like the use of mtree.

Pushed to master:
- b92b3bb Add NOS validation script

PR #179, opened 2015-04-28
Accton AS7700_32X: restrict memory size to 1920M for running diag

david56:


cbrune:

- a246e8f Accton AS7700_32X: restrict memory size to 1920M for running diag

PR #180, opened 2015-04-28
add '60_diag_chainload' to grub.d for generating diag chainload entry

david56:
The diag chainload entry will disappear after doing the following
steps:

1. Run diag installer to install diag.
2. Enter ONIE uninstall mode to re-generate ONIE's ``grub.cfg``.
3. Run ONIE updater to upgrade ONIE on the box.

The patchset adds an extra script, ``60_diag_chainload``, to ``grub.d`` for generating diag chainload entry.  This script is used by demo OS installer, ONIE updater, and x86_64 ``uninstall-arch``.  NOS installer can call this script if it would like to keep hardware vendor's diag entry in NOS grub menu.

These patches have been tested on AS5712_54X platform.

I also update the trim list in ``demo.make`` to fix error message raised during demo OS rebooting.

Please help to review the patchset.  Thanks.

### DIFF
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 #  Copyright (C) 2014 Curt Brune <curt@cumulusnetworks.com>
-#  Copyright (C) 2014 david_yang <david_yang@accton.com>
+#  Copyright (C) 2014-2015 david_yang <david_yang@accton.com>
 #
 #  SPDX-License-Identifier:     GPL-2.0
 
@@ -272,17 +272,7 @@ EOF
 
 if [ "$demo_type" = "OS" ] ; then
     # If there is a diag image add a chainload entry for it.
-    diag_label=$(blkid | grep -- '-DIAG"' | sed -e 's/^.*LABEL="//' -e 's/".*$//')
-    if [ -n "$diag_label" ] ; then
-        onie_platform="$(onie-sysinfo -p)"
-        cat <<EOF >> $grub_cfg
-menuentry '$diag_label $onie_platform' {
-        search --no-floppy --label --set=root $diag_label
-        echo    'Loading $diag_label $onie_platform ...'
-        chainloader +1
-}
-EOF
-    fi
+    /mnt/onie-boot/onie/grub.d/60_diag_chainload >> $grub_cfg
 fi
 
 # Add menu entries for ONIE -- use the grub fragment provided by the
--- a/installer/x86_64/grub.d/60_diag_chainload
+++ b/installer/x86_64/grub.d/60_diag_chainload
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+#  Copyright (C) 2015 david_yang <david_yang@accton.com>
+#
+#  SPDX-License-Identifier:     GPL-2.0
+
+# This file provides a GRUB menu entry for diag.
+#
+# Place this file in /etc/grub.d and grub-mkconfig will use this file
+# when generating a grub configuration file.
+
+# If there is a diag partition add a chainload entry to the host OS GRUB
+# configuration.  Everything sent to stdout will end up in the host OS
+# GRUB configuration.
+diag_label=$(blkid | grep -- '-DIAG"' | sed -e 's/^.*LABEL="//' -e 's/".*$//')
+[ -z "$diag_label" ] && exit 0
+
+onie_platform="$(onie-sysinfo -p)"
+
+cat <<EOF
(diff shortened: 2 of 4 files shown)


cbrune:
This definitely is a problem.  I also encountered it recently working on and testing a UEFI version of ONIE. Things get quite a bit better with UEFI -- stayed tuned for that.

Nevertheless I think I have a cleaner way of handling the DIAG GRUB entry for legacy BIOS.  I would like to hear your thoughts.

Instead of adding a new file, like 60_diag_chainload, how about adding the logic to the existing 50_onie_grub file.  That is a little nicer as scripts, NOSes, etc only have to invoke one script.

My changes are basically the same as yours, I just added this to the bottom of 50_onie_grub:

```shell
# Look for an ONIE compatible DIAG image and add a boot entry for it.
diag_label=$(blkid | grep -- '-DIAG"' | sed -e 's/^.*LABEL="//' -e 's/".*$//')
if [ -n "$diag_label" ]  ; then
    cat <<EOF >> $grub_cfg
menuentry '$diag_label' {
        search --no-floppy --label --set=root $diag_label
        echo    'Loading $diag_label $onie_platform ...'
        chainloader +1
}
EOF
fi
```

The benefits I see are:
- no new file, 60_diag_chainload
- the demo OS installer does not need to check whether diag or not, just run 50_onie_grub every time
- the uninstaller does not need to do *anything* regarding the DIAG entry.  Since NOS's are using 50_onie_grub during install the DIAG GRUB entry is *already* in ONIE grub.cfg

I also like that this approach eliminated the need to append to ONIE's grub.cfg in what I felt were strange ways.

This approach hinges on all NOS and DIAG installers using 50_onie_grub, which they should be doing as that has been the design all along.

david56:
Your approach is great!

We found another issue by doing the following steps:

1. run demo diag installer to install demo DIAG
2. run demo OS installer to install demo OS
3. run demo DIAG installer again

After doing the 3 steps above, the GRUB in MBR will be overridden to demo DIAG's.  It makes demo OS entry *disappeared* on the GRUB menu.  Maybe demo OS/DIAG installer should not always install GRUB into MBR.  i.e., if the demo OS/DIAG partition has existed, do not install GRUB into MBR again.

    # Pretend we are a major distro and install GRUB into the MBR of
    # $blk_dev.
    grub-install --boot-directory="$demo_mnt" --recheck "$blk_dev" || {
        echo "ERROR: grub-install failed on: $blk_dev"
        exit 1
    }

cbrune:
@david56 - Can you make a PR that adds the DEMO diag logic to 50_onie_grub ?  That will solve a number of problems in the short term.

david56:
Sure.

cbrune:
Regarding the re-install of the DIAG making the NOS unavailable -- I agree that behaviour is not ideal.

I am thinking of might happen in the real world - what you are really talking about is updating the DIAG in the field, after a NOS is installed:

- will people re-install (or upgrade) the diag after the NOS is already installed?
- the ONIE docs state that the DIAG must be installed right after ONIE:  https://github.com/opencomputeproject/onie/wiki/Demo-Diag -- I know documenting correct behaviour is not the same thing as making behaviour fool proof.
- is it acceptable in the field to wipe the NOS when installing a new HW DIAG ?
- in the future with UEFI ONIE the MBR is no longer used and this is not an issue.

I don't see how an installer can know whether or not to install into the MBR for certain.  Yes, the existence of the partition is a clue, but it is not a 100% guarantee.

The entire legacy BIOS MBR method for ONIE and OSes is troublesome.  Every OS wants to own the MBR, but only one can.  This is what makes the UEFI (stay tuned) so much better.

Just thinking out loud here -- I'm leaning towards not fixing this issue, but am open to suggestions.  I think the major downside is having to reinstall a NOS after reinstalling the DIAG.  That might be acceptable.

david56:
> The entire legacy BIOS MBR method for ONIE and OSes is troublesome. Every OS wants to own the MBR, but only one can. This is what makes the UEFI (stay tuned) so much better.

Is GRUB able to do centralized management? I mean the MBR is always own by ONIE and there are scripts, ONIE, DIAG, NOS, and something else, in ``grub.d``.  For each time of install/update, the installer/updater copies its grub script to ``grub.d`` and calls an ONIE service to refresh the ``grub.cfg``.  In the uninstall case, ONIE can remove NOS script from ``grub.d`` then refresh ``grub.cfg``.  I'm not sure if the GRUB has the capability to do this.  Or it has much complexity to maintain the machanism.

cbrune:
GRUB does not handled centralized management well.  It is goofy, because everyone wants to own the MBR, but only one OS at a time can.  UEFI is much better in this regard.

PR #181, opened 2015-04-29
demo.make: update trim list

david56:
As mentioned in PR #180


cbrune:
pushed to master:

- 1b55b75 demo.make: update trim list

PR #186, opened 2015-05-06
Accton AS7700_32X u-boot: correct IFC clock frequency to 100MHz

david56:


cbrune:

- a058473 Accton AS7700_32X u-boot: correct IFC clock frequency to 100MHz

PR #187, opened 2015-05-07
Create TLV EEPROM block in disk.

vic1231:
This pull request is only for review and to see create one more partition as TLV EEPROM.
And with uninstall utility, partition of TLV EEPROM will not removed when uninstall NOS.
I think it can't not get a pull permit and merge to master branch.
Is these files belongs to my local side issue?


vic1231:
Add one file, machine/wnc/wnc_model/busybox/patches/onie-syseeprom-wnc_modify.patch.
This version just only modify with mark mtd and norflash check in: sys_eeprom_mtd.c/read_sys_eeprom() and write_sys_eeprom() function.
My develop board was embeded a SSD card, 256 GB.
Is mark mtd and norflash check code on my patch file(/wnc/wnc_model/busybox/patches/onie-syseeprom-wnc_modify.patch) permitted?

cbrune:
I do not want to add support to ONIE for storing the TLV EEPROM data on the hard drive. It seems like a really bad idea to store vital product data on the hard disk. The VPD should be stored somewhere more "out of the way" like an EEPROM or NOR / NAND flash.

If your development board does not have EEPROM or NOR or NAND flash then just disable support for the onie-syseeprom command during development.

For production the product needs to have EEPROM or NOR or NAND flash in order to be ONIE compliant. See the ONIE hardware requirements:
https://github.com/opencomputeproject/onie/wiki/Design-Spec-x86-HW-Requirements#x86-hardware-requirements

ericmswu:
hello cbrune,

I want to clarify the request for accessing EEPROM via disk interface. The purpose is not to store EEPROM content in real hard disk. For some existing OCP micro server card (Panther+), it does not have EEPROM on it, but we want to leverage the card on switch design. To support the TLV EEPROM requirement, we simulate a disk "to hold the content of EEPROM" to host via some external non-volatile device. by doing in this way, we do not need usb2i2c converter (save cost) to simulate an i2c device to host. and the application can just use open("/dev/sd?") & read to get the content of EEPROM. 
 
can we have the flexibility to support EEPROM in this way and also make it compatible with ONIE?

thanks,
eric

cbrune:
Hi Eric,

Read all the comments in this thread and most more if you have further questions:
http://lists.opencompute.org/pipermail/opencompute-onie/2015-June/000719.html

Storing vital product data (VPD) in a hard disk is *really* bad idea.

ericmswu:
hi cbrune,
as mentioned, it's not a real hard disk. Due to no eeprom on Panther+, the TLV non-voltile storage (actually, it's located in EEPROM in external controller) is accessed via disk interface (then redirect to EEPROM in external controller). by doing in this way, the TLV contents can be easily read by "read" api

thanks,
Eric

cbrune:
OK, but it sounds a little wacky.  Go ahead and make a pull request and let's see what it looks like.

vic1231:
Hi cbrune:
Let me clarify Eric's description, he's one of my project partner.
For now, you know my TLV EEPROM is accessed in "CPU => USB => MCP2221 => I2C => EEPROM" way. I call this EEPROM A here.
And, in out HW design now, we have a BCM(a co-processor) connected to a EEPROM, I call this EEPROM B here.
The access way with EEPROM A is allowed from you(or Ellen).
However, we focus on access EEPROM B now. The way is "CPU => USB => BMC (Virtual USB-to-I2C) => BMC (Virtual EEPROM) => I2C => EEPROM" is our target.
As you seen, the TLV info will stored in a real EEPROM(which is EEPROM B) finally. But this way is via BMC, not mcp2221.
And my code section for access EEPROM A now is like "open("/dev/i2c-0, O_RDWR);".
In the way of access EEPROM B which may be change to "open("/dev/sd?, O_RDWR);".
But code has not implemented now, and I think it is no necessary to pull.
Because in HW view, TLV info will stored in a real EEPROM(which is EEPROM B); in SW view, the device opening interface will changed from /dev/i2c to /dev/sdx where /dev/sdx in a virtual HD interface, not real HD, SSD or something like that.
Sorry, curt, if you are confused with my description. Then you can think these descriptions is just explain a different way to access a EEPROM.
You know not everyone will knows every method clearly for access EEPROM, some way may be their know how.
Short speaking, is it allowed with any method of access a "real EEPROM" for TLV?
This is the key point that I want to check with you.

mandeepsandhu:
Hi @vic1231,

Sorry to jump into your conversation. I assume you have a BMC (Baseboard Management Controller) in your system which connects to the real EEPROM (B) via I2C.

So from the host, shouldn't you be using **IPMI FRU** commands to access the contents of the system?

We are planning a similar system. The EEPROM connected to the BMC stores data according to the **IPMI FRU specification**, in which case the ONIE-TLV info is kept in the **Internal Use Area** FRU section. The `onie-syseeprom` command would then make use of IPMI FRU commands to extract this information.

We have had to follow this approach as the IPMI specification require us to have the main system EEPROM store data according to the FRU specs. 

HTH,
-mandeep

cbrune:
Hi @vic1231 -- thanks for the clarification.  I understand now.

From the ONIE perspective, accessing the EEPROM as you describe should be fine:
```
open("/dev/sd?, O_RDWR);
```

As long as the final target device is a real EEPROM I have no problem with it.  I am just *very* cautious about storing EEPROM data in the same hard drive as the ONIE kernel and initramfs.

vic1231:
Thanks, Curt. I also confused before I check with Eric's problem, too.
And mandeep:
I have studied some IPMI introduction before, but still don't know much
better with this. I will survey more informations. Base on your assume,
I'll check my BMC is connected to EEPROM (B) via I2C. But the method you
described, is it need porting something SW more, write BMC code, or HW
specifications?


2015-07-03 2:47 GMT+08:00 Curt Brune <notifications@github.com>:

mandeepsandhu:
@vic1231 For communicating with the BMC you will need to have IPMI firmware running on it. Mostly this is proprietary software (from vendors like **AMI**/**Avocent**) so I can't help you much on that front. On the main OS (host), you can use tools like `ipmitool` to communicate with the BMC using IPMI commands (see `ipmitool`'s manpage).

PR #188, opened 2015-05-07
x86_64 install-arch: move 'onie_boot_dev' to proper place

david56:
``onie_boot_dev`` is not defined before preserving ``grub_env``.
Move the statement of assigning ``onie_boot_dev`` to the place prior
to preserve ``grub_env``.


cbrune:
Good catch!  Thanks for that.

Pushed to master:
- a9f6a48 x86_64 install-arch: move 'onie_boot_dev' to proper place

PR #189, opened 2015-05-08
virtual machine: support multiple disk devices

cbrune:
Previously the ONIE boot process only created /dev device nodes for
the first virtual disk device.  Running the virtual machine with
multiple disk drives did not work properly as the minor numbers for
the /dev device nodes were wrong.

This patch fixes up that problem by reading the proper minor device
numbers from sysfs for the virtual block devices.

This was tested by using qemu to start a machine with multiple disk
drives.


cobracmder:
+1

cbrune:
pushed to master:

- 9dfc654 virtual machine: support multiple disk devices

PR #190, opened 2015-05-08
UEFI tools

cbrune:
Add a few tools needed for working with UEFI.


cobracmder:
+1

cbrune:
Regarding the compiler I'm not sure what you can do except upgrade your compiler.

The ONIE project targets the stable Debian distribution.  Debian is usually 'old' compared to most distributions, which is why I chose it.  The old version of gcc on Debian wheezy is 4.7.2, so your gcc 4.4.7 is quite old.

For reference current Debian jessie uses 4.9.2 and Ubuntu 14.04 uses 4.8.2.

david56:
Just found another issue:

    x86_64-onie-linux-uclibc-gcc  -DEFIBOOTMGR_VERSION=\"0.11.0-onie\" -Wsign-compare -Wall -Werror -g -D_FILE_OFFSET_BITS=64 -Os --sysroot=/home/david_yang/onie/build/accton_as6712_32x-r0/dev-sysroot -Isrc/lib -Isrc/include   -c -o src/lib/scsi_ioctls.o src/lib/scsi_ioctls.c
    src/lib/scsi_ioctls.c:27:24: fatal error: linux/nvme.h: No such file or directory

This looks like the linux kernel should have to upgrade to 3.4 or newer for building efibootmgr.

cbrune:
Indeed the 3.2 kernel is not good enough for UEFI.

I did all of this work using the kvm and a 3.14 series kernel.  When the UEFI changes are ready I plan to move the KVM forward to a new kernel version also.

cbrune:
pushed to master:

- 685180e x86_64-uefi: add dosfstools package
- 8542c9d x86_64-uefi: add support for efibootmgr tool

PR #191, opened 2015-05-08
grub: build GRUB2 i386-pc and x86_64-efi for target and host

cbrune:
This patch updates the grub.make fragment to build GRUB2 for the
following scenarios:

- GRUB2 i386-pc for the target platform
- GRUB2 x86_64-efi for the target platform
- GRUB2 i386-pc for the build host
- GRUB2 x86_64-efi for the build host

The i386-pc and x86_64-efi are both built for the target platform in
order to support run time detection and installation of ONIE on either
legacy BIOS or UEFI firmware.  Having both sets of tools available
allows the ONIE installer to install the correct binaries at run time.

The i386-pc and x86_64-efi are both built for the build host in
order to support the creation of a single .ISO image that will work
with either legacy BIOS or UEFI firmware.


cobracmder:
+1

cbrune:
pushed to master:

- a4250d9 grub: build GRUB2 i386-pc and x86_64-efi for target and host

PR #192, opened 2015-05-08
Use GRUB2 to create the ISO image

cbrune:
Improve the ISO image creation process by only using GRUB2.  With this approach syslinux is no longer needed.

This patchset also includes a few platform modifications, where syslinux constructs were removed.

@david56 -- For Accton see 3c039bb
@willkuo103 -- For Alpha Networks see eb06f46


cbrune:
pushed to master:

- 8e8b33c alphanetworks: remove CONSOLE_FLAG parameter
- fd8af1f accton: remove CONSOLE_FLAG parameter
- 2471f4a Remove syslinux package
- 4155da0 x86: ISO image creation improvements

PR #193, opened 2015-05-12
grub: revise usage which is similar to the one in PC

david56:
This patch intents to revise the GRUB usage which is similar to the
one in PC.  These usages are:

- GRUB is always installed in MBR and ONIE partition during ONIE
  installing.  DIAG and NOS partitions never installed.  This also
  reduces potential risk of GRUB broken.
- The main logic to manage ``grub.cfg`` is formed to a set of
  functions in ``onie-blkdev-common`` in ONIE. Installer just invokes
  the function to refresh ``grub.cfg``.  It keeps installer script
  simple.
- The ``grub.d`` directory has script files come from installer used
  to generate menu entries for ONIE, DIAG, or NOS.
- In uninstall mode, ONIE removes NOS script file from ``grub.d`` and
  then refresh ``grub.cfg``
- There is no limiation of the order to install DIAG and NOS.

This patch has been tested on Accton AS6712_32X.

This usage maybe not the best.  It intents to move the main logic of GRUB management to ONIE itself and keep installer simple.  Please help to review the patch.  Thank you very much.


cbrune:
This proposal is interesting, but the entire ONIE / NOS GRUB situation is very difficult with the MBR and legacy BIOS.

The key observation is that the legacy BIOS and corresponding GRUB MBR method is not well suited for installing multiple operating systems.  Installing multiple operating systems is exactly what we are doing here with ONIE, a DIAG and a NOS.  Every OS wants to be in control, but there is only one MBR.

Moving to UEFI obviates all of these issues.  One of the UEFI design goals was to deal with the multi-OS problem and they have done a pretty good job.

The current implementation puts the NOS in control, more or less.  Your implementation puts ONIE in control, more or less.  There are pluses and minuses to both views.  In the end though, neither solution is easy to work with.

As long as we use legacy BIOS the GRUB/MBR situation will be difficult, weird and something we just have to deal with.

Let's move to UEFI :)

Side comment on the order of installing DIAG and NOS -- As long as the NOS and DIAG share a common storage medium you will always want to install the DIAG first.  The NOS installer could very likely claim all available disk space for itself, leaving no room for the DIAG.  The only way for the DIAG to claim space is to be installed first.  That will be true for legacy BIOS as well as UEFI.

On the other hand the HW platform could have separate, dedicated DIAG storage, that the OS cannot use.  For example a dedicated NOR/NAND/SPI-ROM flash part.

david56:
I see.  The patch will influence all present installer/updater behaviors.  It seems to have no enough advantages to change current design.  Thank you for your opinions.  I'm closing the PR.

PR #194, opened 2015-05-14
Add acpica-tools to ONIE

cobracmder:
Add acpica-tools to ONIE (disabled by default). This is the first
commit that will be used for an APD aware ONIE diagnostic image.

Tested against 2015.05 kvm_x86_64.


cbrune:

- e23df06 Add acpica-tools to ONIE

PR #163, opened 2015-05-14
Debian Jessie requires libtool-bin package

cobracmder:
In Debian Jessie, *libtool* package (which used to include ```libtoolize``` and ```libtool``` along with aclocal macros) was broken up to have a separate package called libtool-bin which now contains ```libtool```.  This means in Jessie, we need to install both *libtool* and *libtool-bin* packages.

What's the best way to handle this as having *libtool-bin* in the required packages will cause an error during the *debian-prepare-build-host* target?


cbrune:
pushed to master:

- 4d156fb Fix ONIE build on Debian Jessie

PR #196, opened 2015-05-21
Support storing the "eeprom" on "disk".

squeakbat:
[I discussed this with Curt and we came to the conclusion that this feature is acceptable as long as nobody admits to using it.]

As a last resort on machines with absolutely no usable eeprom,
we can put the eeprom contents on the boot device.

It is an abomination.

Since there is plenty of unused space in the gpt area, we use
the 2k bytes at the end, just before the first partition.


cbrune:
I agree this should only be used in extremely dire circumstances, or during development while the hardware is in flux.

I like that this implementation does not use a partition.  I think this approach will also help out @vic1231 who wanted something similar in PR #187 .

Pushed to master:
- 146461d Support storing the "eeprom" on "disk".

vic1231:
Sure, I will try this before my developing board has really i2c EEPROM.

2015-05-22 0:10 GMT+08:00 Curt Brune <notifications@github.com>:

PR #200, opened 2015-05-28
xorriso: use system default version

cbrune:
During development of UEFI I experimenting with different versions of
the xorriso program.  The initial patches allow xorriso to be a
parameter to the onie-mk-iso.sh script.

Now that the dust has settled the stock distribution version of
xorriso is sufficient and the XORRISO parameter is no longer required.

This patch removes the xorriso argument and does a little clean up of
the onie-mk-iso.sh script.


cbrune:
pushed to master:

- d8179bf xorriso: use system default version

PR #202, opened 2015-05-29
Accton AS7700_32X: add eth1addr and eth2addr for diag test

david56:


cbrune:
pushed to master:

- bb295dc Accton AS7700_32X: add eth1addr and eth2addr for diag test

PR #205, opened 2015-06-02
add kexec-tools to build system

cbrune:
This patch adds the kexec-tools (kexec and kdump) to the ONIE build
system.

These tools are built by default for x86 platforms and included in the
final image.

For powerpc these tools can be built, but are not by default.  To
build these tools for a powerpc platform add this to the platform's
machine.make:

  KEXEC_ENABLE = yes


cbrune:
pushed to master:

- 02f34e4 add kexec-tools to build system

PR #206, opened 2015-06-02
Powerpc 4.0.4

cbrune:
These commits give the fsl_p2020rdbpca machine (the demonstration P2020 PowerPC platform) a long over due update.

Additionally support for the 4.0.4 kernel is brought to PowerPC for the first time.


cbrune:
plan to sit on this until LTS kernel 4.1 is available

PR #207, opened 2015-06-04
Check eeprom r/w function

vic1231:
I use usb to i2c chip as an interface in my project. So the kernel/ adds a driver. So the eeprom r/w function is different with others. This request just for check eeprom r/w function content validity, and these codes works well. Do not merge to master, the project still in develop, the model name also not ready yet.


cbrune:
Thanks for posting your 'in development' version.  It helps a lot to share early and often -- it allows for making changes before the final version.

I have a few general comments and specific comments.

Specific comments:

1. I am trying to understand what requires the writing of the two zero bytes.

Is the two bytes required by the MCP2221?  Or is the two byte write required by the EEPROM attached to the MCP2221?

2. Did you back port the MCP2221 driver from another kernel version?  Did you make any platform specific changes to the driver itself?

If the MCP2221 driver is simply a back port, without any platform specific changes, it will be better to put into the main kernel tree for everyone to use instead of in the machine specific patch area.  In other words this patch belongs in patches/kernel/3.2.35/, not machine/wnc/wnc_model/kernel/.

General comments:

The comments on the two commits needs to be a lot better.  Right now they have no comments.  I understand these are testing commits, so no harm.  Just keep in mind the philosophy of making a good patch:

vic1231:
Hi Curt:
First, about requires the writing of the two zero bytes, as the information I get, it is required by EEPROM. From http://www.atmel.com/Images/doc0670.pdf Figure 9 & Figure 11. The EEPROM use two bytes (MSB + LSB) to setup the read / write offset.

Second, I don;t know the meaning of "back port". I don't port the driver to another kernel version, and I changed MCP2221_STD_DELAY_MS value.
But now there is another issue to me. The driver is provided by Microchip, the vendor. I may need to contact Microchip about the license of the driver code to decide pull it up or not.

cbrune:
Back porting means a version of the driver already exists in an upstream kernel version.  For example ONIE is currently based on an older 3.2.X kernel series.  Sometimes hardware support is added to a more recent kernel, like 3.19.X for example. The work of "porting" a recent driver from the 3.19.X to an older 3.2.X kernel is called "back porting".

So I am asking if a version of this MCP2221 driver is already in an upstream kernel release or not.  You mention you received the driver from the vendor.  You should check with the vendor whether the driver is already upstream or not.   Also check if including it in ONIE is OK.

vic1231:
Okay, I only built this driver at kernel 3.14.27. I think it is a back porting with newest ONIE linux kernel version?
As I know, this MCP2221 driver is not in an upstream kernel release.
I'll check it for including mcp2221 driver in ONIE is OK or not.
And at least I check the content of eeprom r/w function in busybox is OK. Now this testing commits can be closed.

PR #209, opened 2015-06-05
Accton AS7700_32X: rename to AS7710_32X

david56:
The new ONIE updater shall be compatible with the following machines:

- accton_as7700_32x-r0
- accton_as7700_32x-r1
- accton_as7710_32x-r0


cbrune:
Does the INSTALL file need to change also?

david56:
Thanks for reminding.  It shall not be compatible with ``accton_as7700_32x-r0``.  I have revised ``install-platform`` and ``INSTALL`` for backward compatibility.

cbrune:

- 48108cf Accton AS7700_32X: rename to AS7710_32X

PR #210, opened 2015-06-05
Accton AS7702_32X: rename to AS7712_32X_1

david56:
The new ONIE updater shall be compatible with the following machines:

- accton_as7702_32x-r0
- accton_as7712_32x_1-r0


cbrune:

- d67b528 Accton AS7702_32X: rename to AS7712_32X_1

PR #211, opened 2015-06-08
efibootmgr / efivar: move to recent versions

cbrune:
This patch moves to the currently released versions of efibootmgr and
efivar:

  efibootmgr -- version 0.20
  efivar     -- version 0.12

This also allows us to drop a bunch of patches that were keeping these
components current.

Testing Done:

Built images for the kvm_x86_64 and exercised using QEMU with OVFM.
Went through:

- installing ONIE UEFI
- installing demo DIAG
- installing demo NOS
- uninstall demo NOS


cbrune:
pushed to master:

- 8f69402 efibootmgr / efivar: move to recent versions

PR #215, opened 2015-06-18
networking.sh

vic1231:
Modify networking.sh
Detect i2c before onie-sysinfo -e.
It can not get any TLV informations from EEPROM if i2c interface is not ready.
In my system, I'm sure i2c is not ready when networking.sh is running.
Maybe my own built-in r/w EEPROM driver is noot ready at that time.
But it works if I sleep 1 seconds before access EEPROM.
I don't know is there anyone have same issue with me.
Should I maintain this code which I added in my local side?


cbrune:
let's hash this out on the existing mailing list thread:
http://lists.opencompute.org/pipermail/opencompute-onie/2015-June/000743.html

PR #216, opened 2015-06-24
efibootmgr v0.12 clean up

cbrune:
efibootmgr v0.12 no longer depends on pciutils, so drop that package
from the ONIE build process.

This patch cleans up a few things:

- removes all traces of pciutils

- removes old efibootmgr patches that should have been removed when we
  moved to v0.12.


cbrune:
pushed to master:

- 91d7656 efibootmgr v0.12 clean up

PR #217, opened 2015-06-25
x86: add flashrom support

cbrune:
This patch adds the 'flashrom' utility and pciutils library to the
ONIE initramfs for x86 builds.  This tool is *very* handy on x86
systems for updating the BIOS firmware.

Adding the utility and library increases the size of the kvm_x86_64
image by ~100,000 bytes.

By default most flashrom methods except for 'internal' and Intel NIC
EEPROMs are disabled from the build.  Individual platforms can define
FLASHROM_MAKE_CONFIG_PLATFORM to enable support for specific flashing
methods, for example:

  FLASHROM_MAKE_CONFIG_PLATFORM = CONFIG_XYZ=yes

See the Makefile in the flashrom source directory for the complete
list of CONFIG_XYZ variables.

Note: Leaving the Intel NIC EEPROMs enabled, as that could be useful
in manufacturing to program the MAC address for eth0.

### DIFF
--- a/build-config/Makefile
+++ b/build-config/Makefile
@@ -325,6 +325,10 @@ endif
 ifeq ($(KEXEC_ENABLE),yes)
   include make/kexec-tools.make
 endif
+ifeq ($(FLASHROM_ENABLE),yes)
+  include make/pciutils.make
+  include make/flashrom.make
+endif
 include make/images.make
 include make/demo.make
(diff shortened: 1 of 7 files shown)


cbrune:
Pushed to master:

- 79992b6 x86: add flashrom support

PR #218, opened 2015-06-30
Kernel 4.1.1

cbrune:
This patch series adds support for Linux kernel 4.1.1.

4.1.1 is the first stable kernel release in the next long term support (LTS) kernel series from kernel.org.

This patch series also updates the ONIE reference platforms, kvm_x86_64 (x86) and fsl_p2020rdbpca (powerpc-softfloat), to use the new kernel series.


cbrune:
Pushed to master:

- 4946fb4 fsl_p2020rdbpca: general updates and use kernel 4.1.1
- 6642856 powerpc: add support for uClibc 0.9.33.2
- f1da01b kvm_x86_64: move to kernel version 4.1.1
- a3522af kernel: move from 4.0.4 to 4.1.1 kernel

PR #219, opened 2015-07-01
image discovery waterfall: add well known server name

### LINKED ISSUE #203: Add well known hostname to image discovery waterfall
In the image discovery watefall add a "well known" hostname (a string) to list of servers.

Proposing here to use the hostname `onie-server` as the well known host name for an ONIE server.

During the image discovery waterfall if ONIE is able to resolve the host name `onie-server` add that server to the list of servers used for locating an image.

This gives flexibility in deployments where the users have control over the DNS server, but not over the DHCP server.

For example a user could set their laptop to the host name "onie-server" when making a DHCP request -- as is common in simple lab environments and home routers, the host name "onie-server" will be added to the local DNS domain.  Then the user can start up a simple python based ONIE server on the laptop and ONIE will find the image via the host name "onie-server".

cbrune:
This patch adds a "well known" hostname to the list of servers used
during the image discovery waterfall process.  The default ONIE server
name is "onie-server".

This server name is added to the list of servers used during the HTTP
waterfall and TFTP waterfall.

Fixes #203

Quoting the issue:

  This gives flexibility in deployments where the users have control
  over the DNS server, but not over the DHCP server.

  For example a user could set their laptop to the host name
  "onie-server" when making a DHCP request -- as is common in simple
  lab environments and home routers, the host name "onie-server" will
  be added to the local DNS domain. Then the user can start up a
  simple python based ONIE server on the laptop and ONIE will find the
  image via the host name "onie-server".

Testing Done:

Tested both the HTTP and TFTP waterfalls using the fsl_p2020rdbpca
(powerpc) reference platform.

1. The demo OS installer image was placed at the root of an HTTP
server, whose hostname was "onie-server", using one of the default OS
installer image names.  The HTTP image discovery process correctly
located, downloaded and installed the image.

2. The demo OS installer image was placed at the root of an TFTP
server, whose hostname was "onie-server", using one of the default OS
installer image names.  The TFTP image discovery process correctly
located, downloaded and installed the image.

### DIFF
--- a/rootconf/default/bin/exec_installer
+++ b/rootconf/default/bin/exec_installer
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-#  Copyright (C) 2013-2014 Curt Brune <curt@cumulusnetworks.com>
+#  Copyright (C) 2013-2015 Curt Brune <curt@cumulusnetworks.com>
 #  Copyright (C) 2014 Matt Peterson <matt-github@peterson.org>
 #  Copyright (C) 2014-2015 david_yang <david_yang@accton.com>
 #
@@ -198,7 +198,7 @@ url_run()
 http_download()
 {
     # Build list of HTTP servers to try
-    http_servers=
+    http_servers="$onie_server_name"
 
     # HTTP server IP only
     if [ -n "$onie_disco_wwwsrv" ] ; then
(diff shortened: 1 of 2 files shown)


cobracmder:
This looks good...how do you want to track making sure the wiki and testing proceedures/OCE are updated to tests these cases?

cbrune:
That is a good question.  I could use some help.

I can handle the spec and the testing procedure parts of the wiki.  I could use help implementing the test case.

cobracmder:
No problem: I can take the OCE test cases...do you want to track the progress with this issue/PR or create another one?

cbrune:
I think creating "issues" for the other tasks is the way to go.  This PR is just for the code.  The issues can refer to this PR #219

cbrune:
pushed to master:

- 55b0b13 image discovery waterfall: add well known server name

PR #222, opened 2015-07-01
networking: add link-local IPv4 address as fallback

### LINKED ISSUE #204: Use 169.254.x.x IPv4 address when DHCP fails
In addition to the standard ONIE fallback IP address (192.168.10.3), follow RFC-3927 and RFC-6890 to allocate a link local IP address as described here:

https://en.wikipedia.org/wiki/Link-local_address#IPv4

cbrune:
Closes #204

When DHCP fails, allocate a link-local IPv4 address.  This is in
addition to the standard ONIE fallback IPv4 address 192.168.10.3.

This patch allocates the link local IPv4 address as described in
RFC-3927.

Testing:

When DHCP is unavailable, verified the following:

- The system configures an additional 169.254.x.y IPv4 address

- The system makes at most 20 attempts to configure an 169.254.x.y
  IPv4 address


nolanl:
Looks great!

cbrune:
pushed to master:

- a02b948 networking: add link-local IPv4 address as fallback

PR #224, opened 2015-07-06
Update OCE for image discovery waterfall: use RFC-3927 link-local IPv4 address #222

cobracmder:
Per #222, update OCE to properly test these newly added test cases.


cobracmder:
pushed to master https://github.com/opencomputeproject/onie/commit/36ae6f391527f83a36b7b3f27a2c269910541643

PR #231, opened 2015-07-09
make-devices.pl: create block device nodes for 'sdc' and 'sdd'

david56:
It will have at most 4 storages in Accton's new machines that need
more device nodes for accessing data.


cbrune:
Looks good.  Thanks!

Pushed to master:

- d65e378 make-devices.pl: create block device nodes for 'sdc' and 'sdd'

PR #221, opened 2015-07-09
Update OCE for image discovery waterfall: add well known server name #219

cobracmder:
Per #219, update OCE to properly test these newly added test cases


cobracmder:
Pushed to master: https://github.com/opencomputeproject/onie/commit/a65e36985ff28257662ed379afc68113e78efe22 Add tests for using well-known server name

PR #233, opened 2015-07-13
Firmware update

cbrune:
This patchset adds the following:

- an additional update image discovery mechanism.  The discovery process will now also look in a well known, persistent directory on the `ONIE-BOOT` partition.

- a new CLI command, `onie-fwpkg`, for managing ONIE update images and the well known, persistent directory on the ONIE-BOOT partition.


cbrune:
pushed to master:

- 477cd47 x86 firmware update: add onie-fwpkg CLI tool
- a8e712b pending firmware update discovery mechanism

PR #236, opened 2015-07-14
oce: build-onie.py use subprocess.check_call()

cbrune:
In order to catch build failures use subprocess.check_call() instead
of subprocess.call().  check_call() will raise a CalledProcessError
exception if the underlying command exit code is non-zero.

This allows scripts that call build-onie.py to check its return code
and take appropriate actions.

Testing:

Ran build-onie.py with a bogus make argument using the --make-args
argument, such that the underlying make command would fail.
build-onie.py displayed a back trace and exited with a non-zero exit
code.


cobracmder:
Good catch.

Pushed to master: https://github.com/opencomputeproject/onie/commit/6c27c3ad9a5e022d1b0e52c2f7be016cf2b5014b

PR #237, opened 2015-07-14
grub: fix race condition in grub build

cbrune:
When running 'make all' with -j24 occasionally the build would fail
for the grub-install and grub-host-install targets.

The grub builds share a common source directory and are supposed to
use separate output directory.  That part was working fine.

The trouble was the build of the grub info pages (documentation)
happens in the source directory, instead of the separate binary output
directories.  The parallel builds would clobber each other building
the documentation.

The fix here is to remove the building of the grub documentation,
which we are not using anyway.

Testing:

Before the fix building an x86 machine from scratch would fail 1 out
of 5 times or so.  After the fix *all* x86 machines built from scratch
successfully.

### DIFF
--- a/patches/grub/disable-build-of-documentation.patch
+++ b/patches/grub/disable-build-of-documentation.patch
@@ -1,0 +1,17 @@
+GRUB2: Disable build of documentation
+
+ONIE does not require the building of the documentation
+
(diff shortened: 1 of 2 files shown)


cbrune:
pushed to master:

- 8c53b04 grub: fix race condition in grub build

PR #240, opened 2015-07-17
kvm_x86_64: allow some flexibility for install disk

cbrune:
Previously the kvm_x86_64 machine defaulted to using /dev/vda for
installing ONIE.  This was fine as long as the VM is configured to use
a virtual IO block device for the storage medium.  This works great
with QEMU.

Some hypervisors do not support virtual IO block devices.  However,
SATA controller are commonly supported.  In order to be more flexible
in the choice of hypervisor this patch let's the installer choose
either /dev/vda or /dev/sda at runtime for installing ONIE.

Testing:

Installed the ONIE image using a virtio disk controller and using a
SATA disk controller.

### DIFF
--- a/machine/kvm_x86_64/installer.conf
+++ b/machine/kvm_x86_64/installer.conf
@@ -10,10 +10,25 @@ description="KVM, VM-X86_64"
 # Default ONIE block device
 install_device_platform()
 {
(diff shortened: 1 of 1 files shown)


cbrune:
pushed to master:

- 3eab41d kvm_x86_64: allow some flexibility for install disk

PR #241, opened 2015-07-21
kvm_x86_64: updated install instructions

quentinzil:
Increase the size of the VM to 1 GB.
Change the image format from raw to qcow2.


cbrune:
Looks good, pushed to master:

- 2c201be kvm_x86_64: updated install instructions

PR #242, opened 2015-07-22
Move default kernel version to 4.1.3

cbrune:
Update kernel version for the latest kernel release.

Testing Done:

Compiled ONIE for the kvm_x86_64 and the fsl_p2020rdbpca machines.
Verified both machines boot and perform as expected.


cbrune:
pushed to master:

- e8710b3 Move default kernel version to 4.1.3

PR #244, opened 2015-07-23
onie-fwpkg: fix shell string interpretation

cbrune:
The string interpretation in Busybox was working fine, however using
Debian's /bin/dash had problems with resulting date string with
embedded spaces.  The busybox string quoting is a little lax compared
to other full shell implementations.

Since this script is intended to work with multiple /bin/sh
environments (Busybox in ONIE, /bin/sh on other Linux distros) we need
to fix up the string quoting to be more robust.

Testing:

- used the script from ONIE's busybox
- used the script on a Debian 7.0 workstation using /bin/dash


cbrune:
pushed to master:

- 3b47eae onie-fwpkg: fix shell string interpretation

Issue #201, opened 2015-05-29
ISO  image still has UEFI boot method when supporting UEFI firmware disabled


david56:
Yesterday, I found the ISO image always combined with BIOS and UEFI booting methods even if ``UEFI_ENABLE = no``.  If the recovery ISO which has linux-3.2.25 boots via ``UEFI: JetFlashTrans...``,  the kernel will have some errors and cause eUSB device disappeared.  ONIE updater can't find the mass storage to install itself eventually.

    ONIE: Rescue Mode ...
    Platform  : x86_64-accton_as5712_54x-r0
    Version   : 2015.5.28-dirty
    Build Date: 2015-05-28T11:16+0800
    ehci_hcd 0000:00:16.0: Found HC with no IRQ.  Check BIOS/PCI 0000:00:16.0 setup!
    ehci_hcd 0000:00:16.0: init 0000:00:16.0 fail, -19
    Info: Mounting kernel filesystems... done.
    Info: Using eth0 MAC address: 00:00:e8:a0:00:00
    Info: eth0:  Checking link... up.
    Info: Trying DHCPv4 on interface: eth0
    ONIE: Using DHCPv4 addr: eth0: 192.168.1.230 / 255.255.255.0
    Starting: dropbear ssh daemon... done.
    Starting: telnetd... done.
    discover: Rescue mode detected.  Installer disabled.
    
    Please press Enter to activate this console.
    To check the install status inspect /var/log/onie.log.
    Try this:  tail -f /var/log/onie.log
    
    ** Rescue Mode Enabled **
    ONIE-RECOVERY:/ # ls /sys/firmware/
    memmap
    ONIE-RECOVERY:/ #

``acpi`` device is also absent.

cbrune:
I don't quite understand from your description what the problem is.

Can you put the console output from the ONIE install process on a gist so we can all look at it?

Also to aid debugging you can add "set -x" near the top of `installer/x86_64/install-arch` to see more output.

david56:
Here you are:

https://gist.github.com/david56/ffedf7ac4d4f32aff0fd

cbrune:
I think I understand the situation, but I do not understand why the 3.2.25 kernel cannot find the USB storage.

[...]

PR #245, opened 2015-07-31
onie-mk-iso.sh: create ISO image without UEFI support if UEFI_ENABLE=no

david56:
This patch has been tested in Accton AS5712_54X including cases with/without UEFI support.


build-config/scripts/onie-mk-iso.sh
```diff
@@ -100,6 +100,11 @@ RECOVERY_EFI_BOOTX86_IMG="$RECOVERY_EFI_BOOT_DIR/bootx64.efi"
 RECOVERY_ELTORITO_IMG="$RECOVERY_ISO_SYSROOT/boot/eltorito.img"
 RECOVERY_EMBEDDED_IMG="$RECOVERY_DIR/embedded.img"
 RECOVERY_UEFI_IMG="$RECOVERY_ISO_SYSROOT/boot/efi.img"
+if [ "$UEFI_ENABLE" = "yes" ] ; then
+    RECOVERY_XORRISO_OPTIONS="$RECOVERY_CONF_DIR/xorriso-options-uefi.cfg"
+else
+    RECOVERY_XORRISO_OPTIONS="$RECOVERY_CONF_DIR/xorriso-options-bios.cfg"
+fi
```
cbrune: Instead of duplicating this logic how about passing the xorriso config file as an argument to this script?  Above when calling onie-mk-iso.sh pass $(RECOVERY_XORRISO_OPTIONS) as an additional argument.

cbrune:
Thanks for looking into this!!

Generally looks good.  After fixing the minor issue with the xorriso config file I think this will beready to go.

david56:
Thank you for your guidance!  The patch has been revised as you said.

david56:
The issue was described in Issue #201

cbrune:
Pushed to master:

- 4e78020 onie-mk-iso.sh: create ISO image without UEFI support if UEFI_ENABLE=no

PR #247, opened 2015-08-13
QuantaMesh Common P2020 CPU board

audihsu-qci:
===============================================================================
  QuantaMesh Common P2020 CPU board
===============================================================================

  Supported Platforms
=======================
- T5032-LY6  (cpu module)
   32x40GbE QSFP+ ports

- T3048-LY8  (cpu module)
   48x1/10GbE SFP+ ports,
   front 4x40GbE QSFP+ ports, and rear 2x40GbE QSFP+ ports

- T3048-LY5A (cpu module)
   48x1/10GbE SFP+ ports and 12x40GbE QSFP+ ports

- T3048-LY9  (cpu module)
   48x100/1000/10GBASE-T,
   front 4x40GbE QSFP+ ports, and rear 2x40GbE QSFP+ ports

- T3048-LY9A (cpu module)
   48x100/1000/10GBASE-T and 6x40GbE QSFP+ ports

- T5032-LY6F (cpu module)
   32x40GbE QSFP+ ports

- T3024-P05  (cpu on board)
   24x1/10GbE SFP+ ports and 2x40GbE QSFP+ ports

- T3024-P05A (cpu on board)
   24x1/10GbE SFP+ ports

 - For more detail product information, please refer to
   http://www.quantaqct.com/

  Specification
===============
 - Modularized CPU board design
 - 1 RJ-45 out-of-band management port (10/100/1000M)
 - 1 RJ-45 console port
 - 1 micro SD card
 - 1 USB 2.0 port

  Special Feature
=================
 - ONIE recovery mechanism

For further information, please take a look at ``machine/quanta/quanta_common_p2020/INSTALL``


cbrune:
Generally this looks like a very nice commit.  I have a few questions I'll ask in the files.

One global comment, about the commit message.  The *very* first line of your commit *must* be a single, one-line summary of commit.  See the ONIE project guidelines for patches:

The first line of your commit message is "=============================".  The problem with this is when someone does a short git log all they will see is "==============================", which does not tell them anything.  Or if you sent these patch via email, the email subject line would be "=================================".

The first line needs to be a summary.  For example "Add support for QuantaMesh Common P2020 CPU board" would be perfectly fine.

machine/quanta/quanta_common_p2020/INSTALL
```diff
@@ -0,0 +1,165 @@
...
+
+To compile demo installer.
+
+  $ cd machine/quanta/quanta_common_p2020
+  $ ./build.sh demo
+
+To compile html and pdf documents.
+
+  $ cd machine/quanta/quanta_common_p2020
+  $ ./build.sh docs
```
cbrune: I doubt if this is true any more.  The ONIE project docs are now kept in the project wiki git repo.

machine/quanta/quanta_common_p2020/INSTALL
```diff
@@ -0,0 +1,165 @@
...
+  => run onie_rescue
+
+
+Updating the ONIE binaries
+==========================
+
+An easier way to update ONIE within u-boot
+
+  => setenv ipaddr 192.168.2.1
+  => copy -o tftp://192.168.2.101/onie-quanta_common_p2020-r0.bin
```
cbrune: Does this new `copy` command simply streamline the above actions from step 3?  It seems like a nice addition.

machine/quanta/quanta_common_p2020/installer/install-platform
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+# Overrides install.sh
+check_machine_image()
+{
+    platform=$(/bin/sed 's/quanta,//g' /proc/device-tree/compatible)
+    image_platform_machine=$(echo $image_machine | sed "s/common_p2020/${platform}/g")
+    echo "ONIE: Platform Machine: $image_platform_machine"
```
cbrune: Having read the entire patch I *think* I see what you are doing here.

Correct me if I'm wrong, but it looks like you have modified u-boot to "fix up" the device tree compatible and model names when booting OINE.  You do this by detecting the type of base board by reading a CPLD register?

If so that is kind of neat.  It is also new behavior that should be documented.  That documentation should go into the commit message and the INSTALL file.

machine/quanta/quanta_common_p2020/kernel/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,415 @@
...
++ * Called very early, device-tree isn't unflattened
++ */
++static int __init quanta_common_p2020_probe(void)
++{
++	unsigned long root = of_get_flat_dt_root();
++	const char *compat_prefix = "quanta";
++	char *compat = of_get_flat_dt_prop(root, "compatible", NULL);
++
++	if(memcmp(compat, compat_prefix, strlen(compat_prefix)) == 0)
++		return 1;
```
cbrune: I presume you only match on the prefix "quanta" here to handle the rewriting of the platform name that comes after "quanta," ?

machine/quanta/quanta_common_p2020/rootconf/sysroot-init/machine-detect.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+platform=$(/bin/sed 's/quanta,//g' /proc/device-tree/compatible)
+cp -f /etc/machine.conf /etc/machine_common_p2020.conf
+/bin/sed "s/common_p2020/${platform}/g" /etc/machine_common_p2020.conf > /etc/machine.conf
+
```
cbrune: Are there any ways this can fail?  What happens if $platform for some reason is blank?

You have added more moving parts here.  So I'm just trying to think about what could go wrong.

machine/quanta/quanta_common_p2020/u-boot/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,2235 @@
...
++	{ 0x12, "p05a_p2020",  "P05A\0",  "P05A-P2020\0",  0x0000d5a1 },
++};
++
++SysEEPROMPath sysEEPROMPath[] = {
++	{ 0x72, 2, 0x54},
++	{ 0x72, 3, 0x54},
++	{ 0x77, 2, 0x54},
++};
++
++int mb_index_get(void) {
```
cbrune: Code quality problems -- too much reliance on global variables

This function returns an 'int', yet below when called below the result is ignored.  Instead the proper functioning relies on the setting of global variables `sysbdinfo` and `eepromPath`.

This function also uses a global variable `motherboard_id` as an input variable.  Why not just pass that value into the function on the stack?

This function, its callers and related data structures should be redesigned to get rid of the global variables and confusing code paths.

One example would be to include the sysEEPROMPath[] info in the sysBoardInfo[] structure.

machine/quanta/quanta_common_p2020/u-boot/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,2235 @@
...
++	int i;
+ 	printf("EEPROM dump: (0x%x bytes)", TLV_INFO_MAX_LEN);
+ 	for (i = 0; i < TLV_INFO_MAX_LEN; i++) {
+ 		if ((i % 16) == 0)
+@@ -214,7 +215,7 @@ static struct tlv_code_desc tlv_code_list[] = {
+ 	{ TLV_CODE_VENDOR_NAME	 , "Vendor Name"},
+ 	{ TLV_CODE_DIAG_VERSION	 , "Diag Version"},
+ 	{ TLV_CODE_SERVICE_TAG   , "Service Tag"},
+-	{ TLV_CODE_VENDOR_EXT	 , "Vendor Extension"},
++	{ TLV_CODE_VENDOR_EXT	 , "Model Name" /* "Vendor Extension" */},
```
cbrune: This cannot be done.

`TLV_CODE_VENDOR_EXT` cannot simply be renamed and used for a different purpose.  The EEPROM spec is a standard and 3rd parties rely on it being consistent.

If the need arises to define a new TLV then that is separate topic we should discuss on the mailing.  The first thing to do is make absolutely sure none of the existing TLVs will work for the project:
https://github.com/opencomputeproject/onie/wiki/Design-Spec-HW-Requirements#type-code-values

machine/quanta/quanta_common_p2020/u-boot/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,2235 @@
...
+ 	case TLV_CODE_DEVICE_VERSION:
+ 		value = simple_strtoul(strval, NULL, 0);
+ 		if (value >= 256) {
+@@ -633,10 +639,8 @@ static bool tlvinfo_add_tlv(u8 * eeprom, int tcode, char * strval)
+ 		new_tlv_len = 2;
+ 		break;
+ 	case TLV_CODE_MANUF_DATE:
+-		if (set_date(data, strval) != 0) {
++		printf("ERROR: Changing manufacturing date is not permitted here.\n");
+ 			return(FALSE);
```
cbrune: Indentation is off here.  white space problem.

machine/quanta/quanta_common_p2020/u-boot/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,2235 @@
...
+-	int ret;
++	int ret, ret2 = 0;
+ #ifdef CONFIG_SYS_EEPROM_BUS_NUM
+ 	unsigned int bus;
+ #endif
+@@ -977,9 +925,29 @@ int read_sys_eeprom(void *eeprom, int offset, int len)
+ 	bus = i2c_get_bus_num();
+ 	i2c_set_bus_num(CONFIG_SYS_EEPROM_BUS_NUM);
+ #endif
++#ifdef CONFIG_MB_DETECT
```
cbrune: To avoid these global variables this function could call a new function which returns the needed i2c info.

machine/quanta/quanta_common_p2020/u-boot/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,2235 @@
...
++#define CONFIG_SYS_I2C_EEPROM_ADDR_LEN		1
++#define CONFIG_SYS_I2C_EEPROM_INFO_ADDR         0x0
++#define CONFIG_SYS_I2C_EEPROM_INFO_LEN          0x0
++#define CONFIG_SYS_EEPROM_MAX_SIZE		256
++#define CONFIG_SYS_EEPROM_LOAD_ENV_MAC
++#define CONFIG_SYS_EEPROM_USE_COMMON_I2C_IO
++#define CONFIG_SYS_EEPROM_PAGE_WRITE_BITS	3
++#define CONFIG_SYS_EEPROM_MAX_NUM_ETH_PORTS	1
++
++#if 0
```
cbrune: please remove dead code if it is not needed

machine/quanta/quanta_common_p2020/u-boot/platform-quanta-common_p2020.patch
```diff
@@ -0,0 +1,2235 @@
...
+ 
+ /*
++ * onie_preboot changes fdt compatible and model to quanta,${motherboard_platform}
++ */
++#define CONFIG_ONIE_PREBOOT \
++        "dtbaddr=0x8000000\0"   \
++        "onie_preboot=cp.b $onie_start $loadaddr ${onie_sz.b} && "      \
++        "imxtract $loadaddr dtb $dtbaddr ; fdt addr $dtbaddr ; "        \
++        "fdt set / compatible quanta,${motherboard_platform} ; "        \
++        "fdt set / model quanta,${motherboard_platform}\0"
```
cbrune: What happens if `imxtract` fails?  All the remaining command will still be executed.

It might be better to use `&&` between all the commands instead of `;`.

audihsu-qci:
Hi Curt,
    Thanks for your effort on reviewing this patch.
I shall have a modification of this patch and a full test.
It takes a bit time. Thanks again.

machine/quanta/quanta_common_p2020/INSTALL
audihsu-qci: Yes, it is taken as a macro command of all the update onie steps.
Simple, and also checks the size of download file.

PR #255, opened 2015-08-28
X86 install arch

david56:
fix two issues in x86_64 install-arch


cbrune:
Looks nice.  Thanks!

Pushed to master:

- 25fc86c x86_64 install-arch: create basic diretories for installing and logging
- ec6149a x86_64 install-arch: correct directory names in error messages

PR #258, opened 2015-09-03
Do not pass the MACHINE variable to sub-makes

cbrune:
Do not pass the MACHINE variable to sub-makes or to other
sub-processes through the environment.  Various sub-projects also
define and use a variable called MACHINE for a completely different
purpose.


cbrune:
Pushed to master:

- 3078442 Do not pass the MACHINE variable to sub-makes

PR #260, opened 2015-09-08
Accton AS5512_54X: add to support ONIE

david56:


cbrune:
Looks good.  Thanks!

Pushed to master:

- 6114c12 Accton AS5512_54X: add to support ONIE

PR #261, opened 2015-09-08
Accton AS7512_32X: add to support ONIE

david56:


cbrune:
Thanks for the contribution!  Looks good.

Pushed to master:

- d051fbc Accton AS7512_32X: add to support ONIE

PR #262, opened 2015-09-11
set uboot_max_size in onie-rom.conf

david56:
- Fixed the issue of generating ONIE .bin image

### DIFF
--- a/machine/accton/accton_as6700_32x/onie-rom.conf
+++ b/machine/accton/accton_as6700_32x/onie-rom.conf
@@ -8,6 +8,7 @@ description="Accton, AS6700_32X"
 format=ubootenv_onie
 
 uimage_max_size=$(( 8 * 1024 * 1024 ))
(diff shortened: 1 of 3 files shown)


cbrune:
The changes seems OK.

The commit comments can be quite a bit better though.  For example, why is the increase needed?  What has changed to increase the size of the u-boot image, etc?

david56:
Curt, thanks for your reminder!  I added the ``uboot-max-size`` to onie-rom.conf to meet the commit's (6868614) new feature.  Both u-boot sizes in the machines are over 512 KiB from the beginning.  The commit messages are refreshed.  Please check them again.  Thank you.

cbrune:
Great, thanks for finding that.

Pushed to master:

- 1a824ad Accton AS7710_32X: set uboot_max_size in onie-rom.conf
- 3ae0b85 Accton AS6700_32X: set uboot_max_size in onie-rom.conf

PR #264, opened 2015-09-15
Accton x86_64 platforms: move busybox patch to common folder

david56:
The platforms are including:

- AS5512_54X
- AS5712_54X
- AS6712_32X
- AS7512_32X
- AS7712_32X


cbrune:
Looks great.  It is always nice to remove common code.

Pushed to master:

- b8fcad9 Accton x86_64 platforms: move busybox patch to common folder

PR #265, opened 2015-09-21
add check for partition name

dorkmatt:
ONIE requires diagnostics partitions to follow 3 requirements as specified in [specifications](https://github.com/opencomputeproject/onie/wiki/Design-Spec-x86-NOS-Interface#gpt-partition-table). This patch adds support for verifying the partition name regex, where prior to this - only the file system name regex and attribution flag was verified.

### DIFF
--- a/contrib/oce/test-nos.sh
+++ b/contrib/oce/test-nos.sh
@@ -169,6 +169,15 @@ check_for_diag()
                     else
                         echo "***Diag $p does not have the correct attribute flags"
                     fi
+
+                    # if found, check partition name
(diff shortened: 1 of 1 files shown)


dorkmatt:
Another point to make, in the current implementation - if all 3 requirements are sent, we only print back 2 as valid, see:
```
ONIE:/ # ./test-nos.sh
Diag /dev/sda3 is valid
Partition name 'as6712_32x-DIAG' is valid
```
Suggest if each one is valid, don't print anything (unless in verbose mode).

cobracmder:
did another PR #266 that took in your verbose request

PR #268, opened 2015-09-30
Because of switch default console speed of BIOS is 57600 now, modify …

vic1231:
…console speed to 57600 for unify.


cbrune:
The commit message for b2111f47a7d18c6122d76b43a4eac7bea1f88f52 could be a little longer.  Can you please elucidate the need for the change or describe the "i2c chip feature" a bit more?  It really helps those who come after you a year from now understand what was done and why.

vic1231:
About the console speed, with your advise,  we will try to change other section to 115200.
So now is no necessary to change console speed.
And for i2c access rule modify, I will make it more clear.

cbrune:
OK, sounds good.  Can you make a new pull request that only contains the i2c chip feature with the updated commit message?

PR #270, opened 2015-10-07
If write over 16 bytes again, it will override with from first byte. …

vic1231:
…Ex, when write 65 byte once, the 65th byte will write to 1st byte address.

Seems it will always write at (n%64) byte, where n is number of bytes to write, and happened with n>64.
Do not know what happened with usb to i2c chip now.
After force try, get a number 16, it will clear overwrite issue, 8 seems OK, too.


cbrune:
Thanks!  Pushed to master:

- 98af6b8 wnc_sst1_n1: fix i2c usb access

PR #272, opened 2015-10-08
Two Inventec  platform support .

vicye:
Inventec D6254QS: add to support ONIE
Inventec DCS7032Q28: add to support ONIE


cbrune:
All of my comments about the D6254QS apply to the DCS7032Q28 as well.

I also suggest you put a little more into the commit message, describing what the machines are.  For example your marketing department would probably like you to include a description of the speeds and feeds for these boxes.

cbrune:
Regarding the kernel patches.

It is very important to keep track of the provenience of back ported patches.  The commit message must include where the patch came from, i.e where did you get the patches?   We need to document the what kernel version the patches came from.

kernel-ich-gpio.patch looks to be a clean back port of the ich-gpio driver.  This is nice, as it adds support for a driver that the 3.2.69 kernel currently lacks.

Is there anything specific to the D6254QS or DCS7032Q28 in that patch, or is it the generic ich-gpio driver from a later kernel version?  If this is a straight back port, with no specific D6254QS or DCS7032Q28 modifications made then it would be better to include this patch in the 'general' kernel patchset, making it available to all boards.

Instead of putting this patch in onie/machine/inventec/d6254qs/kernel/kernel-ich-gpio.patch, put it in onie/patches/kernel/3.2.69.

That would be better, so there is only 1 copy of the patch in the tree.  That change should be a separate ONIE commit on its own, separate from the D6254QS and DCS7032Q28 machines.

The kernel-igb.patch is a different story.  This patch looks like it is replacing the existing igb driver with a different version.  We can't make that a general patch as it might disrupt existing machines which are using igb already.  It is fine that this patch remain specific to inventech.

For machines that share common kernel patches (which looks to be the case here) the ONIE patch system has an optimization so you only need to carry one copy of the patches.  Instead of placing the kernel-igb.patch patch in `machine/inventec/<machine>/kernel` put it in `machine/inventec/kernel`.  Then all inventech machines will pick up that patch, instead of specifying the patch per machine.

For an example of this technique see https://github.com/opencomputeproject/onie/tree/master/machine/alphanetworks where all machines share a common i2ctools patch directly.  The same technique works for the kernel and u-boot patches.

cbrune:
I just noticed the machine directories are not following the naming convention.  The vendor name must be in the machine directory name.

The directories should be:
```
machine/inventec/inventec_d6254qs
machine/inventec/inventec_dcs7032q28
```

vicye:
I close the pull request , and I'll issue one new pull request later.

PR #277, opened 2015-10-09
Create Debian Jessie ONIE Installer

cbrune:
This patch contains a recipe for creating an ONIE compatible installer
from a Debian Jessie mini.iso image


cbrune:
Pushed to master:

- d2f0ca4 Create Debian Jessie ONIE Installer

PR #279, opened 2015-10-22
Accton x86_64 platform: add management port workaround to kernel driver

david56:
The platforms are including:

- AS5512_54X
- AS7512_32X

### DIFF
--- a/machine/accton/accton_as5512_54x/kernel/series
+++ b/machine/accton/accton_as5512_54x/kernel/series
@@ -1,1 +1,1 @@
-
+driver-igb-bcm54616-workaround.patch
--- a/machine/accton/accton_as7512_32x/kernel/series
+++ b/machine/accton/accton_as7512_32x/kernel/series
@@ -1,1 +1,1 @@
-
+driver-igb-bcm54616-workaround.patch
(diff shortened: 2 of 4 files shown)


cbrune:
@david56 - thanks for sending this along.

Please update the commit message to clearly state what problem this work around is fixing.

Is this something all NOS will need to do for AS5512_54X and AS7512_32X?

Is this something that is fixed upstream in more recent kernel version?

cbrune:
I noticed these machines also include this patch: AS5812_54T, AS5812_54X, AS6812_32X

Is this work around specific to your CPU module?  Or is the work around of a general nature, and should be promoted to the top level and applied for all ONIE machines?

If this is specific to Accton's CPU module, then a better approach would be program the Accton management port EEPROM with any necessary fixups instead of modifying the kernel driver.

The patch as it stands would never be accepted by an upstream Linux kernel maintainer.

david56:
The commit message has been refreshed.

I’m not sure whether it is a common issue or not.  So I put the patch in Accton’s machine directory.  We are doing more test on the other machines to decide what machines need to have the patch.  So there may have more machines applying the patch.

Issue #250, opened 2015-08-21
x86_64: grub may become invalid if power off during making install boot mode sticky


david56:
In recent ONIE install mode, it always reinstall grub and set boot mode to `install` when auto discover starting.  We found it may make grub become invalid if power off suddenly.

```
Version   : 2015.6.9
Build Date: 2015-06-09T15:55+0800E: OS Install Mode ...
Info: Mounting kernel filesystems... done.
scsi 6:0:0:0: Direct-Access     ATP      ATP IG eUSB      1100 PQ: 0 ANSI: 4
sd 6:0:0:0: [sda] 15724544 512-byte logical blocks: (8.05 GB/7.49 GiB)
sd 6:0:0:0: [sda] Write Protect is off
sd 6:0:0:0: [sda] No Caching mode page present
sd 6:0:0:0: [sda] Assuming drive cache: write through
sd 6:0:0:0: [sda] No Caching mode page present
sd 6:0:0:0: [sda] Assuming drive cache: write through
sd 6:0:0:0: [sda] No Caching mode page present
sd 6:0:0:0: [sda] Assuming drive cache: write through
sd 6:0:0:0: [sda] Attached SCSI disk
scsi 7:0:0:0: Direct-Access     JetFlash Transcend 8GB    8.07 PQ: 0 ANSI: 4
sd 7:0:0:0: [sdb] 15261696 512-byte logical blocks: (7.81 GB/7.27 GiB)
sd 7:0:0:0: [sdb] Write Protect is off
sd 7:0:0:0: [sdb] Write cache: disabled, read cache: enabled, doesn't support DPO or FUA
sd 7:0:0:0: [sdb] Attached SCSI removable disk
Info: Mounting ONIE-BOOT on /mnt/onie-boot ...
Info: Using eth0 MAC address: 00:00:e8:a0:00:00
Info: eth0:  Checking link... up.
Info: Trying DHCPv4 on interface: eth0
ONIE: Using DHCPv4 addr: eth0: 192.168.1.230 / 255.255.255.0
Starting: dropbear ssh daemon... done.
Starting: telnetd... done.
Installing for i386-pc platform.
Installation finished. No error reported.
(power off)
```

Power on again, the grub becomes invalid.

```
error: file `/grub/i386-pc/normal.mod' not found.
Entering rescue mode...
grub rescue>

error: file `/grub/i386-pc/extcmd.mod' not found.
Entering rescue mode...
grub rescue>
```

[...]

PR #281, opened 2015-10-22
install mode: sync data to storage after making install boot mode sticky

david56:
Force to write data to storage after making install boot mode sticky.
It will prevent grub to be invalid after entering linux shell and power
off suddenly.

The issue was reported in
https://github.com/opencomputeproject/onie/issues/250

This patch has been tested in Accton AS5512_54X.

### DIFF
--- a/rootconf/x86_64/sysroot-lib-onie/boot-mode-arch
+++ b/rootconf/x86_64/sysroot-lib-onie/boot-mode-arch
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 #  Copyright (C) 2014-2015 Curt Brune <curt@cumulusnetworks.com>
+#  Copyright (C) 2015 david_yang <david_yang@accton.com>
 #
 #  SPDX-License-Identifier:     GPL-2.0
 
@@ -26,6 +27,7 @@ install_remain_sticky_arch()
     else
         bios_boot_onie_install
     fi
+    sync;sync
 
     return 0
 }


rootconf/default/etc/init.d/discover.sh
```diff
@@ -68,6 +69,7 @@ do_start() {
             install_remain_sticky_arch || {
                 echo "Error: problems making install boot mode sticky" > /dev/console
             }
+            sync;sync
```
cbrune: Nice, I am glad this helps #250.

Instead of fixing it here, however, I would make this change in rootconf/x86_64/sysroot-lib-onie/boot-mode-arch.  This problem only effects x86, not powerpc or arm.

david56: @cbrune 
For powerpc or arm,  it has stream buffer for MTD devices.  ONIE makes install boot mode sticky by writing variables to uboot-env.  So I think it should effect powerpc or arm, too.

cbrune: I don't think so.  u-boot env variables are kept in a mtd device, not a mtdblock device.  sync applies to *filesystems*.  There is no filesystem on a mtd device.

david56: > I don't think so. u-boot env variables are kept in a mtd device, not a mtdblock device. sync applies to filesystems. There is no filesystem on a mtd device.

I see.  I have revised the patch.  Thanks.

cbrune:
Looks great, thanks!

@ChangeTip, send some bits to @david56!

Pushed to master:
- ada8cb8 install mode: sync data to storage after making install boot mode sticky

changetip:
Hi @david56, @cbrune sent you a tip worth 1 bits, and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/910187)**.



rashed0:
http://rashed0.tip.me/

rashed0:
http://rashed0.tip.me/

rashed0:
Help Me Please 
http://rashed0.tip.me/

PR #282, opened 2015-10-24
regression: IPv6 neighbor discovery broken

cobracmder:
It looks like commit: https://github.com/opencomputeproject/onie/commit/a02b948cec3eaa6fcd7ea800d667d4afacfbbcf3 is the offending commit.

In particular, this line: https://github.com/opencomputeproject/onie/blob/master/rootconf/default/etc/init.d/networking.sh#L87

When an interface is flushed, all addresses are removed (including IPv6 link local).  In order for an IPv6 to come back, the interface must be downed and then brought back up.  In order for IPv6 neighbor discovery to work, ONIE must have received an IP address via:
* static from command line (GRUB or uboot)
* DHCPv4

For resolution, currently looking at ways to remove all non-link local addresses to resolve this bug.

### DIFF
--- a/rootconf/default/etc/init.d/networking.sh
+++ b/rootconf/default/etc/init.d/networking.sh
@@ -83,8 +83,8 @@ config_ethmgmt_fallback()
     local intf=$1
     shift
 
-    # Remove any previously configured IP address
-    ip addr flush $intf
+    # Remove any previously configured, IPv4 addresses
+    ip -f inet addr flush dev $intf
 
     # Assign sequential static IP to each detected interface
     local interface_base_ip=$(( $base_ip + $intf_counter ))


cbrune:
Nice.  Thanks.

@ChangeTip, send some bits to @cobracmder !

Pushed to master:
- 68583f5 network config: flush all IPv4 addresses

changetip:
Hi @cobracmder, @cbrune sent you a tip worth 1 bits, and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/910265)**.



rashed0:
Hello
Please Help Me On
https://www.changetip.com/tipme/rashed0

rashed0:
https://www.changetip.com/tipme/rashed0

PR #286, opened 2015-10-29
workaround solution for halt issue on Inventec D6254QS and DCS7032Q28…

vicye:
On Inventec D6254QS and DCS7302Q28 platform, if "halt" , "poweroff" command are issued, DUT cannot power on.  we give the workaround soutiion as follows:

halt , poweroff command  behaves like  "reboot" command

### DIFF
--- a/machine/inventec/inventec_d6254qs/busybox/patches/series
+++ b/machine/inventec/inventec_d6254qs/busybox/patches/series
@@ -1,1 +1,2 @@
 onie-syseeprom-write-delay.patch
+workaround-halt-issue.patch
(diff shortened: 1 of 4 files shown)


cbrune:
You shed some light on the issue in the PR message, but the commit message itself has no details.

Please update the commit message to include details about the problem and why is the work around needed.

Is this a hardware (board design) problem that you plan to fix in subsequent revisions of the hardware?  This is *very* bad and something every NOS vendor will have to work around.  Customers will be surprised when "poweroff" does not power off, but reboots instead.  How will people quiesce the disks, say for moving the switch from one rack to another?

On a personal level I have chased this problem with at least two other hardware vendors -- it will drive your technical support and customers crazy as the problem will brick the box.  In the end, the hardware vendors in question fixed their hardware (either board revision, CPLD revision or BIOS revision) to make "poweroff" work correctly.  That is by far the best solution and will say you agony down the road.

I strongly encourage you to fix the hardware.

vicye:
Thanks for your advice.  After discussing with my colleagues, we decided to resolve the issue from BIOS. 
Therefore, the pull request  is closed.

cbrune:
Thank you!!  That will save everyone a lot of trouble.

PR #289, opened 2015-10-29
Accton WEDGE_16X: add to support ONIE

david56:
The machine has 16x40GbE QSFP ports on Broadcom Trident2 switch
ASIC.  CPU module is Intel Avoton C2750.


cbrune:
Nice, thanks for the contribution!

Pushed to master:

- b021777 Accton WEDGE_16X: add to support ONIE

PR #293, opened 2015-11-10
add resize2fs to filesystem

mrhpearson:
Mostly I'm just seeing if I can do a pull request....but we do need this change in ONIE so hopefully this all works :)
We need resize2fs to support our installer on the platform so added this change. I couldn't find a way to make it platform specific but I figure it might be valid on other platforms as well. 

Thanks
Mark


cbrune:
Looks good, pushed to master:

- ca5ee38 add resize2fs to filesystem

@ChangeTip, send some bits to @mrhpearson

changetip:
Hi @mrhpearson, @cbrune sent you a tip worth 1 bits, and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/918010)**.

PR #294, opened 2015-11-17
Accton WEDGE_16X: correct script finding install device

david56:


cbrune:
Pushed to master:

- f8b857d Accton WEDGE_16X: correct script finding install device

PR #295, opened 2015-11-19
Add option for video console and keyboard

cbrune:
Add a build option to allow a machine to use a standard video console and attached keyboard, instead of using the serial console.

This option makes the kvm_x86_64 virtual machine easier for new users to start using quickly.


cbrune:
Pushed to master:

- e6b52c6 kvm_x86_64: add keyboard support
- 086c560 x86_64: vga console support

sspans:
Nice work, that really helps.

cbrune:
@sspans - great, glad to hear that helps!

PR #300, opened 2015-11-30
kexec-tools: back port upstream patch for better device tree handling

cbrune:
Back port upstream patch for better device tree handling:
https://git.kernel.org/cgit/utils/kernel/kexec/kexec-tools.git/commit/?id=0e1030b345b9e7c1c64b85dde32a282a19b825f0

With this patch in place for PowerPC systems one can kexec without specifying a .dtb file, simply reusing the device tree from the running ONIE kernel.


cbrune:
pushed to master:

- cb2d0c4 kexec-tools: add patch for better device tree support
- fe7e1fa kexec-tools: enable patching of the kexec-tools package

PR #302, opened 2015-12-03
grub: reduce time took during making install boot mode sticky

david56:
The modules installed by grub during making install boot mode sticky
are the same.  The Modules are installed two times.  It takes 5 ~ 9
seconds to finish installing grub.  By the new way, instead of
re-installing modules, backup and restored modules take around one
second.

It also reduces the probability of grub corrupted by power cycle.
Please see commit ada8cb8.

The issue was reported in
https://github.com/opencomputeproject/onie/issues/250

This patch has been tested in Accton AS7712_32X.

### DIFF
--- a/build-config/make/demo.make
+++ b/build-config/make/demo.make
@@ -51,7 +51,9 @@ DEMO_TRIM = \
    etc/rc0.d/K25discover.sh	\
    etc/rc3.d/S50discover.sh	\
    etc/rc6.d/K25discover.sh	\
+   etc/rcS.d/S03boot-mode.sh	\
    etc/init.d/discover.sh	\
+   etc/init.d/boot-mode.sh	\
    bin/discover			\
    bin/uninstaller		\
    bin/onie-uninstaller		\
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -156,9 +156,21 @@ demo_install_grub()
     local demo_mnt="$1"
     local blk_dev="$2"
 
+    # get running machine from conf file
+    [ -r /etc/machine.conf ] && . /etc/machine.conf
+
+    if [ "$onie_firmware" = "coreboot" ] ; then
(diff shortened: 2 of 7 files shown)


david56:
The revised codes are in ``bios_boot_onie_install()``.  Should we have another function such as ``core_boot_onie_install()`` to serve coreboot if it has different ways from PC BIOS?
In commit d1bd566, I found grub modules in ``/usr/lib/grub/i386-pc`` and ``/mnt/onie-boot/grub/i386-pc`` are the same.  It seems to copy modules from ``/usr/lib/grub/i386-pc`` is more better.

david56:
> You can't assume the GRUB architecture is i386-pc. It could also be i386-coreboot. See commit d1bd566.
>
> Generally looks good except for i386-pc vs. i386-coreboot. You need to check the $onie_firmware variable to see which GRUB architecture to use.
>
> It looks like there is another coreboot bug (unrelated to your changes) on line 222 for the $core_image variable. Similar coreboot bugs for the $core_image variable also exists in demo/installer/x86_64/install.sh and installer/x86_64/install-arch.

Thanks for your guidance.  Checked the coreboot install directory, it is core.elf rather than core.img.  So I also revised installer for ONIE and demo OS.

Please check the updated patches.

cbrune:
The boot-mode patch looks OK.

It all looks good.  Thanks for taking care of the coreboot cases!

@changetip send @david56 a cpu.

Pushed to master:

- 01a9c5a installer: specify correct core image and grub target for grub-install
- 7d814cf boot-mode: move statements to be a standalone service
- 31e5d80 grub: reduce time took during making install boot mode sticky

changetip:
Hi @david56, @cbrune sent you a tip worth a cpu (3,223 bits/$1.28), and I'm here to deliver it ➔ **[collect your tip](https://www.changetip.com/collect/926417)**.

PR #303, opened 2015-12-04
onie-fwpkg: fix firmware update status output

cbrune:
The update status output function of onie-fwpkg command contains a
typo.  Instead of dumping firmware update results, it tries to dump
*pending* firmware updates.

This patch fixes the typo and correctly dumps the update results.


cbrune:
pushed to master:

- 0042462 onie-fwpkg: fix firmware update status output

PR #305, opened 2015-12-16
grub: refine codes of setting console config and linux cmdline

david56:
It has many duplicate codes in build system, ONIE updater and DEMO
installer.  The patch refines the codes and makes them more clear.
It is backward compatible with present NOS/DIAG installer.

The new ONIE grub file ``/mnt/onie-boot/onie/grub/grub-variables`` can
be imported in installer before generating ONIE/NOS/DIAG grub.cfg.  It
helps to create/export ``GRUB_SERIAL_COMMAND`` and ``GRUB_CMDLINE_LINUX``
for the installer and ``50_onie_grub``.

The patch has been tested in Accton AS7712_32X.


cbrune:
This is a noble effort, thanks for looking into it.  The serial console configuration has become a bit messy.

Looking over the patches they seem OK.  It needs a lot of testing before I will feel 100% confident.

To your question I believe it is reasonable for a NOS to change `GRUB_SERIAL_COMMAND` and `GRUB_CMDLINE_LINUX`.  The final customer (NOS owner) may want to use a different serial baud rate for example, which changes both of those variables.

All the testing can be done with the virtual machine, which is nice as reboots are fast.

For testing the things to verify are:

Test Case 1
----------------
- start with blank disk
- install ONIE
- install diag
- install NOS
- uninstall NOS  (things tend to break here)

Test Case 2 (same was 1, without the diag)
----------------------------------------------------------
- start with blank disk
- install ONIE
- install NOS
- uninstall NOS  (things tend to break here)

Test Case 3 (virtual machine only)
----------------------------------------------
- build virtual machine with `SERIAL_CONSOLE_ENABLE=no`
- install ONIE
- install NOS
- uninstall NOS  (things tend to break here)

david56:
All the test cases have passed.

BTW, when I built kvm_x86_64 machine, it showed the messages:

    stg import: no 'series' file found in /home/david_yang/onie/build/kvm_x86_64-r0/kernel/patch/series
    make: *** [/home/david_yang/onie/build/kvm_x86_64-r0/stamp/kernel-patch] Error 2

It seems that the series can not be an empty file.  Adding comment at the first line will fix the issue.

Furthermore, I found the console configuration in ONIE ( https://github.com/opencomputeproject/onie/blob/master/installer/x86_64/grub.d/50_onie_grub#L98-L108 ) and DEMO ( https://github.com/opencomputeproject/onie/blob/master/demo/installer/x86_64/install.sh#L342-L349 ) are inconsistent.  When ``SERIAL_CONSOLE_ENABLE=no``, ONIE does not set the console configuration.  But DEMO has.  Which one is correct/better?

cbrune:
@david56 -- for your stg error.  Something else must be going wrong for you as empty series files are used all the time and work fine.  No need to add a comment to the top.

cbrune:
@david56 - for the SERIAL_CONSOLE_ENABLE question.  The apparent inconsistency was done on purpose.  They are both correct in my opinion.

For the case of the installer (the one with the `grep ttyS`) I was concerned about backward compatibility with existing NOS deployments.  For example consider the case of a old box, with an older ONIE and an installed NOS.  The HW vendor decides to push out an ONIE upgrade for the platform, which includes support for the `SERIAL_CONSOLE_ENABLE` mechanism.

In order for that to work the old NOS needs to set `SERIAL_CONSOLE_ENABLE` to 'yes' or 'no'.  But the NOS is old, created before the `SERIAL_CONOSLE_ENABLE` feature existed.  So what to do?  The NOS is not going to change.

Admittedly I'm guessing a bit, but I figured if the NOS puts anything about `ttyS` in its kernel command line then its intent is to use a serial console.

For the DEMO case, the ONIE project owns both sides -- the ONIE side and the NOS side.  We can modify the DEMO to be *explicit* about its intent regarding the `SERIAL_CONSOLE_ENABLE` feature.

david56:
I see.  I have removed the commit of changing ``series``

cbrune:

- 3a3c5d7 grub: refine codes of setting console config and linux cmdline

Issue #174, opened 2015-04-15
Faster/Obvious/Easier abort of ONIE install on X86


cbrune:
Posting this issue on someone else's behalf.

Quoting an issue reported from the field:

```
While playing with GRUB on X86, I discovered a couple of unpleasant things, when combined cause a real headache/nightmare:

1) After selecting 'ONIE' in GRUB, it defaults to ONIE Install mode and before you can read all the text the first time (i.e. 4 seconds), it executes ONIE installer

2) ONIE installer loads. Then there isn't a fast easy way (at least no instructions on the screen...anxiety rising..) if any exists, to abort the autoloading/install of the default image via the waterfall (onie-installer-x86_64 found very quickly via the local webserver)

3) While unsuccessfully trying to stop the ONIE install, you watch with horror as ONIE downloads and installs a new OS, overwriting the existing setup. Again, no way I could figure out how to stop that either....

Net result, you have whatever default image is configured on your webserver, and all vestiges of the config and license on your functioning switch has now been erased and must be restored from backup.. you did create a backup.. right?

I would propose that in GRUB, when you go to the ONIE page, that GRUB not be set to autoexecute ONIE Installer in 3 seconds.

I would further propose that an abort sequence be accessible and printed to the screen when ONIE Install mode starts that would allow halting the install for at least 10 seconds after ONIE loads.
```

mandeepsandhu:
+1 to this. I have seen this happen in our lab (thankfully it was only a test setup).

Maybe we can default to the _least-destructive_ `Rescue` mode for the ONIE GRUB menu once it's installed on the system?

Also, once the installer has started, there's no reliable way to stop it right? The installer might be changing the partition table and an abort mid-way might render the box un-bootable.

Or do you mean to say that ONIE would give around 10 secs _before_ executing the NOS installer? This would be definitely helpful as it gives us a window to abort the install (in which case it drops to `ONIE` shell?)

cbrune:
The poster makes some valid points.

### Default to `rescue` instead of `install`

I think it *may* be possible to default  to 'rescue' mode.  I say *may* because the default behaviour out of the box from the factory absolutely has to be 'install' mode.  That is the key to the entire automation story.

Now for what happens when a user is manually going through the GRUB menus is a different story.

I can't say off the top of my head, but it seems like we could:

- change the order of the menu items, putting `rescue` first
- at ONIE install time (not OS install time) make sure `install` is set in the grubenv
[...]

PR #308, opened 2016-01-05
ONIE GRUB support u-boot-like boot commands

david56:
The patchset intends to make ONIE GRUB:

1. Support u-boot-like boot commands that are useful for auto test programs.
2. Support boot command feeded by DIAG/NOS installer that can easily chage boot target between ONIE, DIAG and NOS.

The patchset is compatible with present DIAG/NOS installers.

### DIFF
--- a/build-config/scripts/onie-mk-installer.sh
+++ b/build-config/scripts/onie-mk-installer.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 #  Copyright (C) 2013,2014,2015 Curt Brune <curt@cumulusnetworks.com>
-#  Copyright (C) 2014-2015 david_yang <david_yang@accton.com>
+#  Copyright (C) 2014,2015,2016 david_yang <david_yang@accton.com>
 #  Copyright (C) 2014 Mandeep Sandhu <mandeep.sandhu@cyaninc.com>
 #
 #  SPDX-License-Identifier:     GPL-2.0
@@ -147,11 +147,7 @@ export GRUB_CMDLINE_LINUX
 
 # variables for ONIE itself
 GRUB_ONIE_SERIAL_COMMAND=\$GRUB_SERIAL_COMMAND
-DEFAULT_CMDLINE="\$GRUB_CMDLINE_LINUX \$GRUB_CMDLINE_LINUX_DEFAULT \$GRUB_ONIE_PLATFORM_ARGS \$GRUB_ONIE_DEBUG_ARGS"
-GRUB_ONIE_CMDLINE_LINUX=\${GRUB_ONIE_CMDLINE_LINUX:-"\$DEFAULT_CMDLINE"}
-ONIE_CMDLINE="quiet \$GRUB_ONIE_CMDLINE_LINUX \\\$ONIE_EXTRA_CMDLINE_LINUX"
 export GRUB_ONIE_SERIAL_COMMAND
-export ONIE_CMDLINE
 
 ## End grub-variables
 EOF
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -313,22 +313,8 @@ cp demo.vmlinuz demo.initrd $demo_mnt
 # store installation log in demo file system
 onie-support $demo_mnt
 
-if [ "$firmware" = "uefi" ] ; then
-    demo_install_uefi_grub "$demo_mnt" "$blk_dev"
-else
-    demo_install_grub "$demo_mnt" "$blk_dev"
-fi
-
 # The persistent ONIE directory location
 onie_root_dir=/mnt/onie-boot/onie
-
-# Create a minimal grub.cfg that allows for:
-#   - configure the serial console
-#   - allows for grub-reboot to work
-#   - a menu entry for the DEMO OS
-#   - menu entries for ONIE
-
-grub_cfg=$(mktemp)
 
 # Set a few GRUB_xxx environment variables that will be picked up and
 # used by the 50_onie_grub script.  This is similiar to what an OS
@@ -342,8 +328,45 @@ grub_cfg=$(mktemp)
 # import console config and linux cmdline
 . $onie_root_dir/grub/grub-variables
 
-# Add common configuration, like the timeout and serial console.
-cat <<EOF > $grub_cfg
+# If ONIE supports boot command feeding,
+# adds DEMO DIAG bootcmd to ONIE.
+if grep -q 'ONIE_SUPPORT_BOOTCMD_FEEDING' $onie_root_dir/grub.d/50_onie_grub &&
+    [ "$demo_type" = "DIAG" ] ; then
+    cat <<EOF > $onie_root_dir/grub/diag-bootcmd.cfg
+diag_menu="Demo $demo_type"
+function diag_bootcmd {
+  search --no-floppy --label --set=root $demo_volume_label
+  echo    'Loading ONIE Demo $demo_type kernel ...'
+  linux   /demo.vmlinuz $GRUB_CMDLINE_LINUX \$ONIE_EXTRA_CMDLINE_LINUX DEMO_TYPE=$demo_type
+  echo    'Loading ONIE Demo $demo_type initial ramdisk ...'
+  initrd  /demo.initrd
+  boot
+}
+EOF
+
+    # Update ONIE grub configuration -- use the grub fragment provided by the
+    # ONIE distribution.
+    $onie_root_dir/grub.d/50_onie_grub > /dev/null
+
+else
+    # Install a separate GRUB for DEMO DIAG or NOS
+    # that supports GRUB chainload function.
+
+    if [ "$firmware" = "uefi" ] ; then
+        demo_install_uefi_grub "$demo_mnt" "$blk_dev"
+    else
+        demo_install_grub "$demo_mnt" "$blk_dev"
+    fi
+
+    # Create a minimal grub.cfg that allows for:
+    #   - configure the serial console
+    #   - allows for grub-reboot to work
+    #   - a menu entry for the DEMO OS
+    #   - menu entries for ONIE
+    grub_cfg=$(mktemp)
+
+    # Add common configuration, like the timeout and serial console.
+    cat <<EOF > $grub_cfg
 $GRUB_SERIAL_COMMAND
 terminal_input $GRUB_TERMINAL_INPUT
 terminal_output $GRUB_TERMINAL_OUTPUT
@@ -352,13 +375,13 @@ set timeout=5
 
 EOF
 
-# Add any platform specific kernel command line arguments.  This sets
-# the $ONIE_EXTRA_CMDLINE_LINUX variable referenced above in
-# $GRUB_CMDLINE_LINUX.
-cat $onie_root_dir/grub/grub-extra.cfg >> $grub_cfg
+    # Add any platform specific kernel command line arguments.  This sets
+    # the $ONIE_EXTRA_CMDLINE_LINUX variable referenced above in
+    # $GRUB_CMDLINE_LINUX.
+    cat $onie_root_dir/grub/grub-extra.cfg >> $grub_cfg
 
-# Add the logic to support grub-reboot
-cat <<EOF >> $grub_cfg
+    # Add the logic to support grub-reboot
+    cat <<EOF >> $grub_cfg
 if [ -s \$prefix/grubenv ]; then
   load_env
 fi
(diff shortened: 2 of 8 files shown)


cbrune:
Hi @david56,

This is a big change and will take some time to digest.  I can't just read it and figure it all out.

I will spend some time this week loading this change on a box and try it out.

Have you tried running the OCE test suite?

@cobracmder - it would be nice if an additional set of eyes also kicked the tires on this one.

cobracmder:
@cbrune copy that

david56:
@cbrune I haven't done it yet. But I did several test cases combined with demo OS, demo DIAG, ONIE (install, rescue, uninstall, update, embed, and recovery ISO), onie-boot-mode, PC BIOS, and UEFI.

cobracmder:
Hey @david56,

So these changes look interesting as they get closer to a UEFI type boot manager in ONIE. (FYI, if you can't tell, I'm a fan of UEFI and how it handles OS boot management.)

After running some OCE tests in KVM and on a 5712, I don't see any issues except when I run ```test-nos.sh``` with a production NOS (not demo OS) and ONL.

In order to use these new features, it requires NOS vendors to update their installers.  This can be done at their discretion for new versions.  But if a user is running a non-updated installer/installation and updates ONIE that includes these changes; they do not see the added benefit.

What are your thoughts on how to handle this case?

david56:
Hello @cobracmder,

Thanks for running tests on the patchset!

UEFI boot manager is really useful for changing the boot sequence between partitions.  I also like it.  The patchset enhances ONIE managing boot order on PC BIOS and UEFI that makes the both boot management more closer.

> In order to use these new features, it requires NOS vendors to update their installers. This can be done at their discretion for new versions. But if a user is running a non-updated installer/installation and updates ONIE that includes these changes; they do not see the added benefit.
> 
> What are your thoughts on how to handle this case?

Yes, you are right.  It is necessary to revise NOS installer to support the new features.

In current ONIE design, NOS installer has to install a copy of GRUB on its own partition for chainloading.  It also has to set NOS partition to be the first boot order in the UEFI situation.  ONIE can not stop the operations from installation.  So that it will go for the present boot approach naturally.

cbrune:
Hello @david56 -

There are quite a few proposed changes here and it has taken me some time to test them out and get my head wrapped around them.  What I did was compare the new behaviours to the old behaviours using the a 2015.11 version.  I think I understand most of what you are doing here.

My following comments are only about the legacy BIOS implementation.  I did not test or look at the UEFI implementation yet.

Definitely there are some great ideas here.

For example I really like the onie_* family of GRUB functions.  It is nice to type `onie_rescue` from the GRUB prompt to boot into rescue mode.  I can see how that would be handy for automated testing -- I agree the GRUB menu is *painful* to deal with from automation.

I also think the changes for the DEMO-DIAG-OS are pretty good.  I can see the `diag_bootcmd` function as also being very useful from an automated testing perspective.  As the hardware vendor, you are responsible for both the ONIE implementation and the DIAG implementation.  Of course as part of that you need to test those components.  I am in favor of anything that helps you test your components.

Also I think it is fine that the DEMO-DIAG-OS reuse ONIE's GRUB -- it does not really need to install it's own GRUB.  Since the hardware vendor owns both ONIE and DIAG I see no problem there.

The changes to the DEMO-OS, however, go to far I think.  We need to maintain a clear separation between ONIE and the NOS.  The DEMO-OS is supposed to instruct NOS vendors "how to build an ONIE-compatible NOS installer".  The DEMO-OS is instructional.

The DEMO-OS *must* install and use its own GRUB, separate from ONIE.  That is the model.  That separation gives some freedom to the NOS vendor about how to build and integrate their OS.  Also most NOS do not simply install one GRUB menu entry (it would not be covered a single nos_menu variable).

I do not think we can predict exactly what each NOS vendor will do, so we should leave that part a bit open ended.

I understand that does not help your automated testing, but real life is like that.

Of course, you are free to implement your own DEMO OS for your own testing purposes.  It is just that the DEMO-OS that is in the upstream ONIE repo needs to be more generic.

With all that said -- cutting over to UEFI should make this easier.  No longer will the various OS's fight over who owns the GRUB.

david56:
@cbrune, 

Thanks for reviewing the patchset.  I'll revise them.

> @mandeepsandhu said:
> +1 to this. I have seen this happen in our lab (thankfully it was only a test setup).
> 
> Maybe we can default to the least-destructive Rescue mode for the ONIE GRUB menu once it's installed on the system?

I noticed that Issue #174 ever talked about about GRUB.  It seems that the new approach should help it.

> Also most NOS do not simply install one GRUB menu entry (it would not be covered a single nos_menu variable).

About the NOS menu entries, I have further proposal in my head these days.  We can preserve several boot commands such as ``nos_bootcmd``, ``nos_bootcmd1``, ..., and ``nos_bootcmd9`` that would be similar to the role of ``ethaddr``, ``eth1addr``, ... in u-boot.

The new install approach *does not* force NOS installer to follow.  It just provides another optional usage for using.  With the new approach, NOS installer has a few various applications of GRUB.

- Present approach -- NOS *must* install and use its own GRUB.  NOS's GRUB is the first boot order.

It looks like:

    the 1st page:
    +---------------------------------------------+
    |*NOS menu 1         (boot NOS directly)      |
    | NOS menu 2         (boot NOS directly)      |
    | NOS menu 3         (boot NOS directly)      |
    | NOS menu 4         (boot NOS directly)      |
    | ONIE               (chainload to ONIE page) |
    |                                             |
    +---------------------------------------------+
    the 2nd page:
    +---------------------------------------------+
    |*ONIE: Install OS   (boot ONIE directly)     |
    | ONIE: Rescue       (boot ONIE directly)     |
    | ONIE: Uninstall OS (boot ONIE directly)     |
    | ONIE: Update ONIE  (boot ONIE directly)     |
    | ONIE: Embed ONIE   (boot ONIE directly)     |
    | DIAG menu          (chainload to DIAG page) |
    |                                             |
    +---------------------------------------------+
    the 3rd page:
    +---------------------------------------------+
    | DIAG menu          (boot DIAG directly)     |
    |*ONIE               (chainload to ONIE page) |
    |                                             |
    +---------------------------------------------+

- New approach -- NOS reuse ONIE's GRUB.  The GRUB has NOS boot commands.  ONIE's GRUB is the first boot order.

It looks like:

    the 1st page:
    +---------------------------------------------+
    |*NOS menu 1          (boot NOS directly)     |
    | NOS menu 2          (boot NOS directly)     |
    | NOS menu 3          (boot NOS directly)     |
    | NOS menu 4          (boot NOS directly)     |
    | ONIE: Install OS    (boot ONIE directly)    |
    | ONIE: Rescue        (boot ONIE directly)    |
    | ONIE: Uninstall OS  (boot ONIE directly)    |
    | ONIE: Update ONIE   (boot ONIE directly)    |
    | ONIE: Embed ONIE    (boot ONIE directly)    |
    | DIAG menu           (boot DIAG directly)    |
    |                                             |
    +---------------------------------------------+

- Hybrid approach -- NOS installs and uses its own GRUB.  NOS menu entry in ONIE page would chainload to NOS page.  ONIE's GRUB also has NOS boot commands that can boot NOS directly.  (NOS vendors ever asked us to add NOS menu entry in ONIE page that can for user go back to NOS page manually.)  This approach still leaves that part (you mentioned above) a bit open ended.

It looks like:

    the 1st page:
    +---------------------------------------------+
    |*NOS menu 1          (boot NOS directly)     |
    | NOS menu 2          (boot NOS directly)     |
    | NOS menu 3          (boot NOS directly)     |
    | NOS menu 4          (boot NOS directly)     |
    | ONIE                (chainload to ONIE page)|
    |                                             |
    +---------------------------------------------+
    the 2nd page:
    +---------------------------------------------+
    |*NOS menu            (chainload to NOS page) |
    | ONIE: Install OS    (boot ONIE directly)    |
    | ONIE: Rescue        (boot ONIE directly)    |
    | ONIE: Uninstall OS  (boot ONIE directly)    |
    | ONIE: Update ONIE   (boot ONIE directly)    |
    | ONIE: Embed ONIE    (boot ONIE directly)    |
    | DIAG menu           (boot DIAG directly)    |
    |                                             |
    +---------------------------------------------+

The later two approaches listed above also works find with UEFI case.  The NOS venders can choose one of them to fit thier requirement in real life.

Just some additional opinions.  How do you think?

cbrune:
I hear what you are saying.  Thanks a lot for putting up these mock-ups of the GRUB menus.  It really helps to understand the ideas.

For the legacy BIOS systems I am very firm about the NOS installing and using its own GRUB.  ONIE is supposed to be dumb and simple.  Anything we do to make it "flexible" also makes it more complex, which leads to more bugs, a larger testing burden (for the HW vendors) and more user documentation and support.

The "new approach" described above I cannot support.  The problems are:
* all HW vendors (and ONIE certification) have to test *all* of the possibilities.  Some OS vendors may choose to install their own GRUB, some might not.  Doesn't matter to the HW vendor, they have to test *all* of the methods.  More work for all HW vendors!
* OS vendors would have choices to make.  How best to instruct them on which choice to make?  Which way should they build their installer, install their own GRUB or not?  More documentation.  More technical support.  More testing.

The "hybrid approach" described above is interesting.  It adds a little bit more complexity, but not much.  It is just as complex as the DIAG OS, which is already working pretty well.  Also it adds a very nice feature, having the "NOS menu" appear on the ONIE GRUB menu.  I think this is worth exploring.

[side note] -- it might be worth enhancing the DEMO OS installer to create more than one menu entry, more accurately simulating real life OS menus.  Perhaps add two entries, with different titles, that boot the same kernel+initrd with identical command line args.

My overall concerns about changing the legacy BIOS GRUB menus are:
* we have something that works today (it could be better I know)
* adding complexity (more testing, more bugs)
* changing existing customer workflows (they understand the current scheme)
* changing existing OS vendor workflows (they understand the current scheme)

I think the place to innovate and make these menus "better" is on the UEFI front.  The UEFI model makes installing multiple OS's easier to manage.  Also for the ONIE project no UEFI machines are shipping yet, so it is an opportunity to apply the lessons learned from the legacy BIOS method and make something better.  I would focus innovating energy on the UEFI model.

david56:
I have revised the patchset.  The support of NOS boot commands and NOS reuses ONIE's GRUB feature are removed.

> According to the documentation, GRUB shell has a case statement which would be appropriate to use here.

I look it up in the document and source code.  I found ``case``, ``esac``, and ``select`` are just *reserved words*.  GRUB does not implement these statements.  So I keep it in ``if ... elif ... fi`` statement.

> Just for consistency I would add a onie_install fucntion.

I add the function and revise it to the menu entry.

> The "hybrid approach" described above is interesting. It adds a little bit more complexity, but not much. It is just as complex as the DIAG OS, which is already working pretty well. Also it adds a very nice feature, having the "NOS menu" appear on the ONIE GRUB menu. I think this is worth exploring.
> 
> [side note] -- it might be worth enhancing the DEMO OS installer to create more than one menu entry, more accurately simulating real life OS menus. Perhaps add two entries, with different titles, that boot the same kernel+initrd with identical command line args.

I'll spend time to do the two features:

- Have the "NOS menu" appear on the ONIE GRUB menu.
- Add two entries, with different titles, that boot the same kernel+initrd with identical command line args.

cbrune:
This looks good to me.  Pushed to master:

- fbbd6ef grub: support DIAG boot command feeding
- 64bb47c grub: support u-boot-like ONIE boot commands

One thing you need to do before this is "complete" is to update the documentation on the ONIE wiki.

You can clone the wiki source like this:
```
git clone https://github.com/opencomputeproject/onie.wiki.git
```

[side note: github does not allow pull requests for the wiki repos.  it would be great if it did.  for this I think it would be good enough to send patches to the mailing list]

The following pages/sections need some update related to these changes:

- https://github.com/opencomputeproject/onie/wiki/Design-Spec-x86-Boot-Loader
I would add a section that describes the new "onie_xxx commands" that can be used from the GRUB prompt.

- https://github.com/opencomputeproject/onie/wiki/Design-Spec-x86-NOS-Interface#x86-nos-interface
The syntax of the `onie-boot-mode` command has changed.

- https://github.com/opencomputeproject/onie/wiki/Design-Spec-x86-NOS-Interface#x86-hardware-diagnostics-interface-optional
Review this entire section.  Definitely the `GRUB Considerations` sections needs to be updated.  Need to convey that ONIE DIAG behaviour changes from 2015.11 to 2016.02.

Need some words to indicate the behaviour changed:
```
For ONIE versions up to and including 2015.11 the diagnostic image 
installs GRUB into the MBR....
``` 

Then another paragraph or two that describe the new DIAG behaviour for 2016.02 and beyond.

Let's leave this PR open until the wiki documentation is updated.

cbrune:
Closing this now that the wiki is updated with the new documentation.

PR #309, opened 2016-01-05
Interface Masters platform update.

abelits:
This is a port of Interface Masters version of ONIE, intended for merging with current ONIE development. Some IM-specific changes are the result of IM covering a wide variety of hardware, so detection rules are more generic, some are required to support a combined /boot for ONIE and OS (no chainloading).


cbrune:
@abelits - Thanks for putting these changes out for review.  I see
quite a lot of effort put into these patches.

A number of these files appear to be copies of common files with
presumably some modifications (hard to tell without seeing diffs).
While these changes are isolated to the im_n29xx_t40n machine, this
will be difficult for someone to maintain over time.

For example what happens when the common ONIE files change?  Someone
would need to track changes to the common code, reexamine all these
im_n29xx_t40n copies and modify them.  That would be hard to do long
term.

The rootconf mechanism was never intended to be used to completely
overwrite/replace the common files.  It is intended to supply platform
specific modifications within a common infrastructure.

Of course these copies were made for a reason, which implies to me
there is something missing in the common infrastructure.  Copies were
made in order to work around some limitation in the common
infrastructure.

In my experience (for any software project, not just ONIE) copying a
common file and making small local changes is usually a red flag.
This usually indicates something is either lacking in the
infrastructure, or a misunderstanding about how to use the
infrastructure exists.

I would like to first understand what the limitations are and then
update the common infrastructure to better accommodate what these
patches are trying to achieve.

To get started on that I would like know for each of the copied files:

- why was a copy necessary?
- what changed exactly?  a diff would be good here.

With that info in hand I think we can figure what to do without
copying common files.

abelits:
Some changes are trivial, and may be removed or considered for adding as options:

machine/im_n29xx_t40n/rootconf/sysroot-bin/discover is a duplicate of rootconf/default/bin/discover , except it never calls sd_localfs because our platform uses large image files, "streamed" to the installer over the network and dynamically allocates the OS/user partition. Whatever remnants of filesystems might be present when new installation is requested, are safer to be left undisturbed until the new partition and filesystem are created.

machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/init-arch is a duplicate of an older rootconf/x86_64/sysroot-lib-onie/init-arch, except if $use_onie_mount_partitions is non-empty, onie_mount_partitions() from /lib/onie/onie-blkdev-common ( machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common ) is used instead of the hardcoded mount procedure. All other changes can be replaced by just using the current version of that file, however mount procedure is useful.

machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-arch is a duplicate of an older rootconf/x86_64/sysroot-lib-onie/uninstall-arch , except it recognizes a legacy label, and uses onie_get_boot_dev() from the same /lib/onie/onie-blkdev-common instead of the hardcoded device detection procedure. This is mostly to support a legacy label and can be removed if hardcoded label in the current version of the file will be replaced with a variable. All other changes can be replaced by just using the current version of that file, so we probably can just omit it.


Most of the significant changes that produce duplicate files are to add support for the /boot partition shared between ONIE and OS:

im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common is a replacement for rootconf/x86_64/sysroot-lib-onie/onie-blkdev-common that fills the same variables using different procedures, and includes functions such as onie_get_boot_label() and onie_mount_partitions() used by other scripts. It supports identification of the shared partition.

machine/im_n29xx_t40n/installer/install-platform is a platform-specific installer (it redefines install_onie() from installer/x86_64/install-arch ) that always re-installs ONIE GRUB. Post-install procedure, if OS is installed, calls GRUB installer and update procedure, so GRUB is left as either current ONIE GRUB or OS GRUB with its menu. The preserved files manipulation procedure is changes to include boot files, and target directories are removed to prevent creating multiple levels of "preserve-update" directories. I believe, preserve-update manipulation should be ported into the mainline ONIE, however I did not want to do that without testing. Boot files update is specific to the shared /boot partition feature.

machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-platform restores GRUB menu and removes OS files from the shared /boot .

install-platform and uninstall-platform are what actually defines procedures for handling a shared directory. Some of that behavior may be generalized in the future, as 

machine/im_n29xx_t40n/arch/x86_64.make (mostly a copy of build-config/arch/x86_64.make ) and machine/im_n29xx_t40n/machine.make is basically a kludgy mechanism that allows the above file to be installed.  machine/im_n29xx_t40n/arch/x86_64.make probably can be reduced to the relevant $(UPDATER_IMAGE_PARTS) update instead of full duplication of the file, or it may be that platform-specific onie-blkdev-common, or parts of it, can be added as a build option.

cbrune:
> Some changes are trivial, and may be removed or considered for adding
> as options:
>  
> machine/im_n29xx_t40n/rootconf/sysroot-bin/discover is a duplicate of
> rootconf/default/bin/discover , except it never calls sd_localfs
> because our platform uses large image files, "streamed" to the
> installer over the network and dynamically allocates the OS/user
> partition. Whatever remnants of filesystems might be present when new
> installation is requested, are safer to be left undisturbed until the
> new partition and filesystem are created.

By "platform" do you mean a "particular OS installer" uses large image
files?  It goes against the design to remove a feature in order to
support a specific OS installer.  Then this machine is different from
other x86 ONIE machines.  That in turn makes it difficult for an OS
vendor to support a wide number of machines from different hardware
vendors.

If some other OS installer targets this machine, it would not be able
to use the local method.  For example this is how to install an OS
using USB.  So USB installs would not be possible on this machine.

> machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/init-arch is a
> duplicate of an older rootconf/x86_64/sysroot-lib-onie/init-arch,
> except if $use_onie_mount_partitions is non-empty,
> onie_mount_partitions() from /lib/onie/onie-blkdev-common (
> machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common )
> is used instead of the hardcoded mount procedure. All other changes
> can be replaced by just using the current version of that file,
> however mount procedure is useful.

Need to understand the new mount procedure.... more reading.

> machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-arch is a
> duplicate of an older
> rootconf/x86_64/sysroot-lib-onie/uninstall-arch, except it
> recognizes a legacy label, and uses onie_get_boot_dev() from the
> same /lib/onie/onie-blkdev-common instead of the hardcoded device
> detection procedure. This is mostly to support a legacy label and
> can be removed if hardcoded label in the current version of the file
> will be replaced with a variable. All other changes can be replaced
> by just using the current version of that file, so we probably can
> just omit it.

A platform can provide a custom `uninstall-platform` file that is
sourced *after* uninstall-arch.  For example you could create
`machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-platform` and
define a new version of `should_delete_msdos_partition()` that covers
the legacy label.  That would be the way to go instead of the
duplication of `uninstall-arch`.

> Most of the significant changes that produce duplicate files are to
> add support for the /boot partition shared between ONIE and OS:

Having ONIE and an OS share a /boot partition is unfortunately a
non-starter.  The system is just not designed to work that way for
legacy BIOS systems (UEFI is another story).  It is expected that the
OS will install it's own GRUB and have it's own /boot partition.

When ONIE goes full UEFI this problem goes away as there is no longer
a GRUB/MBR for OS to wrestle over.  ONIE on legacy BIOS with GRUB in
the MBR is admittedly itself a bit kludgy and a compromise.  However,
it does allow a traditional Linux OS to install its own GRUB.

See the work to enable a native Debian installer on an ONIE enabled
system:

  https://github.com/opencomputeproject/onie/tree/master/contrib/debian-iso

> im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common is a
> replacement for rootconf/x86_64/sysroot-lib-onie/onie-blkdev-common
> that fills the same variables using different procedures, and
> includes functions such as onie_get_boot_label() and
> onie_mount_partitions() used by other scripts. It supports
> identification of the shared partition.
>
> machine/im_n29xx_t40n/installer/install-platform is a
> platform-specific installer (it redefines install_onie() from
> installer/x86_64/install-arch ) that always re-installs ONIE
> GRUB. Post-install procedure, if OS is installed, calls GRUB
> installer and update procedure, so GRUB is left as either current
> ONIE GRUB or OS GRUB with its menu. The preserved files manipulation
> procedure is changes to include boot files, and target directories
> are removed to prevent creating multiple levels of "preserve-update"
> directories. I believe, preserve-update manipulation should be
> ported into the mainline ONIE, however I did not want to do that
> without testing. Boot files update is specific to the shared /boot
> partition feature.
>
> machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-platform
> restores GRUB menu and removes OS files from the shared /boot .
> install-platform and uninstall-platform are what actually defines
> procedures for handling a shared directory. Some of that behavior
> may be generalized in the future as options.

I see the above changes stem from the shared /boot directory concept.

> machine/im_n29xx_t40n/arch/x86_64.make (mostly a copy of
> build-config/arch/x86_64.make ) and
> machine/im_n29xx_t40n/machine.make is basically a kludgy mechanism
> that allows platform-specific onie-blkdev-common file to be
> installed. machine/im_n29xx_t40n/arch/x86_64.make probably can be
> reduced to the relevant $(UPDATER_IMAGE_PARTS) update instead of
> full duplication of the file, or it may be that platform-specific
> onie-blkdev-common, or parts of it, can be added as a build option.

I'm not advocating doing this, but I believe the platform specific
`machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common`
would naturally overwrite the core ONIE version of the same file.
Looking at these lines from images.make:

https://github.com/opencomputeproject/onie/blob/master/build-config/make/images.make#L205-L218

The core `x86_64/sysroot-lib-onie/*` files go in first followed by the
`$MACHINEDIR/rootconf/sysroot-lib-onie/*` files.  It think that would
just work.

abelits:
> By "platform" do you mean a "particular OS installer" uses large image
files? It goes against the design to remove a feature in order to
support a specific OS installer. Then this machine is different from
other x86 ONIE machines. That in turn makes it difficult for an OS
vendor to support a wide number of machines from different hardware
vendors.

> If some other OS installer targets this machine, it would not be able
to use the local method. For example this is how to install an OS
using USB. So USB installs would not be possible on this machine.

Our hardware does not have any storage that is accessible to the user -- there are no USB ports, SD cards or any fixed storage devices other than the bootable SSD used by ONIE and OS. The only filesystem that survives installation is one that contains ONIE itself, so that would be the only usable location for the installer. We can revert the change and allow installation from a local filesystem, I just don't think, it will produce any result other than longer discovery procedure.

> A platform can provide a custom uninstall-platform file that is
sourced after uninstall-arch. For example you could create
machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-platform and
define a new version of should_delete_msdos_partition() that covers
the legacy label. That would be the way to go instead of the
duplication of uninstall-arch.

I think, we can do that.

> Having ONIE and an OS share a /boot partition is unfortunately a
non-starter. The system is just not designed to work that way for
legacy BIOS systems (UEFI is another story). It is expected that the
OS will install it's own GRUB and have it's own /boot partition.

I have already succeeded in making this /boot partition sharing work, and allow easy switch between ONIE and OS-provided GRUB, so I believe, "non-starter" would be an exaggeration as far as the possibility of development is concerned. It may be that my choice of directories (starting from /boot , without any specific subdirectory) is bad, or it is a matter of policy to only have ONIE-originated files on ONIE partition, but I think, it may be a good thing.

I should remove any assumptions that ONIE and OS use shared /boot (I have one such place that after ONIE update scans OS /etc/fstab , and replaced UUIDs, and reinstalls GRUB -- it should only do so if /boot was originally ONIE-BOOT /boot and GRUB was installed that way), to make sure that installer that uses separate /boot will always work, however I think, it would be useful to allow shared /boot .

> When ONIE goes full UEFI this problem goes away as there is no longer
a GRUB/MBR for OS to wrestle over. ONIE on legacy BIOS with GRUB in
the MBR is admittedly itself a bit kludgy and a compromise. However,
it does allow a traditional Linux OS to install its own GRUB.

I do not believe, "ONIE goes full UEFI" is a worthwhile goal, and I am certainly not looking forward to that. While MBR BIOS boot procedure is not a great design, it is a consistently implemented design across all hardware, and GRUB support of it provides us with an easy and reliable way of utilizing it. In my opinion, GRUB is what UEFI should have been but is not.

> See the work to enable a native Debian installer on an ONIE enabled
system:

> https://github.com/opencomputeproject/onie/tree/master/contrib/debian-iso

This is probably very useful for some systems, however we wanted to optimize the installation time, reduce the number of possible error conditions, and prevent any possibility that failed unattended installation will result in a seemingly usable system, so we made an image-based installation procedure that installs a filesystem as an image, then mounts it and performs post-installation steps that make it bootable and mountable (updating GRUB and /etc/fstab). This is the installer: https://github.com/InterfaceMasters/onie-x86-installer , and this is an image generator that runs Debian package manager, debotstrap, etc: https://github.com/InterfaceMasters/radish . This is now our standard installation procedure for installation servers (so some ONIE installation servers are installed with ONIE), and it works very well. I don't remember when I have last seen Debian ISO.

> I see the above changes stem from the shared /boot directory concept.

Except for the preserved files directory, though I am not sure if it is a problem without shared /boot

> I'm not advocating doing this, but I believe the platform specific
machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common
would naturally overwrite the core ONIE version of the same file.
Looking at these lines from images.make:

> https://github.com/opencomputeproject/onie/blob/master/build-config/make/images.make#L205-L218

> The core x86_64/sysroot-lib-onie/* files go in first followed by the
$MACHINEDIR/rootconf/sysroot-lib-onie/* files. It think that would
just work.

I remember, it did not work that way in some earlier version, I will check if it works now.

cbrune:
> > If some other OS installer targets this machine, it would not be able
> > to use the local method. For example this is how to install an OS
> > using USB. So USB installs would not be possible on this machine.
>  
> Our hardware does not have any storage that is accessible to the user
> -- there are no USB ports, SD cards or any fixed storage devices other
> than the bootable SSD used by ONIE and OS. The only filesystem that
> survives installation is one that contains ONIE itself, so that would
> be the only usable location for the installer. We can revert the
> change and allow installation from a local filesystem, I just don't
> think, it will produce any result other than longer discovery
> procedure.

Interesting.  I see the point.  I think we can accommodate that choice
on a per machine basis with a small modification to the
`rootconf/default/bin/discover` file.

Today that file sources `/lib/onie/discover-arch` for architecture
specific details.  We could also have the script source a new platform
specific file `/lib/onie/discover-platform`.

We could make each discovery method (the `sd_XXX` functions)
conditional on a configuration option.  The default would be enable,
but a platform could set it to disabled.

I can take a crack at that.  I created #312 to track.

> > A platform can provide a custom uninstall-platform file that is
> > sourced after uninstall-arch. For example you could create
> > machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/uninstall-platform and
> > define a new version of should_delete_msdos_partition() that covers
> > the legacy label. That would be the way to go instead of the
> > duplication of uninstall-arch.
>  
> I think, we can do that.

Great, one more item off the todo list.

> > I'm not advocating doing this, but I believe the platform specific
> > machine/im_n29xx_t40n/rootconf/sysroot-lib-onie/onie-blkdev-common
> > would naturally overwrite the core ONIE version of the same file.
> > Looking at these lines from images.make:
> >  
> > https://github.com/opencomputeproject/onie/blob/master/build-config/make/images.make#L205-L218
> >  
> > The core x86_64/sysroot-lib-onie/* files go in first followed by the
> > $MACHINEDIR/rootconf/sysroot-lib-onie/* files. It think that would
> > just work.
> I remember, it did not work that way in some earlier version, I will check if it works now.

Great, let me know if that works.  One less item.

> > Having ONIE and an OS share a /boot partition is unfortunately a
> > non-starter. The system is just not designed to work that way for
> > legacy BIOS systems (UEFI is another story). It is expected that the
> > OS will install it's own GRUB and have it's own /boot partition.
>  
> I have already succeeded in making this /boot partition sharing work,
> and allow easy switch between ONIE and OS-provided GRUB, so I believe,
> "non-starter" would be an exaggeration as far as the possibility of
> development is concerned. It may be that my choice of directories
> (starting from /boot , without any specific subdirectory) is bad, or
> it is a matter of policy to only have ONIE-originated files on ONIE
> partition, but I think, it may be a good thing.
>  
> I should remove any assumptions that ONIE and OS use shared /boot (I
> have one such place that after ONIE update scans OS /etc/fstab , and
> replaced UUIDs, and reinstalls GRUB -- it should only do so if /boot
> was originally ONIE-BOOT /boot and GRUB was installed that way), to
> make sure that installer that uses separate /boot will always work,
> however I think, it would be useful to allow shared /boot .

On the shared boot partition I think we have to agree to disagree.  As
the custodian of the ONIE project I am considering the larger
community and ecosystem when I express my opinions here.

On one hand we have over 27 x86 machines from 8 hardware vendors that
are following the ONIE spec as it is written today.  Adding one
machine that does things differently adds a number of headaches,
including:

- who will support end users using this ONIE version?  Who on the ONIE
  mailing list will help them?

- who will support OS vendors targeting this platform?  Who on the
  ONIE mailing list will help them?

- how will OCP ONIE certification testing work for this platform?  Are
  changes required to the OCE testing infrastructure to support this
  platform?  Who is going to do that?

On the other hand, however, I do not want to stifle innovation.  Maybe
you are on to something and as long as the changes are localized to
this platform it seems manageable.  We should allow it and see what
happens.

So I have my expressed my concerns, but let's get this platform into
the repo.

> > When ONIE goes full UEFI this problem goes away as there is no longer
> > a GRUB/MBR for OS to wrestle over. ONIE on legacy BIOS with GRUB in
> > the MBR is admittedly itself a bit kludgy and a compromise. However,
> > it does allow a traditional Linux OS to install its own GRUB.
>  
> I do not believe, "ONIE goes full UEFI" is a worthwhile goal, and I am
> certainly not looking forward to that. While MBR BIOS boot procedure
> is not a great design, it is a consistently implemented design across
> all hardware, and GRUB support of it provides us with an easy and
> reliable way of utilizing it. In my opinion, GRUB is what UEFI should
> have been but is not.

We can agree to disagree about UEFI also.  UEFI is coming, especially
for folks who need/want to see Secure Boot implemented.

I do value your thoughtful concerns.  Perhaps you can share some of
these on the ONIE mailing list.

cbrune:
Make that 28 x86 machines -- DELL just upstreamed another machine today:
- 5740ba2f57b1 Dell S6100 Data center Switch.

abelits:
I have committed more changes to bring platform support closer to the  master branch. makefile changes happened to be still necessary -- onie-blkdev-common file still has to be separately copied into the updater at https://github.com/opencomputeproject/onie/blob/master/build-config/arch/x86_64.make#L88-L89 , so I have made an alternative modification, added a variable for all files that should be copied along with or over the original ones.

Shared /boot support is now safe -- fstab and GRUB updates are only performed if OS has /boot in its /etc/fstab, and its UUID is no longer present after ONIE installation.

cbrune:
Looks OK.

I pushed the changes to master with a few minor changes to commit messages (no code changes):

- squashed several of the commits (9e247ce, 288ec73, 4d1696d, b541e79) into the original commit 5426ae9.  These were just undoing things done in 5426ae9.
- moved 1146ac0 ahead of 5426ae9 (logically we add common functionality first and then make use of the new functionality in a machine, not the other way round)
- minor rewording of commit message:
  - 1146ac0 - added a little bit more description
  - 8da00da, 9ecec60 -- added "im_n29xx_t40n:" prefix to title (practice is to prefix machine specific commit message with machine name"

The end results are here:
- 588f52a UPDATER_IMAGE_PARTS_PLATFORM support
- ea93a19 Interface Masters platform update.
- bb00c7e im_n29xx_t40n: Update OS /etc/fstab on ONIE update only if /boot is shared.
- 51e1105 im_n29xx_t40n: OS GRUB re-installation is also conditional.

abelits:
Thanks!

PR #313, opened 2016-01-19
Alphanetworks Platforms: Add SNX60A0-486T to support ONIE.

willkuo103:
The machine has 48 ports of 10 Gbps RJ-45, 2 ports of 40 Gbps QSFP+ and 4 ports of 100Gbps QSFP28 ports on Broadcom Trident2+ switch ASIC.  CPU module is Intel Rangely
C2558.


machine/alphanetworks/alphanetworks_snx60a0_486t/kernel/drivers-base-dd-alpha-snx60a0-486t.patch
```diff
@@ -0,0 +1,24 @@
...
+ 			driver_deferred_probe_trigger();
+ 	} else if (ret != -ENODEV && ret != -ENXIO) {
+-		/* driver matched but the probe failed */
+-		printk(KERN_WARNING
+-		       "%s: probe of %s failed with error %d\n",
+-		       drv->name, dev_name(dev), ret);
++		/* Skip the probe error message for these three devices on Alphanetworks platforms. */
++		if (strncmp(dev_name(dev), "0000:00:14.0", 12) &&
++			strncmp(dev_name(dev), "0000:00:14.2", 12) &&
++			strncmp(dev_name(dev), "0000:00:14.3", 12)) {
```
cbrune: What are these devices?
Why are they failing?

machine/alphanetworks/alphanetworks_snx60a0_486t/i2ctools/001-i2ctools-sys-eeprom-alpha-snx60a0-486t.patch
```diff
@@ -0,0 +1,17 @@
+Patch to add sys_eeprom information
```
cbrune: Using i2ctools for the onie-syseeprom command is no longer supported since March 16th 2015.  See this message to the mailing list:

We only use the busybox implementation of onie-syseeprom going forward.  The message to the group describes how to use the busybox method and a pointer to an example.

Let me know if you have any problems with that.

machine/alphanetworks/alphanetworks_snx60a0_486t/installer.conf
```diff
@@ -0,0 +1,14 @@
+# Alpha Networks SNX60A0-486T ONIE installer configuration
+
+description="AlphaNetworks, SNX60A0-486T"
+
+# Default ONIE block device
+install_device_platform()
+{
+    echo /dev/sda
```
cbrune: Is this always true?  Even when a USB disk is plugged in?

willkuo103: Yes, It is always true. The device node, "/dev/sda", is always represented for the SSD in our device even the USB disk is plugged in.

machine/alphanetworks/alphanetworks_snx60a0_486t/kernel/drivers-base-dd-alpha-snx60a0-486t.patch
willkuo103: These devices are Intel Rangeley GbE controllers. Only the 2nd GbE controller is used in our device and the other GbE controllers are not be used. Therefore, the error message will be displayed on console when the driver is trying to configure these unused GbE controllers. This piece of the code just mask the error message to prevent the users from confusing with these error messages.

willkuo103:
Hi Curt,

I have changed to use the busybox method for accessing onie-syseeprom.
If you have any question, please let me know.

Thanks,
Will

cbrune:
This looks good.  Thanks for moving to the busybox method.

Before putting this in the repo, however, please collapse these two patches together.  As it is now the old i2ctools code would go into the repo with the first patch and then be removed by the second patch -- that old i2ctool code would be in the git log forever, which is gross and unnecessary.

I suggest closing this pull request and open a new one with the updated busybox eeprom method.

willkuo103:
No problem. I closed this request and will open a new one. Thanks a lot.

PR #314, opened 2016-01-22
Alphanetworks Platforms: Add SNX60A0-486T to support ONIE.

willkuo103:
The machine has 48 ports of 10 Gbps RJ-45, 2 ports of 40 Gbps QSFP+ and 4 ports of 100Gbps QSFP28 ports on Broadcom Trident2+ switch ASIC. CPU module is Intel Rangely
C2558.


cbrune:
Looks good, pushed to master:

- 4491c58 Alphanetworks Platforms: Add SNX60A0-486T to support ONIE.

PR #323, opened 2016-01-27
demo NOS: demo two menu entries and add menu entry chainload NOS

david56:
For the patchset, I implement two features discussed in PR #308.
- Have the "NOS menu" appear on the ONIE GRUB menu.
- Add two entries, with different titles, that boot the same kernel+initrd with identical command line args.

### DIFF
--- a/demo/installer/x86_64/install.sh
+++ b/demo/installer/x86_64/install.sh
@@ -170,44 +170,7 @@ demo_install_grub()
     # keep grub loading ONIE page after installing diag.
     # So that it is not necessary to set "ONIE" as default boot
     # mode in diag's grub.cfg.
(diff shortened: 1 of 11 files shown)


cbrune:
Hello @david56 -

These both look reasonable.  a7be9b8 clearly looks to be OK.

865fdbc also looks OK, but I need to load it up and test it out to be sure.

As the merge window for 2016.02 is currently closed these changes will not go into that release.

I will get back to testing these changes after the 2016.02.  Too much to do until that is out.

ehdoyle:
@david56  - Is this still relevant? It may have to be reworked to apply to the master branch as component upgrades have broken the builds of some existing platforms.

PR #334, opened 2016-02-23
Fixes #333 - remove fall back to static IP

### LINKED ISSUE #333: Static IPv4 address allocation fallback scheme
The fallback static IPv4 address scheme is outdated and causing problems.

This thoroughly described here:

http://lists.opencompute.org/pipermail/opencompute-onie/2016-February/001026.html

Close this bug after the patch, documentation and OCE tests are updated.

capveg:
networking would fall back to 192.168.3.10 by default if there was
no DHCP in the network.  This eclipsed a customer's IP address :-(

I've only done limited testing here - feedback welcome.

Please review: @cobracmder @cbrune

### DIFF
--- a/rootconf/default/etc/init.d/networking.sh
+++ b/rootconf/default/etc/init.d/networking.sh
@@ -71,69 +71,6 @@ config_ethmgmt_dhcp4()
 
 }
 
-# Fall back ethernet management configuration
-config_ethmgmt_fallback()
-{
-
-    local base_ip=10
-    local prefix=24
-    local default_hn="onie-host"
-    local intf_counter=$1
-    shift
-    local intf=$1
-    shift
-
-    # Remove any previously configured, IPv4 addresses
-    ip -f inet addr flush dev $intf
-
-    # Assign sequential static IP to each detected interface
-    local interface_base_ip=$(( $base_ip + $intf_counter ))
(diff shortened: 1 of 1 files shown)


cobracmder:
Thanks @capveg .  I'm currently swamped with summit duties so I might have to punt running OCE on a few platforms until after the summit.  Would that be an issue?

Also, I know github makes this a PITA, but do you have a patch/update for the documentation as well?

Thanks.

capveg:
No worries about looking at the patch until after OCP Summit.  This just seems to be bad luck that it affected this one customer and it will take time to propagate through from HEAD to the vendors to the customers anyway.  I'll try to take a swing at the documentation as well.

capveg:
FYI - apparently I have super user powers and was able to edit the wiki inline.  Since it was just to remove the one line and it had already been approved by you and Curt, I just went ahead and did it.  So documentation should be fixed now too.

capveg:
Was just pointed out to me by a coworker that my original patch deleted the RFC-3927 algorithm as well.  Added that back.

cbrune:
Thanks @capveg for the patch and updating the documentation!

I cleaned up the PR slightly (squashed to a single commit) and applied:

- 23fdae8 networking: don't fall back to 192.168.3.10

Leaving #333 open to track OCE updates.

PR #337, opened 2016-03-03
Update onie-sys-eeprom-platform.patch

dantejohn:
Fixed fail when building for quanta common rangeley.


cbrune:
Thanks -- pushed to master:

- 1215d22 quanta_common_rangeley: Update onie-sys-eeprom-platform.patch

PR #340, opened 2016-03-07
Alphanetworks Platforms: Add SNH60A0-320Fv2 to support ONIE.

willkuo103:
The machine has 32 x 100GbE QSFP28 ports on Broadcom Tomahawk switch ASIC. CPU module is Intel Rangely C2558.


cbrune:
Looks good, pushed to master:

- 31b2c91 Alphanetworks Platforms: Add SNH60A0-320Fv2 to support ONIE.

PR #345, opened 2016-03-15
Machine specific, add usb to i2c chip driver. 

vic1231:
Also fix some description and blank space with onie-syseeprom-rw-wnc.patch to avoid busybox patch fail.
The driver of mcp2221 is MODULE_LICENSE("GPL v2");
The path with this eeprom is: CPU---USB hub---mcp2221---i2c---eeprom


cbrune:

- 3ca3dde wnc_sst1_n1: Add MCP2221 usb to i2c chip driver

PR #346, opened 2016-03-15
Update spelling and title case

lathama:
Correct bound to baud then notice and correct title case.


cbrune:
Thanks!  Pushed to master:

- 1e43382 Update spelling and title case

PR #348, opened 2016-03-24
Add switch-silicon-vendor to onie-sysinfo

cobracmder:
Add switch silicon vendor string to be part of the onie-sysinfo
output when using '-a' or when using '-S'.

Signed-off-by: Carlos Cardenas <carlos@cumulusnetworks.com>


cbrune:

- 6015dc9 Add switch-silicon-vendor to onie-sysinfo

PR #353, opened 2016-04-19
Allow Ethernet management MAC addresses to be pre-programmed

cbrune:
Historically ONIE programs the Ethernet management MAC addresses from
those defined in the the TLV EEPROM (type codes 0x24 and 0x2A).  This
worked well on the initial embedded systems ONIE supported, where the
NIC did not have its own internal EEPROM.

Today, most x86 machines have Ethernet management NICs that already
have a MAC address assigned by the vendor.  Overriding that MAC
address with one from the ONIE TLV EEPROM is redundant and wasteful.

This patch provides a configuration mechanism that allows machines to
disable the historic ONIE behaviour.  When ONIE does not assign the
MAC address, the MAC address will be assigned by the NIC driver,
coming from the internal EEPROM of the NIC.

The historic behaviour is enabled by default and no existing machines
need to change their configurations.

To enable the new behaviour set the SKIP_ETHMGMT_MACS variable to
"yes" in machine.make:

  # Use MAC address programmed in Ethernet management NIC
  SKIP_ETHMGMT_MACS = yes

### DIFF
--- a/build-config/Makefile
+++ b/build-config/Makefile
@@ -1,7 +1,7 @@
 #-------------------------------------------------------------------------------
 #>
 #
-#  Copyright (C) 2013,2014,2015 Curt Brune <curt@cumulusnetworks.com>
+#  Copyright (C) 2013,2014,2015,2016 Curt Brune <curt@cumulusnetworks.com>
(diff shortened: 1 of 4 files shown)


cbrune:
pushed to master:

- 4e38978 Allow Ethernet management MAC addresses to be pre-programmed

PR #332, opened 2016-02-19
Answer to simple yes/no questions is 'yes' when install packages

pichuang:
It need be more automatically

Signed-off-by: Phil Huang <pichuang@cs.nctu.edu.tw>

PR #358, opened 2016-05-08
Pass -y to apt-get to avoid prompts

lool:
Hi,

This avoids the build getting stuck on APT prompting to confirm the list of packages to install. This happens when some dependencies of the requested packages are needed to fulfill the installation.

Cheers,
- Loïc Minier


lool:
It seems a similar change was submitted as https://github.com/opencomputeproject/onie/pull/332

cbrune:
Pushed PR #332

PR #359, opened 2016-05-08
Sort files lists before comparing

lool:
Hi,

Directory listing is not guaranteed to be ordered; to compare directory listings these need to be sorted first. 

This also throws in a LC_ALL=C before the sort to make sure behavior is consistent across locales as sort is locale specific. The overall logic would work whatever the locale since both sorts are run with the same locale, but this guarantees identical on disk intermediate files and consistent failure messages whatever the locale.

Cheers,
- Loïc Minier


cbrune:
Thanks - pushed to master:

- 9e2b96c Sort files lists before comparing

PR #362, opened 2016-05-16
Facebook Wedge100: add to support ONIE

david56:
This machine has 32x100GbE QSFP ports on Broadcom Tomahawk switch ASIC.  The CPU is Intel Bay Trail.


cbrune:
Thanks, looks good.  Pushed to master:

- 3a93b3c Facebook Wedge100: add to support ONIE

PR #363, opened 2016-05-16
Px switch2

matt-morgan-plexxi:
Plexxi Switch 2 ONIE support.  Unlike our other ONIE work for our older switches this contains no Plexxi specific changes so we would like to try to get this in to the mainline of ONIE code.


cbrune:
Hi @matt-morgan-plexxi -

Thanks for sending in the patches.  I have a few general comments about the patchset -- 

Firstly, when making a pull request please do not include commits that add and revert a commit from the set.  That does nothing but pollute the git history.  That is like adding 1 and then subtracting 1.  Just don't submit the patch to be reverted in the first place.  For the series in question these comments apply to 253fe7a and 77d8580.

Secondly, a pull request should not contain a "Merge remote-tracking branch" -- all that will be taken care of when the pull request is pulled in.  These types of commits just pollute the git history.  This comment applies to 791d58c.

Please rework your pull request to not have these commits.

The commits that add real changes (58e40e0 and f1c89a3) are mostly OK.  A few comments in line.

machine/plexxi/px_switch2/kernel/config
```diff
@@ -0,0 +1,37 @@
...
+CONFIG_VIRTIO_RING=y
+CONFIG_VIRTIO_PCI=y
+CONFIG_VIRTIO_BALLOON=y
+CONFIG_PARAVIRT_GUEST=y
+CONFIG_PARAVIRT=y
+CONFIG_PARAVIRT_SPINLOCKS=y
+CONFIG_PARAVIRT_CLOCK=y
+CONFIG_KVM_CLOCK=y
+CONFIG_KVM_GUEST=y
+CONFIG_HAVE_KVM=y
```
cbrune: Looks like you picked up all the kernel config from the kvm_x86_64 virtual machine.  Do you really want that in there?

machine/plexxi/px_switch2/machine.make
```diff
@@ -0,0 +1,33 @@
...
+ifeq ($(VENDOR_REV),0)
+  MACHINE_REV = 0
+else
+  $(warning Unknown VENDOR_REV '$(VENDOR_REV)' for MACHINE '$(MACHINE)')
+  $(error Unknown VENDOR_REV)
+endif
+
+# Vendor ID -- IANA Private Enterprise Number:
+# http://www.iana.org/assignments/enterprise-numbers
+# Open Compute Project IANA number
```
cbrune: This comment is not right.

machine/plexxi/px_switch2/kernel/config
matt-morgan-plexxi: Good point.  I guess it worked so I just left it in there.  Do you want it removed?

cbrune: If it were my code I would remove it.  It is not a deal breaker for me, just unnecessary.

cbrune: As for the old code submitted previously -- it was never pulled into the ONIE repo, so it does not exist in the project.  Please base your changes off the ONIE master branch when making a pull request.

See item 2 here:

PR #364, opened 2016-05-17
Accton AS7716_32X: revise order of block devices for ONIE installation

david56:


cbrune:
Looks, good.  Pushed to master.

PR #367, opened 2016-05-17
Accton 833435_B21: add to support ONIE

david56:
The machine has 8x1/10G SFP+ and 4x1/10GBASE-T ports on Broadcom
Maverick switch ASIC.  CPU module is Intel Rangeley C2558.


cbrune:
Looks, good.  Pushed to master.

PR #368, opened 2016-05-17
Accton AS7312_54X: add to support ONIE

david56:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on Broadcom Tomahawk switch ASIC.  CPU module is Intel Rangeley C2538.


cbrune:
Looks good, pushed to master.

PR #370, opened 2016-05-17
Plexxi Switch 2 ONIE support.

matt-morgan-plexxi:
Updated after comments from Curt.


machine/plexxi/px_switch2/i2ctools/i2ctools-plexxi.patch
```diff
@@ -0,0 +1,114 @@
+
```
cbrune: I thought I made this comment on the first PR, but looking back I do not see it.  Perhaps I forgot to click the "Comment" button....

Anyway -- Since March 16, 2015 (over a year ago) the ONIE project no longer accepts i2ctools patches for EEPROM access.  That announcement and reasons are described here:

We will only put our limited resources into updating and maintaining the busybox version of the onie-syseeprom command going forward.

That email describes how to use busybox for EEPROM access instead of i2ctools.

cbrune:
This pull request looks a lot better.  Thanks for cleaning it up.  The only problem now is the i2ctools patches for EEPROM access.

matt-morgan-plexxi:
Removed i2ctools and now use busybox EEPROM.

cobracmder:
@matt-morgan-plexxi , can you squash those two commits into one (so the git history is not polluted with adding a file and immediately removing it in the next commit)?

cobracmder:
Looks all shiny and chrome.
Shiny

matt-morgan-plexxi:
We bought a case of this at Plexxi.

http://www.amazon.com/Wilton-710-5521-Silver-Color-Mist/dp/B005KTVG86/ref=sr_1_sc_2?ie=UTF8&qid=1463684561&sr=8-2-spell&keywords=shiny+and+chrom+spray

Witness!

(BTW, read the reviews)

cbrune:
@matt-morgan-plexxi -- thanks for the work.  Pushed to master:

- ae3a6d8 Plexxi Switch 2 ONIE support.

PR #372, opened 2016-05-18
Add support for arm64 toolchain and Freescale LS1043ARDB and LS2080ARDB

pankajbansal3073:
Add support for arm64 toolchain and Freescale LS1043ARDB and LS2080ARDB


build-config/scripts/fetch-package
```diff
@@ -58,7 +58,7 @@ else
     download=no
     while [ $# -ge 1 ] ; do
         URL="$1/$FILE"
-        wget --no-verbose -O "${OUTDIR}/${FILE}" $URL
+        wget --no-verbose --no-check-certificate -O "${OUTDIR}/${FILE}" $URL
```
cobracmder: Why is ```--no-check-certificate``` necessary ?

build-config/make/crosstool-ng.make
```diff
@@ -43,24 +43,34 @@ crosstool-ng-download: $(CROSSTOOL_NG_DOWNLOAD_STAMP)
...
+		 fi
 	$(Q) touch $@
 
 SOURCE += $(CROSSTOOL_NG_SOURCE_STAMP)
 crosstool-ng-source: $(CROSSTOOL_NG_SOURCE_STAMP)
 $(CROSSTOOL_NG_SOURCE_STAMP): $(CROSSTOOL_NG_DOWNLOAD_STAMP)
 	$(Q) rm -f $@ && eval $(PROFILE_STAMP)
 	$(Q) echo "==== Extracting upstream $(CROSSTOOL_NG_DESC) ===="
-	$(Q) $(SCRIPTDIR)/extract-package $(CROSSTOOL_NG_BUILD_DIR) $(DOWNLOADDIR)/$(CROSSTOOL_NG_TARBALL)
+	$(Q) if [ $(awk 'BEGIN{ print "'$(CROSSTOOL_NG_VERSION)'"<="'1.22.0'" }') -eq 1 ]; then \
```
cobracmder: This is kinda dangerous as specifying an arbitrary number above what's define retrieves the master branch.  Might consider something like ```GIT-REF``` were ```REF``` can be a branch or a particular commit to checkout.  This way builds can be reproducible.

cobracmder:
Hi @pankajbansal3073 ,

Thanks for taking a stab at this.  Following the contribution guidelines (https://github.com/opencomputeproject/onie/wiki/Contributing), in particular this section (https://github.com/opencomputeproject/onie/wiki/Contributing#general-patching-philosophy), we need to break up the commits into logical portions.

For example, (98ae4a2), needs to be broking up into various commits.  77a9885 needs to be squashed in another commit (you're already changing that file, put the fix in that commit). Same with c336ebb.

My recommendation would be the following:
* have a commit that adds the kernel y'all want to use (4.4.3 from the looks of it) and make sure to have conf for the arm64 platform (having an x86_64 config might be useful for testing with kvm_x86_64)
* have a commit that adds your crosstool-ng / xtools changes (what is missing from your PR is downloading gcc-5.3.0 since that is what is being configured)
* have a commit that adds glibc (this is missing from your PR right now, there's no way to downloading glibc)
* have a commit for the qemu_arm64 platform (a VM that demonstrates the toolchain and components produce a working, ONIE Compliant implementation)
* perform PR on those commits to make sure the other 80+ platforms still build and work

Once that PR has been pulled in, then
* have a commit for LS1043ARDB
* create PR for platform
* have a commit for LS2080ARDB
* create PR for platform

build-config/scripts/fetch-package
pankajbansal3073: to download files from https url instead of http url

pankajbansal3073:
i will revert all the changes and make commits as you instructed

PR #373, opened 2016-05-18
Rename vendor name from fsl to nxp to match ONL install

shengzhou:
This patch resolved the issue of ONL install failure due to "onie_platform" via ONIE for NXP platforms.


cobracmder:
Hi @shengzhou ,

I see the motivation in making these changes but I thought (I could be wrong) that the platforms in fsl where reference board designs?  I.e. I buy a reference board from FreeScale and now I can put ONIE on it.  Is that correct @cbrune?

The other comment I have is this commit (24fab8a), doesn't change all the necessary files:
* machine/fsl/fsl_p2020rdbpca/onie-rom.conf
* machine/fsl/fsl_p2020rdbpca/machine.make
* etc....

My advice is rather than change all of those platforms at once, have one commit per platform (along with testing for each platform) to ensure all files are changed accordingly.

shengzhou:
Hi @cobracmder , 

Yes, for those reference boards which have been supported in ONIE, if one have the board, he can put ONIE on it, even also can install ONL on it(e.g. T2080RDB, LS2080, etc).  
When one install ONL on T2080RDB via ONIE,  ONIE executes the installer of ONL, installer will check the environment "onie_platform"  which generated by ONIE during build as following:
  onie_platform=$(PLATFORM)
  PLATFORM  = $(ARCH)-$(MACHINE_PREFIX)
  MACHINE_PREFIX = $(MACHINE)-r$(MACHINE_REV)
  make -j4 MACHINEROOT=../machine/nxp MACHINE=nxp_t2080rdb all
So we have to change  the directory machine/fsl to machine/nxp, as all freescale/nxp platforms are in machine/fsl folder, they have to be updated in once rather than separate platform.  I have tested the patch on T2080RDB. this patch change doesn't imapct ONIE on the other serveral platforms.

cbrune:
@shengzhou -- do you have any concerns about legacy customers who have the previous version of ONIE installed?

I'm pretty sure trying to update the old ONIE to the new ONIE using "onie-self-update" will fail as the machine names are different.  That is a safety check during ONIE updates, making sure the correct image is installed on the correct machine type.  See this default installer code:

  https://github.com/opencomputeproject/onie/blob/master/installer/install.sh#L115-L126

Other than that I have no concerns about this change.

A question of historical interest -- why were the ONL names updated before the ONIE names?  Usually ONIE is created first.

shengzhou:
@cbrune   As fsl had been merged as new nxp, for long-term consideration, the vendor name should be updated to nxp to match other open source projects in which nxp is used as vendor name.  For legacy customers who have the previous version of ONIE installed, they have to manually update to new onie instead of the automatic "onie-self-update", of cource, if they don't need to update onie, just keep using legacy version.

cbrune:
I still see files named `fsl` and references to `fsl` in the code -- is that a problem for you?  For example this file is still named `fsl` and it contains references to `fsl` strings.

```
monster-15:~/onie-cbrune/machine$ grep fsl nxp/nxp_p2020rdbpca/kernel/platform-fsl-p2020rdb-pca.patch 
platform fsl p2020rdb patch
        compatible = "fsl,P2020RDB";
                        compatible = "fsl,p2020-fcm-nand",
                                     "fsl,elbc-fcm-nand";
```

shengzhou:
@cbrune  For those patches of kernel and u-boot, it still keeps fsl without change because there is no checking vendor name dependency. so don't care the files named fsl because the files are for u-boot or kernel, not part of onie base itself.

cbrune:
Pushed to master:

- e0c3a9d Rename vendor name from fsl to nxp to match ONL install

PR #375, opened 2016-05-19
Add arm64 (aarch64) architecture armv8a support in onie

pankajbansal3073:
Added support for arm64 (aarch64) architecture armv8a in onie


cobracmder:
Hey @pankajbansal3073 ,

Thanks for breaking this up; the changes look more straight forward than before.

Please allow some time for people to review these changes.

On the testing front, have y'all:
* compiled and ran ONIE on: x86_64 (i.e. kvm_x86_64), ppc (fsl p2020), and armv7a (qemu_armv7a)
* compiled aarch64 toolchain

Thanks.

build-config/Makefile
```diff
@@ -162,7 +162,7 @@ ifneq (,$(MAKECMDGOALS))
       $(error Unable to find valid machine configuration directory.)
     endif
     include $(MACHINEDIR)/machine.make
-    ARCHDIR ?= $(realpath ./arch)
+    ARCHDIR ?= $(realpath ./arch/$(ONIE_ARCH))
```
cobracmder: This looks correct; however, when attempting to build the ```kvm_x86_64``` platform (the default) I'm receiving an error.

A simple way to reproduce the error is:
```
$ make download
Makefile:166: /x86_64.make: No such file or directory
make: *** No rule to make target '/x86_64.make'.  Stop.
```

The same error appears when specifying the platform:
```
$ make MACHINE=kvm_x86_64 download
Makefile:166: /x86_64.make: No such file or directory
make: *** No rule to make target '/x86_64.make'.  Stop.
```

cobracmder: I was able to get it working with the following change:
```
ARCHDIR ?= $(shell realpath ./arch/$(ONIE_ARCH))
```

Except now, I get the following error:
```
Makefile:166: [snip]/x86_64.make: No such file or directory
make: *** No rule to make target '[snip]/x86_64.make.  Stop.
```

This is because the arch directory is named ```x86_x64``` vs ```x86_64```.

pankajbansal3073: i have corrected this mistake.

pankajbansal3073: Thanks for reviewing my changes
1. i have compiled qemu_armv7a and arm64 toolchain aarch64 successfully with these changes.
2. i have corrected the mistake with x86_64 architecture.

cobracmder:
Can you squash the fixes (ARCHDIR and x86_64 dirname) into the previous commit?

Thanks.

pankajbansal3073:
i have squashed the x86_64 dirname commit to previous commit

pankajbansal3073:
Updated patch in kexec-tools to support Aarch64 (arm64)

pankajbansal3073:
i have made some more changes in build scripts. with these changes now the single onie binary image and updater image is building successfully for arm64 targets

cobracmder:
@pankajbansal3073 , this is looking good (currently building a sampling of x86_64, ppc, and arm32 platforms).

What command(s) are you using to build the aarch64 toolchain/kernel?

pankajbansal3073:
@cobracmder , i am using "make -j4 MACHINEROOT=../machine/fsl MACHINE=fsl_ls1043ardb xtools-build" and "make -j4 MACHINEROOT=../machine/fsl MACHINE=fsl_ls1043ardb all" commands to build aarch64 toolchain and onie binary. But i have not yet created a pull request for machine changes yet. (https://github.com/pankajbansal3073/onie.git)

cobracmder:
@pankajbansal3073 , cool...would it be possible to add a `qemu_armv8a` machine platform (separate commit, of course) so the full _shebang_ can be exercised?

pankajbansal3073:
@cobracmder , I have added the qemu virtual machine and UEFI image for it too. since this virtual machine is not using any dtb/ftd image, I am getting the error in full binary creation. But the toolchain, Linux kernel, and initrd images are being generated successfully.

I have also updated the script to run these images in qemu.
when I am running the Linux kernel in qemu I am getting kernel panic. :

/init: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
[    6.283551] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00

even though libcrypt.so.1 file is present in initrd, still the kernel panic is coming. I am not able to understand this. 
if we run this kernel with qemu_armv7a initrd file, then kernel is booting successfully. Could you please help to check this problem ?

cobracmder:
@pankajbansal3073 , You bet; I'll take a look at what's going on.

cobracmder:
Hey @pankajbansal3073 , I got a few questions/observations (based on what's in the code):
* What's the firmware for the arch (```armv8a/aarch64```)?  In the arch makefile, it's UBoot.  Is this correct?
* What's the firmware for the ```qemu_armv8a``` platform?  In the ```machine.make file```, I see ```UBOOT_ENABLE = no```  but not mention of enabling UEFI support.
* In the ```launch-qemu.sh``` script, I see you are trying to use UEFI with the kernel/ramdisk but the platform wasn't compiled with ```UEFI_ENABLE = yes``` in the ```machine.make``` but in the arch makefile it's trying to build uboot images (itb and dtb).

My advice based on those observations and questions is the following: If ```armv8a``` is going to be UBoot based, then ```qemu_armv8a``` should be UBoot based as well (like ```qemu_armv7a```).  If ```armv8a``` is going to be UEFI based, then ```qemu_armv8a``` should be UEFI based as well (like ```kvm_x86_64```).

pankajbansal3073:
@cobracmder : i have added UEFI and Grub support in Qemu virtual machine, but still the compilation is failing. can you please check ? also the kernel panic problem when trying to boot onie image is still not solved. i am not able to understand why system is not able to find libcrypt.so.1, even though it is present in initrd image. Can you please help ?

build-config/scripts/onie-mk-iso.sh
```diff
@@ -154,9 +154,8 @@ if [ "$UEFI_ENABLE" = "yes" ] ; then
 
     # Generate UEFI format GRUB image
     mkdir -p $RECOVERY_EFI_BOOT_DIR
-    $GRUB_HOST_BIN_UEFI_DIR/grub-mkimage \
-        --format=x86_64-efi \
-        --directory=$GRUB_HOST_LIB_UEFI_DIR \
+    grub-mkimage \
```
pankajbansal3073: While running the grub-mkimage command in GRUB_HOST_BIN_UEFI_DIR for arm64-efi format, I was getting this error : "relocation 0x2 is not implemented yet". Which is why I used the grub-mkimage of build machine.

pankajbansal3073:
@cobracmder : I finally got the ONIE image compiled with UEFI and GRUB support for aarch64. But now when I am trying to boot this image, I am entering in grub rescue mode:

EFI Boot menu -> Boot Maintenance Manager -> Boot From File -> CD ROM -> EFI -> BOOT -> bootx64.efi

Welcome to GRUB!

error: file `/boot/grub/arm64-efi/normal.mod' not found.
Entering rescue mode...
grub rescue>

cobracmder:
@pankajbansal3073 : Nice progress.  Give me a day or two to investigate the GRUB issue (I have some OCP business to attend to over the next day or so).

BTW, what is your build environment?  I ask because Debian Jessie doesn't have the minimum version of qemu for a cortex-a53 cpu.  Are you using Debian Jessie with Backports (I believe Qemu 2.5) or another Debian based distro?

build-config/scripts/onie-mk-iso.sh
cobracmder: Is this an issue with the compiler (gcc 5.3.0)?  Is there a newer version that resolves that issue?

cobracmder:
@pankajbansal3073, Here are some the things I ran into:
* crosstool-ng (the master branch) requires a new application to be installed (help2man); need to modify the ```debian-prepare-build-host``` target to include that package.  (I just now saw this when I created a new build machine with qemu 2.5)
* using ```grub-mkimage``` from the host yields an error (```cannot open /usr/lib/grub/arm64-efi/moddep.lst: No such file or directory.```) 

Looking at the ```onie-mk-iso.sh``` script I'm trying to figure out why isn't there a section for aarch64 uefi like there is for i386-pc for legacy bios, and x86_64 for uefi.  By changing the format to generate a UEFI format image from ```x86_64-efi``` to ```arm64-efi``` without changing some of the other areas that declare x86 images might explain the issue you are seeing.

pankajbansal3073:
@cobracmder , to solve the kernel panic problem, i compiled the onie kernel and rootfs using external toolchain which i downloaded from (https://releases.linaro.org/components/toolchain/binaries/5.3-2016.02/aarch64-linux-gnu/gcc-linaro-5.3-2016.02-x86_64_aarch64-linux-gnu.tar.xz)

i have also made changes in machine.make file to use an externally provided toolchain.

Now the kernel is booting successfully in virtual machine.

The grub rescue issue is still present. i am checking more on it.

pankajbansal3073:
@cobracmder Finally i have solved the grub install problem too.
Now the onie image is booting fine and demo OS is also being installed correctly. i have pushed the changes, plz review. 

Also i have changed many common files across architectures for aarch64. could you please check, how to refine these changes ?

cobracmder:
@pankajbansal3073, Awesome.

I'm current re-reviewing all the commits/changed files and will post my comments here. Give me a day or two.

From a cursory look:
* make sure to update all the copyright statements in the files you touch to include yourself.  If it's a new file, use the template we use and add yourself to the copyright line.
* commit d5db22d may have to be split out into various commits (updating lzo, updating efivar, updating lvm2, adding popt patches for aarch64)

cobracmder:
@pankajbansal3073 , here's what I've got so far (in conjunction with those two from above):
* missing package dependency in ```Makefile:debian-prepare-build-host``` for ```help2man```
* ```onie-mk-installer.sh```: this PR removed all x86_64 ability to copy additional installer configuration files.  Since this is a shared script for all platforms, we need it to be able to handle x86_64 and aarch64 platforms.
* ```onie-mk-iso.sh```: this PR did a number of things that break other platforms such as:
   * changed number of args
   * removed the ```GRUB_HOST_*_I386``` checks
   * this script can now only create ISOs for aarch64, we need to preserve i386 (legacy BIOS) and x86_64-efi as well
* ```images.make```: need to restore the I386 arguments as other platforms need this
* ```qemu_armv8a/machine.make```: need to not define your local CROSSBIN variable defining a path to use the linaro toolchain
* ```lvm2.make```: doesn't install correctly in dev-sysroot during build of ```qemu_armv8a```.  Take a look at the following errors:
```
/usr/bin/install -c -p -m 644 -D lvmlocal.conf /etc/lvm/lvmlocal.conf
/usr/bin/install: cannot create directory /etc/lvm: Permission denied
```

pankajbansal3073:
@cobracmder, i have updated the copyright statements in various files and made changes so that arm64 and x86_64 targets are building successfully. please check

cobracmder:
Thanks @pankajbansal3073 .  It's the start of a holiday weekend over here but I should be able to go through everything by next Thursday.  Thanks again for your hard work.

pankajbansal3073:
@cobracmder Finally i have solved the kernel panic issue with crosstool-ng toolchain too. Now aarch64 target can be built and run using either external toolchain (like linaro toolchain) or onie internal toolchain (crosstool-ng toolchain). Please review my changes and suggest if any further changes are needed or not ?

cobracmder:
Nice. :smiley:

cobracmder:
Hey @pankajbansal3073 , so I've gone through all the commits and have the following feedback.

* Everything looks nice and tighty; thanks for doing that
* We are still missing package `help2man` in the main `Makefile` for `debian-prepare-build-host`.  This package is required by using the newer crosstool-ng that is referenced for armv8a.
* `lvm2` is failing with the following error(s) which needs to be fixed:
```
/usr/bin/install -c -p -m 644 -D lvmlocal.conf /etc/lvm/lvmlocal.conf
/usr/bin/install: cannot create directory /etc/lvm: Permission denied
```
It looks like something is not being handled correctly in `lvm2-conf-install-directory-fixup.patch`.
* When using the following workflow:
```
$ make -j 4 MACHINE=qemu_armv8a download
$ make -j 4 MACHINE=qemu_armv8a all
```
I noticed that build dependencies become out of order.  For example, `GRUB` is built before `crosstool-ng`.  When building serially (`make MACHINE=qemu_armv8a all`), everything works as expected, except the added time it takes to build.
* when building `qemu_armv8a` `uClibc` is still being downloaded even though `glibc` is specified in the crosstool-ng configuration file.  This needs to be tracked down. (minor nit)

pankajbansal3073:
@cobracmder i have fixed these errors.

- help2man and other packages needed for onie compilation added to debian-prepare-build-host.

- lvm2 error fixed by updating lvm2-conf-install-directory-fixup.patch

- build dependencies is not an issue. actually grub host is compiled before compiling crosstool-ng, which doesn't require cross compiled gcc, it just needs native gcc.

- uClibc will now only be downloaded if its specified in crosstool-ng config file.

cobracmder:
Wow that was fast. I'll take a look at the changes in the morning.

cobracmder:
@pankajbansal3073 

I've gone through all the patches (looking good) and even a variety of builds.

The only nit I have is from the following workflow:
* build x86_64, ARM32, ARM64, or PPC platform (like kvm_x86_64)
* build another platform with different CPU arch, without doing a `make distclean` (repeat)

Once I build the aarch64 platform (qemu_armv8a), regardless of order, I must now perform a `distclean`.

For example, in this snippet, I built `qemu_armv8a` first and then attempted to build `kvm_x86_64` and got the following:
```
==== Extracting upstream BusyBox ====
Error: tarball does not exist: /home/carlos/working/onie/build/download/busybox-1.20.2.tar.bz2
/home/carlos/working/onie/build-config/scripts/extract-package: Usage
/home/carlos/working/onie/build-config/scripts/extract-package <package build dir> <tarball>

Extract the <tarball> into the specified <build directory>.
mkdir: created directory ‘/home/carlos/working/onie/build/kvm_x86_64-r0/sysroot’
make/busybox.make:66: recipe for target '/home/carlos/working/onie/build/kvm_x86_64-r0/stamp/busybox-source' failed
```

I don't know how serious an impact that would be for folks doing mass builds, but it's something to be aware of.  I think the next step is to get the ONIE Meister, @cbrune, to review this PR.

pankajbansal3073:
@cobracmder This is expected. This problem is occuring because the build is controlled by stamps and the download directory for any arch is same i.e. build/download. So when qemu_armv8a build is done, it downloads busybox 1.22 and creates download stamp in same directory. When other arch build is triggered which needs busybox 1.20, it finds busybox download stamp in download directory, therefore it doesn't attempt to download busybox again. This problem can be solved if the download stamps are stored in project specific directory i.e. build/qemu_armv8a/stamp, and not in download directory.

cbrune:
I think the easiest/best way to solve this is to add the package version to the download stamp name.  This should apply to *all* download stamps in the ONIE project.  For example the `BUSYBOX_DOWNLOAD_STAMP` could be changed from:

```
BUSYBOX_DOWNLOAD_STAMP  = $(DOWNLOADDIR)/busybox-download
```

to:

```
BUSYBOX_DOWNLOAD_STAMP  = $(DOWNLOADDIR)/busybox-download-$(BUSYBOX_VERSION)
```

pankajbansal3073:
@cobracmder @cbrune Fixed.

cobracmder:
@pankajbansal3073 nice.  Going to kick the tires on it.

cobracmder:
The latest commit fixes that workflow problem.

build-config/make/busybox.make
```diff
@@ -4,6 +4,7 @@
 #  Copyright (C) 2014 david_yang <david_yang@accton.com>
 #  Copyright (C) 2014 Mandeep Sandhu <mandeep.sandhu@cyaninc.com>
 #  Copyright (C) 2014 Nikolay Shopik <shopik@inblock.ru>
+#  Copyright (C) 2016 Pankaj Bansal <pankajbansal3073@gmail.com>
 #
```
cbrune: Moving the version of busybox forward is generally a good thing.

My concern here is that the new busybox version introduced here (1.24.2) does not include all the feature patches from the older version.  We had this same problem for u-boot, where folks wanted to use different versions of u-boot.

We have enhanced busybox with a number of feature patches.  These patches should also be applied to a newer version.  For example any ONIE compatible system will need all the "feature-onie-syseeprom" patches.

For u-boot we solved this by having a "common" patch directory that applied to all u-boot versions.  Otherwise we have to maintain patches for both busybox versions separately.  That is painful.

build-config/make/crosstool-ng.make
```diff
@@ -43,24 +45,34 @@ crosstool-ng-download: $(CROSSTOOL_NG_DOWNLOAD_STAMP)
...
 	$(Q) rm -f $@ && eval $(PROFILE_STAMP)
 	$(Q) echo "==== Extracting upstream $(CROSSTOOL_NG_DESC) ===="
-	$(Q) $(SCRIPTDIR)/extract-package $(CROSSTOOL_NG_BUILD_DIR) $(DOWNLOADDIR)/$(CROSSTOOL_NG_TARBALL)
+	$(Q) if [ $(shell awk 'BEGIN{ print "'$(CROSSTOOL_NG_VERSION)'"<="'$(CROSSTOOL_NG_LATEST_RELEASE_VERSION)'" }') -eq 1 ]; then \
+			$(SCRIPTDIR)/extract-package $(CROSSTOOL_NG_BUILD_DIR) \
+			$(DOWNLOADDIR)/$(CROSSTOOL_NG_TARBALL); \
+	     else \
+			git clone https://github.com/crosstool-ng/crosstool-ng \
+			$(CROSSTOOL_NG_DIR); \
+	     fi
```
cbrune: I understand what you are doing here, but just grabbing "master" will likely lead to pain and misery.  It will change out from under you at some point and cause trouble.

We have the same problem with the GRUB project -- they do not make releases.  All one can really do is grab a copy of master from some point in time and base your work on that.  I think a similar approach can work here as well.

For example the master branch of crosstool-ng is at commit id 6e7c6165.  We can fetch a tarball based on that commit id like this:

```
$ wget https://github.com/crosstool-ng/crosstool-ng/archive/6e7c6165.tar.gz
$ ls -l 
-rw-r--r-- 1 curt Development 1930561 Jul 25 07:48 6e7c6165.tar.gz
```

Then we say the CROSSTOOL_NG_VERSION is "6e7c6165" and change the URL to point at github instead of crosstool-ng.org.  Also we can upload the tarball to the ONIE mirror.

The point is -- treating everything as a tarball will simplify things.  Having a special case to "git clone" is no longer be needed.

build-config/make/xtools.make
```diff
@@ -21,17 +19,22 @@ GCC_VERSION ?= 4.9.2
...
-XTOOLS_CONFIG		= conf/crosstool/gcc-$(GCC_VERSION)/uClibc-$(UCLIBC_VERSION)/crosstool.$(ONIE_ARCH).config
+# Default GCC version to build for the toolchain
+GCC_VERSION ?= 4.9.2
+
+XTOOLS_LIBC ?= uClibc
+XTOOLS_LIBC_VERSION ?= 0.9.33.2
+XTOOLS_CONFIG		?= conf/crosstool/gcc-$(GCC_VERSION)/$(XTOOLS_LIBC)-$(XTOOLS_LIBC_VERSION)/crosstool.$(ONIE_ARCH).config
 XTOOLS_ROOT		= $(BUILDDIR)/x-tools
-XTOOLS_VERSION		= $(ONIE_ARCH)-g$(GCC_VERSION)-lnx$(LINUX_RELEASE)-u$(UCLIBC_VERSION)
+XTOOLS_VERSION		= $(ONIE_ARCH)-g$(GCC_VERSION)-lnx$(LINUX_RELEASE)-$(XTOOLS_LIBC)-$(XTOOLS_LIBC_VERSION)
```
cbrune: This looks like a nice addition.

patches/crosstool-NG/1.22.1/series
```diff
@@ -0,0 +1 @@
+Use-awk-hardlink.patch
```
cbrune: What problem does this patch solve?  The stg patch header should contain the use information about why the patch is needed and what it is solving.

build-config/Makefile
```diff
@@ -288,10 +289,14 @@ endif
...
 ONIE_MIRROR	?= http://mirror.opencompute.org/onie
 
-include make/crosstool-ng.make
 include make/kernel-download.make
-include make/uclibc-download.make
-include make/xtools.make
+ifneq ($(XTOOLS_LIBC),glibc)
+  include make/uclibc-download.make
+endif
+ifneq ($(XTOOLS_ENABLE),no)
```
cbrune: I am not a fan of double negatives.  Can this be written as:
```
ifeq ($(XTOOLS_LIBC),uclibc)
```

and

```
ifeq ($(XTOOLS_ENABLE),yes)
```

Also does it make sense to have the XTOOLS_LIBC checkout outside of the XTOOLS_ENABLE check?  Meaning, does the setting of XTOOLS_LIBC matter at all if XTOOLS_ENABLE is disabled?

build-config/Makefile
```diff
@@ -417,7 +426,8 @@ DEBIAN_BUILD_HOST_PACKAGES	= build-essential stgit u-boot-tools util-linux \
 				  gperf device-tree-compiler python-all-dev xorriso \
 				  autoconf automake bison flex texinfo libtool libtool-bin \
 				  realpath gawk libncurses5 libncurses5-dev bc \
-				  dosfstools mtools
+				  dosfstools mtools git gawk wget help2man libexpat1 \
```
cbrune: gawk is already in the list, two lines above.

build-config/Makefile
```diff
@@ -317,7 +322,11 @@ ifeq ($(PARTED_ENABLE),yes)
   include make/parted.make
 endif
 ifeq ($(GRUB_ENABLE),yes)
-  include make/grub.make
+  ifeq ($(ARCH),arm64)
+    include make/grub_aarch64.make
+  else
+    include make/grub.make
+  endif
```
cbrune: This is ugly including a separate arch specific grub.make at this level.  i would put the necessary aarach64 changes in grub.make.  However, I haven't go to the grub changes -- maybe what you have here will make sense after I read that section.

build-config/arch/armv8a/busybox.config
```diff
@@ -0,0 +1,1069 @@
+#
+# Automatically generated make config: don't edit
+# Busybox version: 1.24.2
+# Sun May 15 21:25:58 2016
```
cbrune: I'm not grokking why we want an architecture specific busybox.config file.  ONIE should behave the same on all architectures.  This will be difficult to maintain.  Again it may make sense once I read through all the patches.

build-config/arch/armv8a/crosstool.armv8a.config
```diff
@@ -0,0 +1,559 @@
+#
+# Automatically generated file; DO NOT EDIT.
+# Crosstool-NG Configuration
+#
```
cbrune: Same comment about the kernel configuration change below.  We already have a location to store that type of configuration:

```
build-config/kernel/<kernel-version>/linux.<cpu-arch>.config
```

build-config/make/images.make
```diff
@@ -120,17 +121,35 @@ CHECKDIR	= $(CHECKROOT)/checkdir
...
+endif
 
 ifeq ($(EXT3_4_ENABLE),yes)
-SYSROOT_LIBS	+= \
-		  libdl.so.0 libdl-$(UCLIBC_VERSION).so \
-		  libpthread.so.0 libpthread-$(UCLIBC_VERSION).so
+  ifeq ($(XTOOLS_LIBC),uClibc)
+    SYSROOT_LIBS	+= \
+	     	          libdl.so.0 libdl-$(UCLIBC_VERSION).so \
+		  	  libpthread.so.0 libpthread-$(UCLIBC_VERSION).so
```
cbrune: Can we use `$(XTOOL_LIBC_VERSION)` here also?  Then this collapses down.  No need to check if it's uClibc or not, just use `$(XTOOLS_LIBC_VERSION)`.

build-config/make/images.make
```diff
@@ -183,16 +208,18 @@ $(SYSROOT_CHECK_STAMP): $(PACKAGES_INSTALL_STAMPS)
...
-	     $(CROSSBIN)/$(CROSSPREFIX)populate -r $(DEV_SYSROOT) \
-		-s $(SYSROOTDIR) -d $(CHECKDIR) && \
-		(cd $(SYSROOTDIR) && find . | LC_ALL=C sort > $(SYSFILES)) && \
-		(cd $(CHECKDIR) && find . | LC_ALL=C sort > $(CHECKFILES)) && \
-		diff -q $(SYSFILES) $(CHECKFILES) > /dev/null 2>&1 || { \
-			(echo "ERROR: Missing files in SYSROOTDIR:" && \
-			 diff $(SYSFILES) $(CHECKFILES) ; \
-			 false) \
-		}
+    ifneq ($(XTOOLS_ENABLE),no)
```
cbrune: can we check for XTOOLS_ENABLE being "yes"?  not a fan of double negatives.

build-config/make/images.make
```diff
@@ -378,7 +405,11 @@ $(RECOVERY_INITRD_STAMP): $(IMAGE_UPDATER_STAMP)
 	$(Q) touch $@
 
 ifeq ($(UEFI_ENABLE),yes)
-  RECOVERY_XORRISO_OPTIONS = $(RECOVERY_CONF_DIR)/xorriso-options-uefi.cfg
+  ifeq ($(ARCH),arm64)
```
cbrune: Here I would not check for $(ARCH).
I would rename the original `xorriso-options-uefi.cfg` to now contain the arch, something like:
```
xorriso-options-x86_64.cfg
```
Then this file can simply refer to `xorriso-options-$(ARCH)-uefi.cfg`.

build-config/make/images.make
```diff
@@ -387,20 +418,36 @@ recovery-iso: $(RECOVERY_ISO_STAMP)
 $(RECOVERY_ISO_STAMP): $(GRUB_HOST_INSTALL_STAMP) $(RECOVERY_INITRD_STAMP) \
 			$(RECOVERY_CONF_DIR)/grub-iso.cfg $(RECOVERY_XORRISO_OPTIONS)
 	$(Q) echo "==== Create $(MACHINE_PREFIX) ONIE Recovery Hybrid iso ===="
-	$(Q) Q=$(Q) CONSOLE_SPEED=$(CONSOLE_SPEED) \
+    ifeq ($(ARCH),arm64)
```
cbrune: Like the above I think this all would collapse down if you rename the original `onie-mk-iso.sh` to `onie-mk-iso-x86_64.sh`.  Then this file can refer to `onie-mk-iso-$(ARCH).sh`.

I haven't read the patches yet, but I do not understand why a separate `onie-mk-iso.sh` is necessary at all.

If the differences are minor another suggestion is to only have one script, `onie-mk-iso.sh`, but pass in the $(ARCH) as a parameter and take appropriate action within the script.

build-config/recovery/xorriso-options-arm64-uefi.cfg
```diff
@@ -0,0 +1,31 @@
+# This file is derived from https://github.com/opencomputeproject/onie/blob/master/build-config/recovery/xorriso-options-uefi.cfg
```
cbrune: what is different between this file and the x86_64 version?

build-config/scripts/onie-arm64-mk-iso.sh
```diff
@@ -0,0 +1,193 @@
+#!/bin/sh
```
cbrune: what is different between this file and the x86_64 version?

demo/installer/armv8a/install.sh
```diff
@@ -0,0 +1,422 @@
...
+    local demo_mnt="$1"
+    local blk_dev="$2"
+
+    # get running machine from conf file
+    [ -r /etc/machine.conf ] && . /etc/machine.conf
+
+    if [ "$onie_firmware" = "coreboot" ] ; then
+        local grub_target="i386-coreboot"
+        local core_img="$demo_mnt/grub/$grub_target/core.elf"
+    else
```
cbrune: Does this else case make sense for aarch64?

demo/installer/armv8a/install.sh
```diff
@@ -0,0 +1,422 @@
+#!/bin/sh
```
cbrune: This whole file looks very similar to the x86_64 file.

If the installer files are very similar between x86_64 and aarch64 then I propose doing what we did for powerpc and arm32 (which are similar).  In that case we called them "u-boot-arch".

In this case perhaps we can go with "grub-arch" as both x86_64 and aarch64 appear to be GRUB based.

cbrune: Otherwise, having a copy that is nearly identical is a nightmare to maintain.

demo/os/armv8a/bin/embed
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
cbrune: Same comment -- these all look like the x86_64 versions.  Suggest to combine into a common "grub-arch".

cbrune: The same comment applies to the demo OS files here and to the "installer" files below as well.

rootconf/armv8a/sysroot-bin/onie-boot-mode
```diff
@@ -0,0 +1,123 @@
+#!/bin/sh
```
cbrune: same comment -- can these aarch64 rootconfig files be combined with x86_64 and called "grub-arch" ?

tools/armv8a/bin/onie-version
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
cbrune: same comment here -- can this be combined into a "grub-arch" ?

patches/kexec-tools/15-25-kexec-tools-add-aarch64-support.patch
```diff
@@ -0,0 +1,828 @@
+Subject: [PATCH 1/1] From: Geoff Levand <geoff@infradead.org> Date: Mon, 15
```
cbrune: This looks great.

build-config/make/grub_aarch64.make
```diff
@@ -0,0 +1,223 @@
+#-------------------------------------------------------------------------------
```
cbrune: how is this file different from grub.make?  Having this copy will be a huge pain to maintain.  It will be better if we can keep everything in grub.make and work out any differences there.

build-config/make/kernel-download.make
```diff
@@ -14,10 +15,10 @@
...
 # Need the Linux kernel downloaded before building xtools
 
-LINUX_VERSION		?= 4.1
-LINUX_MAJOR_VERSION	= $(firstword $(subst ., ,$(LINUX_VERSION)))
-LINUX_MINOR_VERSION	?= 23
-LINUX_RELEASE		?= $(LINUX_VERSION).$(LINUX_MINOR_VERSION)
+LINUX_VERSION		?= 4.1.23
+LINUX_MAJOR_VERSION	= $(shell echo $(LINUX_VERSION) | awk 'BEGIN { FS="."; } {print $$1;}')
+LINUX_MINOR_VERSION	= $(shell echo $(LINUX_VERSION) | awk 'BEGIN { FS="."; } {print $$3;}')
+LINUX_RELEASE		?= $(LINUX_VERSION)
```
cbrune: What problem is this solving?  I do not see the difference.

build-config/make/mtdutils.make
```diff
@@ -9,7 +10,8 @@
 # This is a makefile fragment that defines the build of mtdutils
 #
 
-MTDUTILS_VERSION	= 1.5.0
+MTDUTILS_VERSION	?= 1.5.0
+MTDUTILS_LATEST_RELEASE_VERSION = 1.5.2
```
cbrune: This all seems a bit complicated.  Perhaps we just move the version of the mtdutils forward to 1.5.2 for all machines and architectures?  We just have to make sure all the machines continue to compile and build properly.

Moving forward to a newer version of a software package is generally a good thing.

build-config/make/mtdutils.make
```diff
@@ -36,17 +38,24 @@ mtdutils-download: $(MTDUTILS_DOWNLOAD_STAMP)
...
 SOURCE += $(MTDUTILS_SOURCE_STAMP)
 mtdutils-source: $(MTDUTILS_SOURCE_STAMP)
 $(MTDUTILS_SOURCE_STAMP): $(TREE_STAMP) | $(MTDUTILS_DOWNLOAD_STAMP)
 	$(Q) rm -f $@ && eval $(PROFILE_STAMP)
 	$(Q) echo "==== Extracting upstream mtdutils ===="
-	$(Q) $(SCRIPTDIR)/extract-package $(MTDUTILS_BUILD_DIR) $(DOWNLOADDIR)/$(MTDUTILS_TARBALL)
+	$(Q) if [ $(shell awk 'BEGIN{ print "'$(MTDUTILS_VERSION)'"<="'$(MTDUTILS_LATEST_RELEASE_VERSION)'" }') -eq 1 ]; then \
+			$(SCRIPTDIR)/extract-package $(MTDUTILS_BUILD_DIR) $(DOWNLOADDIR)/$(MTDUTILS_TARBALL); \
+		 else \
+			git clone git://git.infradead.org/mtd-utils.git \
```
cbrune: Same comments about cloning master -- this will bite us eventually.  Let's pick a release (a commit) and make a tarball of it.  Same comments I made on the crosstool-ng.

machine/qemu_armv8a/INSTALL
```diff
@@ -0,0 +1,48 @@
+==============================================
```
cbrune: This is great -- having a virtual machine implementation for this CPU architecture is wonderful.

build-config/make/lvm2.make
```diff
@@ -1,6 +1,7 @@
 #-------------------------------------------------------------------------------
```
cbrune: This is great.  Generally moving software packages forward is a good thing.

I can upload the 2_02_155 version to the ONIE mirror.

patches/popt/aarch64_support.patch
```diff
@@ -0,0 +1,733 @@
+#-------------------------------------------------------------------------------
+#
+#  Copyright (C) 2016 Pankaj Bansal <pankajbansal3073@gmail.com>
+#
+#  SPDX-License-Identifier:     GPL-2.0
```
cbrune: This looks great.

The stg patch comment should indicate where this patch came from.  I'm assuming you back ported this patch from somewhere.

build-config/make/lzo.make
```diff
@@ -9,7 +10,7 @@
 # This is a makefile fragment that defines the build of lzo
 #
 
-LZO_VERSION		= 2.06
+LZO_VERSION		= 2.09
```
cbrune: Generally fine to move packages forward.

build-config/make/efivar.make
```diff
@@ -82,7 +83,11 @@ $(EFIVAR_INSTALL_STAMP): $(SYSROOT_INIT_STAMP) $(EFIVAR_BUILD_STAMP)
 		cp -av $(DEV_SYSROOT)/usr/bin/$$file $(SYSROOTDIR)/usr/bin/ ; \
 	     done
 	$(Q) for file in $(EFIVAR_LIBS); do \
-		cp -av $(DEV_SYSROOT)/usr/lib/$$file $(SYSROOTDIR)/usr/lib/ ; \
+		if [ "$(ARCH)" == "arm64" ]; then \
+		    cp -av $(DEV_SYSROOT)/usr/lib64/$$file $(SYSROOTDIR)/usr/lib/ ; \
+		else \
+		    cp -av $(DEV_SYSROOT)/usr/lib/$$file $(SYSROOTDIR)/usr/lib/ ; \
+		fi \
```
cbrune: looks good.

build-config/make/acpica-tools.make
```diff
@@ -15,7 +15,7 @@ ACPICA_TOOLS_TARBALL_URLS	+= $(ONIE_MIRROR) https://acpica.org/sites/acpica/file
 ACPICA_TOOLS_BUILD_DIR	= $(MBUILDDIR)/acpica-tools
 ACPICA_TOOLS_DIR		= $(ACPICA_TOOLS_BUILD_DIR)/acpica-unix-$(ACPICA_TOOLS_VERSION)
 
-ACPICA_TOOLS_DOWNLOAD_STAMP	= $(DOWNLOADDIR)/acpica-tools-download
+ACPICA_TOOLS_DOWNLOAD_STAMP	= $(DOWNLOADDIR)/acpica-tools-$(ACPICA_TOOLS_VERSION)-download
```
cbrune: This is cool.  Long overdue.  Thanks.

cbrune:
OK, so I made a ton of comments.  One theme, which I like, is you moved the version of some packages forward.  From my list:
- crosstool-ng (pull tarball based on commit id, instead of master)
- mdtutils
- grub
- lvm2
- lzo

For each of these I can upload a tarball to the ONIE_MIRROR when you are ready.  Let me know the versions.

build-config/make/busybox.make
pankajbansal3073: I tried applying busybox 1.20.2 patches to busybox 1.25.0, but i am facing many compilation errors with this. i suggest that you please merge these changes to busybox 1.25.0 and update the busybox version.

build-config/make/crosstool-ng.make
pankajbansal3073: The master branch is not working correctly. we need master branch + (https://github.com/crosstool-ng/crosstool-ng/pull/410/commits) for crosstool-ng. if you can make a tarball of this and upload on onie server, then arm64 builds can work OK

patches/crosstool-NG/1.22.1/series
pankajbansal3073: if you take https://github.com/crosstool-ng/crosstool-ng/pull/410/commits for crosstool-ng, then this patch is not needed.

build-config/Makefile
pankajbansal3073: double negatives fixed.
Since Crosstools and uClibc are downloaded seprately, if we don't do this check, then uClibc will be downloaded, even if glibc is selected in xtools. This was a concern of @cobracmder .

build-config/Makefile
pankajbansal3073: fixed

build-config/Makefile
pankajbansal3073: grub_aarch64 removed. made changes for arm64 and x86_64 in same grub.make

build-config/arch/armv8a/busybox.config
pankajbansal3073: kept a common busybox 1.25.0 version config

build-config/arch/armv8a/crosstool.armv8a.config
pankajbansal3073: fixed these locations

build-config/make/images.make
pankajbansal3073: fixed

build-config/make/images.make
pankajbansal3073: fixed. Now the bios cfg and uefi cfg are seprated and will be applied based on firmware_type bios or uefi, these both will be merged and applied if firmware_type is selected as auto.

build-config/make/images.make
pankajbansal3073: Passed the ARCH as parameter and kept the file common for x86_64 and arm64

build-config/recovery/xorriso-options-arm64-uefi.cfg
pankajbansal3073: fixed. Now the bios cfg and uefi cfg are seprated and will be applied based on firmware_type bios or uefi, these both will be merged and applied if firmware_type is selected as auto.

build-config/scripts/onie-arm64-mk-iso.sh
pankajbansal3073: Passed the ARCH as parameter and kept the file common for x86_64 and arm64

demo/installer/armv8a/install.sh
pankajbansal3073: installer files common for x86_64 and arm64 as grub-arch

demo/installer/armv8a/install.sh
pankajbansal3073: installer files common for x86_64 and arm64 as grub-arch

demo/os/armv8a/bin/embed
pankajbansal3073: made files common for x86_64 and arm64 as grub-arch

rootconf/armv8a/sysroot-bin/onie-boot-mode
pankajbansal3073: made files common for x86_64 and arm64 as grub-arch

tools/armv8a/bin/onie-version
pankajbansal3073: made files common for x86_64 and arm64 as grub-arch

build-config/make/grub_aarch64.make
pankajbansal3073: grub_aarch64 removed. made changes for arm64 and x86_64 in same grub.make

build-config/make/kernel-download.make
pankajbansal3073: removed these changes

build-config/make/mtdutils.make
pankajbansal3073: please upload the latest mtdutils commit tar mtd-utils-e4c8885.tar.gz to onie server

build-config/make/mtdutils.make
pankajbansal3073: please upload the latest mtdutils commit tar mtd-utils-e4c8885.tar.gz to onie server

pankajbansal3073:
@cobracmder @cbrune i have made the changes as you suggested. The only problem i am facing is in busybox. as i am not familiar with the syseeprom patches you made for busybox 1.20.2, i am not able to fix the compilation errors in busybox 1.25.0. please merge these patches and upgrade the busybox version to 1.25.0 for arm64 to compile

build-config/make/mtdutils.make
cbrune: OK.  It looks like e4c8885 is the current HEAD of the master branch:

build-config/make/lvm2.make
cbrune: I uploaded lvm-2_02_155 to the ONIE mirror:
```
monster-15:/tmp/goo$ wget http://mirror.opencompute.org/onie/lvm2-2_02_155.tar.xz
--2016-08-01 08:42:59--  http://mirror.opencompute.org/onie/lvm2-2_02_155.tar.xz
Resolving mirror.opencompute.org (mirror.opencompute.org)... 23.253.235.92
Connecting to mirror.opencompute.org (mirror.opencompute.org)|23.253.235.92|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1469316 (1.4M) [application/x-xz]
Saving to: `lvm2-2_02_155.tar.xz'

100%[====================================================================================>] 1,469,316   3.38M/s   in 0.4s    

2016-08-01 08:42:59 (3.38 MB/s) - `lvm2-2_02_155.tar.xz' saved [1469316/1469316]
```

The sha1sum is:
```
monster-15:/tmp/goo$ sha1sum lvm2-2_02_155.tar.xz 
e1dd87307a0127e3f1469408051fed2a6add2922  lvm2-2_02_155.tar.xz
```

Let me know if you agree with the version and sha1sum.

build-config/make/lzo.make
cbrune: I uploaded lzo 2.09 to the ONIE mirror:
```
monster-15:/tmp/goo$ wget http://mirror.opencompute.org/onie/lzo-2.09.tar.gz
--2016-08-01 09:03:45--  http://mirror.opencompute.org/onie/lzo-2.09.tar.gz
Resolving mirror.opencompute.org (mirror.opencompute.org)... 23.253.235.92
Connecting to mirror.opencompute.org (mirror.opencompute.org)|23.253.235.92|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 594855 (581K) [application/x-gzip]
Saving to: `lzo-2.09.tar.gz'

100%[====================================================================================>] 594,855     1.96M/s   in 0.3s    

2016-08-01 09:03:46 (1.96 MB/s) - `lzo-2.09.tar.gz' saved [594855/594855]
```

and the sha1 is:
```
monster-15:/tmp/goo$ sha1sum lzo-2.09.tar.gz 
e2a60aca818836181e7e6f8c4f2c323aca6ac057  lzo-2.09.tar.gz
```

Let me know if you agree with this version and sha1.

build-config/make/mtdutils.make
pankajbansal3073: its good, you can update

build-config/make/lvm2.make
pankajbansal3073: its good, you can update

build-config/make/lzo.make
pankajbansal3073: its good, you can update

pankajbansal3073:
@cobracmder @cbrune please take crosstool-ng source code till commit https://github.com/crosstool-ng/crosstool-ng/commit/11cb2ddd43fd1ff493f4b7cd63f1cf654294165f and upload it on onie server.

pankajbansal3073:
Hello @cobracmder @cbrune 

i have updated my patches. can you please review these patches again ?

rootconf/grub-arch/sysroot-lib-onie/init-arch
```diff
@@ -85,7 +85,16 @@ UUID=$uefi_uuid	$uefi_esp_mnt	$uefi_esp_fs_type	defaults,rw,errors=remount-ro	0
...
+    
+    local disk=$(onie_get_boot_disk)
+    local uefi_part=$(onie_get_uefi_partition)
+	
+    # check if we have an mmcblk device
+    local blk_suffix=
+    echo ${disk} | grep -q mmcblk && blk_suffix="p"
+	
+    local uefi_dev="${disk}$blk_suffix$uefi_part"
+
```
david56: The mmc device name returned from onie_get_boot_disk() also has the suffix.

rootconf/grub-arch/sysroot-lib-onie/onie-blkdev-common
```diff
@@ -48,8 +48,10 @@ onie_get_boot_dev()
 onie_get_boot_disk()
 {
     local disk=$(onie_get_boot_dev)
-    # Strip off the numbers
-    disk="${disk%%[0-9]*}"
+    # Strip off the partition numbers from end
+    disk="${disk%%[0-9]}"
+    # remove partition suffix p in case of MMC card
+    echo ${disk} | grep -q mmcblk && disk="${disk%%p}"
```
david56: The same comment above.

rootconf/grub-arch/sysroot-lib-onie/onie-blkdev-common
```diff
@@ -82,8 +84,12 @@ onie_get_uefi_uuid()
     local disk=$(onie_get_boot_disk)
     local uefi_part=$(onie_get_uefi_partition)
 
+    # check if we have an mmcblk device
+    local blk_suffix=
+    echo ${disk} | grep -q mmcblk && blk_suffix="p"
```
david56: The same comment above.

build-config/make/mtdutils.make
```diff
@@ -38,8 +39,8 @@ $(MTDUTILS_DOWNLOAD_STAMP): $(PROJECT_STAMP)
 	$(Q) rm -f $@ && eval $(PROFILE_STAMP)
 	$(Q) echo "==== Getting upstream mtdutils ===="
 	$(Q) $(SCRIPTDIR)/fetch-package $(DOWNLOADDIR) $(UPSTREAMDIR) \
-		$(MTDUTILS_COMMIT).tar.gz $(MTDUTILS_TARBALL_URLS)
-	$(Q) cd $(DOWNLOADDIR) && ln -fs $(MTDUTILS_COMMIT).tar.gz $(MTDUTILS_TARBALL)
+					$(MTDUTILS_COMMIT).tar.gz $(MTDUTILS_TARBALL_URLS)
+	     cd $(DOWNLOADDIR) && ln -fs $(MTDUTILS_COMMIT).tar.gz $(MTDUTILS_TARBALL)
```
david56: It looks that no changes here except `$(Q)`.

demo/installer/grub-arch/install.sh
```diff
@@ -72,9 +72,12 @@ create_demo_gpt_partition()
     # Find next available partition
     last_part=$(sgdisk -p $blk_dev | tail -n 1 | awk '{print $1}')
     demo_part=$(( $last_part + 1 ))
+    # check if we have an mmcblk device
+    blk_suffix=
+    echo ${blk_dev} | grep -q mmcblk && blk_suffix="p"
 
     # Create new partition
-    echo "Creating new demo partition ${blk_dev}$demo_part ..."
+    echo "Creating new demo partition ${blk_dev}$blk_suffix$demo_part ..."
```
david56: Sorry.  It's my fault.  It needs the suffix here.  BTW, I found it has issue in the statement of extrating `blk_dev`

    blk_dev=$(blkid | grep ONIE-BOOT | awk '{print $1}' |  sed -e 's/[1-9][0-9]*:.*$//' | sed -e 's/\([0-9]\)\(p\)/\1/' | head -n 1)

where `sed -e 's/\([0-9]\)\(p\)/\1/'` would not remove partition number correctly.  For instance:

    # blk_dev=/dev/mmcblk0p1
    # echo $blk_dev | sed -e 's/\([0-9]\)\(p\)/\1/'
    /dev/mmcblk01

rootconf/grub-arch/sysroot-lib-onie/init-arch
david56: Please skip the comment.  Sorry.

demo/installer/grub-arch/install.sh
david56: I rechecked it again.  It is not an issue.  :smiley:

build-config/scripts/onie-mk-demo.sh
```diff
@@ -25,6 +25,10 @@ if [ "$arch" = "powerpc-softfloat" -o "$arch" = "armv7a" ] ; then
     # Both of these architectures share common installer code as they
     # are both based on u-boot.
     arch_dir="u-boot-arch"
+elif [ "$arch" = "x86_64" -o "$arch" = "armv8a" ] ; then
+    # These architectures share common installer code as they
+    # are based on grub.
+    arch_dir="grub-arch"
```
cbrune: @david56 - this PR is for adding aarch64 (armv8a).  we do not have any armv8a machines in the project yet.  are you saying you expect some future aarch64 machines to use u-boot?

i would really like to avoid that situation -- some aarch64 using u-boot and some using grub.  i would expect all aarch64 machines to be UEFI and grub.

david56: > are you saying you expect some future aarch64 machines to use u-boot?

Yes.  As I know, the Freescale's aarch64 demo board is using u-boot.

pankajbansal3073:
@david56 @cobracmder @cbrune 
Code updated as per comments

build-config/scripts/onie-mk-demo.sh
cbrune: @david56 , @pankajbansal3073 -- well this is a fundamental issue to me.  Either armv8a (aarch64) is going to be GRUB or it will be U-Boot.  It cannot be either.

powerpc -- is only U-BOOT
x86 -- is only GRUB (though could be "bios" or "uefi")

This is a big topic and I will raise the awareness on the mailing list shortly.  I only want ONIE to support UEFI/GRUB for aarch64.  That is the modern approach, what servers are doing.

build-config/conf/busybox.config
```diff
@@ -1,7 +1,7 @@
 #
 # Automatically generated make config: don't edit
-# Busybox version: 1.20.2
-# Wed Oct 29 14:52:19 2014
+# Busybox version: 1.25.1
+# Wed Oct 12 03:08:41 2016
```
cbrune: This new busybox config turns on a lot of features that were previously disabled (or did not exist).

Generally growing the size of ONIE is not a problem, except for some of the older powerpc platforms.

In particular does the accton/accton_as5610_52x machine build OK with this configuration?

I saw you added a mechanism for a machine to override the busyconf, which is how the older machines could be supported.

build-config/conf/crosstool/gcc-4.9.2/uClibc-0.9.33.2/crosstool.armv7a.config
```diff
@@ -17,28 +17,31 @@ CT_MODULES=y
...
 CT_REMOVE_DOCS=y
 CT_INSTALL_DIR_RO=y
-CT_STRIP_ALL_TOOLCHAIN_EXECUTABLES=y
+CT_STRIP_HOST_TOOLCHAIN_EXECUTABLES=y
+CT_STRIP_TARGET_TOOLCHAIN_EXECUTABLES=y
 
 #
 # Downloading
 #
-CT_FORBID_DOWNLOAD=y
```
cbrune: This comment applies to all the architectures.

CT_FORBID_DOWNLOAD must be set to 'y'.

This is an old argument going back to July 2014.  See commit 749bd360896686c7edfec405b98495e184d99906.

The gist of the problem:   CT did not validate checksums on tarballs.  so we download the tarballs and verify checksums outside of CT and then tell CT to use the local tarballs.  with that approach we never want CT to download files directly.

build-config/conf/crosstool/gcc-4.9.2/uClibc-0.9.33.2/crosstool.armv7a.config
```diff
@@ -259,30 +253,19 @@ CT_BINUTILS_binutils=y
 #
 # GNU binutils
 #
+# CT_BINUTILS_CUSTOM is not set
+CT_BINUTILS_VERSION="2.24"
 # CT_CC_BINUTILS_SHOW_LINARO is not set
-CT_BINUTILS_V_2_25=y
```
cbrune: Why is the binutils version changing for every architecture?

build-config/conf/crosstool/gcc-4.9.2/uClibc-0.9.33.2/crosstool.armv7a.config
```diff
@@ -358,68 +333,26 @@ CT_LIBC_XLDD=y
...
-CT_CC_GCC_4_6_or_later=y
-CT_CC_GCC_4_7_or_later=y
+CT_CC_GCC_CUSTOM=y
+CT_CC_GCC_CUSTOM_LOCATION="${DOWNLOADDIR}/gcc-${GCC_VERSION}.tar.bz2"
+CT_CC_GCC_CUSTOM_VERSION="${GCC_VERSION}"
+CT_CC_GCC_VERSION="${GCC_VERSION}"
 CT_CC_GCC_4_8_or_later=y
-CT_CC_GCC_4_9=y
 CT_CC_GCC_4_9_or_later=y
+CT_CC_GCC_5_or_later=y
```
cbrune: why is GCC_5_or_later set when this file is for gcc 4.9.2 ?

build-config/recovery/xorriso-options-uefi.cfg
```diff
@@ -1,6 +1,7 @@
+# This file is derived from https://github.com/opencomputeproject/onie/blob/master/build-config/recovery/xorriso-options-uefi.cfg
 #-------------------------------------------------------------------------------
 #
-#  Copyright (C) 2015 Curt Brune <curt@cumulusnetworks.com>
```
cbrune: you can't remove my copyright.

build-config/scripts/onie-mk-demo.sh
```diff
@@ -21,13 +21,7 @@ if  [ ! -d $installer_dir ] || \
...
 fi
 
-if [ "$arch" = "powerpc-softfloat" -o "$arch" = "armv7a" ] ; then
-    # Both of these architectures share common installer code as they
-    # are both based on u-boot.
-    arch_dir="u-boot-arch"
-else
-    arch_dir="$arch"
-fi
+    arch_dir="$rootfs_arch"
```
cbrune: looks like a whitespace problem here

build-config/scripts/onie-mk-installer.sh
```diff
@@ -45,13 +46,7 @@ fi
...
 }
 
-if [ "$arch" = "powerpc-softfloat" -o "$arch" = "armv7a" ] ; then
-    # Both of these architectures share common installer code as they
-    # are both based on u-boot.
-    arch_dir="u-boot-arch"
-else
-    arch_dir="$arch"
-fi
+    arch_dir="$rootfs_arch"
```
cbrune: looks like a whitespace problem here

build-config/scripts/onie-mk-tools.sh
```diff
@@ -26,8 +26,10 @@ shift 4
 #    located within the ONIE repo $tools_dir.  These tools are *not*
 #    present in the ONIE runtime image.
 
-[ -d "${tools_dir}/${arch}" ] || {
-    echo "ERROR: arch tools directory '${tools_dir}/${arch}' does not exist."
+    arch_dir="$rootfs_arch"
```
cbrune: looks like a whitespace problem here

patches/busybox/dhcp-hostname-sanitization-optional.patch
```diff
@@ -1,54 +0,0 @@
-back port udhcpc: make hostname sanitization optional patch.
```
cbrune: is this patch now part of the upstream busybox 1.25.1?

patches/busybox/flashcp-progress-100.patch
```diff
@@ -1,23 +0,0 @@
-Fix issue that flashcp progress cannot reach 100%
```
cbrune: is this upstream now?

patches/crosstool-NG/complibs-ppl-Fix-build-of-ppl-0.11.2-with-gcc-4.9.patch
```diff
@@ -1,117 +0,0 @@
-From 6ab70a6fb2a760dfd55c88ae61283b761311d754 Mon Sep 17 00:00:00 2001
```
cbrune: is this patch now upstream in crosstool-NG?

build-config/make/efivar.make
```diff
@@ -82,7 +83,11 @@ $(EFIVAR_INSTALL_STAMP): $(SYSROOT_INIT_STAMP) $(EFIVAR_BUILD_STAMP)
 		cp -av $(DEV_SYSROOT)/usr/bin/$$file $(SYSROOTDIR)/usr/bin/ ; \
 	     done
 	$(Q) for file in $(EFIVAR_LIBS); do \
-		cp -av $(DEV_SYSROOT)/usr/lib/$$file $(SYSROOTDIR)/usr/lib/ ; \
+		if [ "$(ARCH)" == "arm64" ] || [ "$(ARCH)" == "x86_64" ]; then \
+		    cp -av $(DEV_SYSROOT)/usr/lib64/$$file $(SYSROOTDIR)/usr/lib/ ; \
+		else \
+		    cp -av $(DEV_SYSROOT)/usr/lib/$$file $(SYSROOTDIR)/usr/lib/ ; \
+		fi \
```
cbrune: I don't quite understand the `if` statement.  previously this file was only for x86_64 using "lib".  this change now makes x86_64 use "lib64".  is that what you meant?

the `else` case would never be taken as only arm64 and x86_64 use efivar.

david56:
I tried to build `accton_as4600_54t` and got error message from busybox:

    networking/lib.a(ipcalc.o): In function `ipcalc_main':
    ipcalc.c:(.text.ipcalc_main+0x278): warning: gethostbyaddr is obsolescent, use getaddrinfo() instead.
    libbb/lib.a(inet_common.o): In function `INET_resolve':
    inet_common.c:(.text.INET_resolve+0x60): warning: gethostbyname is obsolescent, use getnameinfo() instead.
    debianutils/lib.a(mktemp.o): In function `mktemp_main':
    mktemp.c:(.text.mktemp_main+0xbc): warning: the use of `mktemp' is dangerous, better use `mkstemp'
    networking/lib.a(nslookup.o): In function `set_default_dns':
    nslookup.c:(.text.set_default_dns+0x30): undefined reference to `__res_state'
    nslookup.c:(.text.set_default_dns+0x58): undefined reference to `__res_state'
    networking/lib.a(nslookup.o): In function `nslookup_main':
    nslookup.c:(.text.nslookup_main+0x38): undefined reference to `__res_init'
    nslookup.c:(.text.nslookup_main+0x44): undefined reference to `__res_state'
    collect2: error: ld returned 1 exit status
    Note: if build needs additional libraries, put them in CONFIG_EXTRA_LDLIBS.
    Example: CONFIG_EXTRA_LDLIBS="pthread dl tirpc audit pam"
    make[1]: *** [busybox_unstripped] Error 1

cbrune:
Hi @pankajbansal3073 -

I really appreciate your work and contributions.  I really do.

In an effort to get this pull request going I have committed to master the patches that deal with upgrading the package versions.  This PR is so big that it is hard to review and manage with the github web page.

These patches from the PR are now in master:

- d14e438 Added patch for build failing on system with kernel >=4.4 due to efivar
- 9fcd702 Updated Efivars for Arm64 (Aarch64)
- dc0a2f5 Updated LZO for Arm64 (Arch64)
- e42a2a6 Updated popt for Arm64 (Arch64)
- a502f9a Updated lvm2 for Arm64 (Aarch64)
- 8a44ad3 Added Grub support for aarch64 in ONIE
- a69286e add patch to support aarch64 (arm64) support in kexec-tools
- 9b2121f Updated MTD Utils for Arm64 (Aarch64)

With these applied I have successfully rebuilt ONIE for a number of machines including PowerPC, ARMv7 and x86_64.  For armv7 and x86_64 I have installed the resulting ONIE and then successfully installed Cumulus Linux on the machine.  PowerPC still needs to be done.

I am in the process of rebuilding more machines to check.

With these patches applied we can next focus on the core bits here.  These patches still need to be dealt with:
- Added Qemu Virtual Machine for armv8a
- Added support for arm64(aarch64) UEFI
- Added Kernel version 4.1.28 config for Arm64 (Aarch64)
- Update Crosstool-NG from 1.21.0 to support Aarch64 toolchain
- Upgrade busybox from 1.20.2 to 1.25.1  

The tricky ones will be busybox and crosstool-ng.  The others seem straight forward once these two are figured out.  I think we are very close.

@pankajbansal3073 what I would do now is:
- close this PR
- open a new PR containing only these patches

As to the UEFI vs U-Boot support.  From discussions on the mailing list and in private the way to go is:

- make UEFI+grub the default for aarch64
- allow a machine.make to specify U-Boot if necessary

pankajbansal3073:
Closing this PR and Opening new PR containing only these patches

- Added Qemu Virtual Machine for armv8a
- Added support for arm64(aarch64) UEFI
- Added Kernel version 4.1.28 config for Arm64 (Aarch64)
- Update Crosstool-NG from 1.21.0 to support Aarch64 toolchain
- Upgrade busybox from 1.20.2 to 1.25.1

PR #377, opened 2016-05-26
Add support for DNI 3048UP

vincent0083:
Add support for DNI 3048UP
u-boot : 2012.10
kernel : 3.2.69


machine/dni/dni_3048up/busybox/conf/config
```diff
@@ -0,0 +1,6 @@
+CONFIG_SYS_EEPROM=y
+CONFIG_SYS_EEPROM_DEVICE_MTD=y
+CONFIG_SYS_EEPROM_MTD_DEVICE="/dev/mtd-sys-eeprom"
+CONFIG_SYS_EEPROM_OFFSET=0
+CONFIG_SYS_EEPROM_SIZE=256
```
cbrune: Any reason to make it so small? 1024 or 2048 gives you future room to expand.

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
```diff
@@ -0,0 +1,1933 @@
...
++ * Free Software Foundation;  either version 2 of the  License, or (at your
++ * option) any later version.
++ *
++ */
++/dts-v1/;
++/include/ "helix4.dtsi"
++
++/ {
++    model = "agema,ag6248c-poe";
++    compatible = "agema,ag6248c-poe";
```
cbrune: These names should match the ONIE machine name.  It should say:  "dni,dni_3048up".

See the design spec for more on the use of these fields:
https://github.com/opencomputeproject/onie/wiki/Design-Spec-U-Boot-Linux-Kernel

machine/dni/dni_3048up/machine.make
```diff
@@ -0,0 +1,42 @@
...
+# Specify Linux kernel version
+LINUX_VERSION       = 3.2
+LINUX_MINOR_VERSION = 69
+
+# Set the desired uClibc version
+UCLIBC_VERSION = 0.9.33.2
+
+# Set the desired U-Boot version
+UBOOT_VERSION = 2012.10
+
```
cbrune: This file is missing the required definition of SWITCH_ASIC_VENDOR	.  For this ASIC it should be set to "bcm".

See the following for the list of required settings:
https://github.com/opencomputeproject/onie/wiki/Porting-Guide

See these commits for more on SWITCH_ASIC_VENDOR:
4492b046bb4660
0cef1ea8e91

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
```diff
@@ -0,0 +1,1933 @@
+From 8cdecead792e797e8ecaf8f50f18cbec08d56a9e Mon Sep 17 00:00:00 2001
+Subject: [PATCH] Support for Agema AG6248C-POE platform (ARM Cortex-A9 /
+ Helix4) networking platform Baseline files sourced from
+ Broadcom LDK 3.4.10
+
+diff --git a/arch/arm/boot/dts/agema_ag6248c-poe.dts b/arch/arm/boot/dts/agema_ag6248c-poe.dts
```
vidyaravipati: Looks like the patch is copy from AG6248C-POE platform  patch sent by David Hu which seems to be NAND based 
Is DNI 3048UP same as AG6248C-POE or different platforms.
Can you please fix the platform name references to DNI 3048UP

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
```diff
@@ -0,0 +1,1933 @@
...
+new mode 100755
+index b45588d..0365ac6
+--- a/arch/arm/mach-iproc/board_bu.c
++++ b/arch/arm/mach-iproc/board_bu.c
+@@ -51,7 +51,7 @@
+ #endif /* HNDCTF */
+ 
+ #include <linux/amba/pl022.h>
+-
++#if 0
```
vidyaravipati: Can you change to macro  i.e.(!defined (DNI_3048UP) 

Does this block cause any issue on DNI-3048UP board

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
```diff
@@ -0,0 +1,1933 @@
...
++	{"ethaddr",			"00:90:4c:06:a5:72"},
+ #ifdef FOUR_PORT_CONFIG
+ 	{"vlan1hwname",		"et2"},
+ 	{"vlan1ports",		"0 1 2 8*"},
+@@ -100,6 +103,8 @@ static nvram_t fake_nvram[] = {
+ #define CONFIG_NAND_BASE		0x1c000000
+ #if (defined(CONFIG_MACH_NS) || defined(CONFIG_MACH_NSP))
+ #define CONFIG_ENV_OFFSET		0xa0000		/* 30000-b0000 - use last 10000 for env */
++#elif defined(CONFIG_MACH_DNI_3048UP)
++#define CONFIG_ENV_OFFSET		0xf0000		/* f0000-fffff - use 10000 for env */
```
vidyaravipati: According to dni_3048up.dts, env_offset is 0x00100000 which doesn't match with this definition.

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
```diff
@@ -0,0 +1,1933 @@
...
+  * GPIO 4-11 from ChipcommonA gpio pin 0 - 7
+diff --git a/drivers/bcmdrivers/mdio/iproc_mdio.c b/drivers/bcmdrivers/mdio/iproc_mdio.c
+index 5ecb92e..ec6e0bd 100755
+--- a/drivers/bcmdrivers/mdio/iproc_mdio.c
++++ b/drivers/bcmdrivers/mdio/iproc_mdio.c
+@@ -523,6 +523,7 @@ ccb_mii_init(void)
+     }
+     mdio = mdio_devices.mdio;
+ 
++#if 0
```
vidyaravipati: Any specific reason why mdio device is not getting registered.

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
```diff
@@ -0,0 +1,1933 @@
...
++	i2c_smbus_xfer(i2c_adap, I2C_CPLD_ADDR, 0, I2C_SMBUS_WRITE, CPLD_SW_RESET_REG, I2C_SMBUS_BYTE_DATA, &data);
++#endif
++
+ 	return platform_driver_register(&usbh_driver);
+ }
+ 
+diff --git a/drivers/mtd/devices/m25p80.c b/drivers/mtd/devices/m25p80.c
+old mode 100644
+new mode 100755
+index 3d6beb7..2
```
vidyaravipati: Is there any specific problem does these rtc driver changes address.

machine/dni/dni_3048up/onie-rom.conf
```diff
@@ -0,0 +1,19 @@
+# DNI 3048UP onie ROM configuration
+
+description="Delta Networks, DNI-3048UP"
+
+# Top Down the Nand flash looks like:
```
vidyaravipati: Is this really Nand flash or NOR flash

machine/dni/dni_3048up/onie-rom.conf
```diff
@@ -0,0 +1,19 @@
...
+# 1. u-boot image  (include shmoo)
+# 2. u-boot environment
+# 3. onie-uimage
+format=contiguous-up
```
vidyaravipati: Is the format really contiguous-up.
From dts entries , board_eeprom seems to exist between uboot-env and ONIE
Format should be uboot_ubootenv-up and please refer to qemu_armv7a for reference 
which has similar layout
 
https://github.com/opencomputeproject/onie/blob/master/machine/qemu_armv7a/INSTALL
If this

machine/dni/dni_3048up/INSTALL
```diff
@@ -0,0 +1,92 @@
...
+type ``"make MACHINEROOT=../machine/dni/ MACHINE=dni_3048up all"``.
+For example::
+
+  $ cd build-config
+  $ make -j4 MACHINEROOT=../machine/dni/ MACHINE=dni_3048up all
+
+When complete, the ONIE binaries are located in
+``build/images``::
+
+  -rw-r--r-- 3604480 May  7 11:38 onie-dni_3048up-r1.bin
```
vidyaravipati: As per  line 35,  u-boot and ONIE kernel are installed separately, seems to be inline with nor flash layour  dni_3048up.dts where u-boot+env and onie binaries need to be generated separately.

Please fix onie-rom.conf format as uboot_ubootenv-up and fix install instructions to reflect that.

Please check qemu_armv7a INSTALL instructions which has similar layout to this board.
https://github.com/opencomputeproject/onie/blob/master/machine/qemu_armv7a/INSTALL

machine/dni/dni_3048up/INSTALL
```diff
@@ -0,0 +1,92 @@
...
+-----------------------------------------------
+
+Copying the image down using TFTP and flash to the SPI flash::
+
+  =>tftp $loadaddr dni_3048up-r1.u-boot
+  =>sf probe 0;
+  =>sf erase 0 0x100000;
+  =>sf write $loadaddr 0 $filesize;
+  
+  =>tftp $loadaddr dni_3048up-r1.itb
```
vidyaravipati: Please use ONIE generated binaries after fixing onie-rom.conf instead of itb.
Refer to comments at line 20

machine/dni/dni_3048up/INSTALL
```diff
@@ -0,0 +1,92 @@
...
+
+The following directions assume the files are on the root of the TFTP
+server.
+
+Step 2 -- Install ONIE kernel (dni_3048up-r1.u-boot dni_3048up-r1.itb)
+-----------------------------------------------
+
+Copying the image down using TFTP and flash to the SPI flash::
+
+  =>tftp $loadaddr dni_3048up-r1.u-boot
```
vidyaravipati: Refer to comment at line 20

machine/dni/dni_3048up/onie-rom.conf
```diff
@@ -0,0 +1,19 @@
...
+
+uboot_machine=dni_34xx
```
vidyaravipati: Should this be fixed as DNI_3048UP

machine/dni/dni_3048up/onie-rom.conf
```diff
@@ -0,0 +1,19 @@
...
+
+uimage_max_size=$(( 16 * 1024 * 1024 ))
+uboot_max_size=$((   3 * 1024 * 1024 ))
```
vidyaravipati: Please confirm uimage_max_size and uboot_max_size are updated to reflect max sizes as per DTS.

vidyaravipati:
@vincent0083 Looks like the ARM related changes for following patches are implemented in
platform-arm-dni_3048up.patch, I would strongly suggest to move these changes to respective patches.

machine/dni/dni_3048up/machine.make
```diff
@@ -0,0 +1,42 @@
...
+# Translate hardware revision to ONIE hardware revision
+ifeq ($(VENDOR_REV),1)
+  MACHINE_REV = 1
+else
+  $(warning Unknown VENDOR_REV '$(VENDOR_REV)' for MACHINE '$(MACHINE)')
+  $(error Unknown VENDOR_REV)
+endif
+
+UBOOT_MACHINE = dni_34xx_spi
```
vidyaravipati: Cosmetic change: Should this be dni_3048up

machine/dni/dni_3048up/busybox/conf/config
vincent0083: I find sys-eeprom max size in u-boot like that:
int max_size = min(TLV_TOTAL_LEN_MAX,
CONFIG_SYS_EEPROM_MAX_SIZE - sizeof(tlvinfo_header_t));
So, i modify busybox config 'CONFIG_SYS_EEPROM_SIZE' to 2048.

thanks

machine/dni/dni_3048up/machine.make
vincent0083: Updated. Thanks!

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
vincent0083: Updated. Thanks!

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
vincent0083: Updated. Thanks!

machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
vincent0083: Updated. Thanks!

machine/dni/dni_3048up/onie-rom.conf
vincent0083: Updated. Thanks!

machine/dni/dni_3048up/INSTALL
vincent0083: I change the definition 'format' to uboot_ubootenv-up,
but the u-boot+env image range not fit mtd partition.
Because current script is:
===onie-mk-bin.sh===========================================
    pad_file=$(tempfile)
    dd if=/dev/zero of=$pad_file bs=$env_sector_size count=1 > /dev/null 2>&1 || {
        echo "ERROR: Problems with dd for $format image"
        exit 1
    }
    cat $UBOOT_BIN $pad_file > ${output_bin}.uboot+env
=========================================================

should i modify the script like that?
=========================================================
    pad_file=$(tempfile)
    dd if=/dev/zero of=$pad_file bs=$env_sector_size count=1 > /dev/null 2>&1 || {
        echo "ERROR: Problems with dd for $format image"
        exit 1
    }
    dd if=$UBOOT_BIN of=${output_bin}.uboot+env ibs=$uboot_max_size conv=sync > /dev/null 2>&1 || {
        echo "ERROR: Problems with dd for $format image"
        exit 1
    }
    cat $pad_file >> ${output_bin}.uboot+env
=========================================================

cbrune: Looking at the .dts file, the `uboot_ubootenv-up` format looks correct for this platform.  There is no need to modify the script.

The ${output_bin} and ${output_bin}.uboot+env files are only used to install ONIE from the u-boot prompt following the platform's INSTALL instructions.  MTD partitions are not used at the u-boot prompt.

These files are not used during subsequent ONIE updates, i.e. when using ONIE to update ONIE, where the mtd partitions matter.  The onie-self-update image does use MTD partitions, but the raw images are used in that case.

Are you able to install these files following the platform's INSTALL instructions?

cbrune: Also, the onie-self-update command works, right?

machine/dni/dni_3048up/INSTALL
```diff
@@ -46,12 +46,12 @@ Step 2 -- Install ONIE kernel (dni_3048up-r1.u-boot dni_3048up-r1.itb)
 
 Copying the image down using TFTP and flash to the SPI flash::
 
-  =>tftp $loadaddr dni_3048up-r1.u-boot
+  =>tftp $loadaddr onie-dni_3048up-r1.bin.uboot+env
```
vidyaravipati: Install instructions still seem to be out of sync.  
Can we make sure INSTALL document reflects with real example for generating binaries, expected binaries after the build and install instructions

Ex:
   $ cd build-config
   $ make -j4 MACHINEROOT=../machine/dni/ MACHINE=dni_3048up all
 
 When complete, the ONIE binaries are located in
 ``build/images``::
 
   -rw-r--r-- 3604480 May  7 11:38 onie-dni_3048up-r1.bin  ---> Two binaries should have been generated, one for u-boot and one for ONIE.

2)
 Install ONIE kernel (dni_3048up-r1.u-boot dni_3048up-r1.itb) --> Update with correct names

cbrune:
Thanks.  Looks good.  Pushed to master:

- 65bcc5f Add support for DNI 3048UP

PR #380, opened 2016-06-07
Update dropbear.sh

gdisneyleugers:
Updated key size to 2048.


gdisneyleugers:
Updated key-size to 2048 because 1024 is deprecated and is no longer recommended for usage.

cbrune:
Thanks for the additional information.

Please squash the commits into a single commit for good repo hygene.

gdisneyleugers:
Will do

PR #381, opened 2016-06-07
Create urandom.sh

gdisneyleugers:
Seeding urandom since dropbear is used, a well seeded entropy source is needed

### DIFF
--- a/rootconf/default/etc/init.d/urandom.sh
+++ b/rootconf/default/etc/init.d/urandom.sh
@@ -1,0 +1,84 @@
+SAVEDFILE=/var/lib/urandom/random-seed
+
+[ -c /dev/urandom ] || exit 0
+
+PATH=/sbin:/bin
+if ! POOLBYTES=$((
+  ($(cat /proc/sys/kernel/random/poolsize 2>/dev/null) + 7) / 8
+)) ; then
+  POOLBYTES=512
+fi
+. /lib/init/vars.sh
+
+. /lib/lsb/init-functions
+
+do_status () {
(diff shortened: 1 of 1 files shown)


rootconf/default/etc/init.d/urandom.sh
```diff
@@ -0,0 +1,84 @@
+SAVEDFILE=/var/lib/urandom/random-seed
```
cbrune: ONIE systems in general do not have persistent storage.

rootconf/default/etc/init.d/urandom.sh
```diff
@@ -0,0 +1,84 @@
...
+
+PATH=/sbin:/bin
+if ! POOLBYTES=$((
+  ($(cat /proc/sys/kernel/random/poolsize 2>/dev/null) + 7) / 8
+)) ; then
+  POOLBYTES=512
+fi
+. /lib/init/vars.sh
+
+. /lib/lsb/init-functions
```
cbrune: Did you even try running this on an ONIE system?

I get this:
```
ONIE:/ # ls -l /lib/init/vars.sh
ls: /lib/init/vars.sh: No such file or directory
ONIE:/ # ls -l /lib/lsb/init-functions
ls: /lib/lsb/init-functions: No such file or directory
```

It looks like you plucked this file from another distro and put it in here.  If that is the case you should add some comments and attributions to the original authors and copyright holders.

gdisneyleugers:
Will add and fix, opps wrong code.

PR #382, opened 2016-06-11
Update dropbear.sh

gdisneyleugers:
Updated key-size to 2048 because 1024 is deprecated and is no longer recommended for usage.

### DIFF
--- a/rootconf/default/etc/init.d/dropbear.sh
+++ b/rootconf/default/etc/init.d/dropbear.sh
@@ -28,9 +28,9 @@ get_keys() {
     get_keys_arch || {
         # If problems just make new keys in ramdisk
         # genereate rsa key
-        dropbearkey -t rsa -s 1024 -f $RSA_KEY > /dev/null 2>&1
+        dropbearkey -t rsa -s 2048 -f $RSA_KEY > /dev/null 2>&1
         # genereate dss key
-        dropbearkey -t dss -s 1024 -f $DSS_KEY > /dev/null 2>&1
+        dropbearkey -t dss -s 2048 -f $DSS_KEY > /dev/null 2>&1
     }
 }
 


david56:
Hello @gdisneyleugers, @cbrune,

I found that the revised code would be never executed because `get_keys_arch` in `rootconf/u-boot-arch/sysroot-lib-onie/dropbear-arch` and `rootconf/x86_64/sysroot-lib-onie/dropbear-arch` would always return `0`.  `get_kes_arch` just has duplicate code to generate keys so that it has no issue up to now.

I have an additional question: are the present keys stored in machines necessary to be refreshed to the new ones with 2048 lenghs automatically?

cbrune:
@david56 has a good point.

However, stepping back a bit, the entire idea of improving the key strength of dropbear in ONIE does not make sense.

In the first place, there is nothing secure about ssh in ONIE.  By design the root password in ONIE is *empty*.  So key strength does not matter.  ONIE ssh (and telnet) is insecure by design.

https://github.com/opencomputeproject/onie/wiki/Quick-Start-Guide#log-in-to-the-target-switch

I think we should drop this patch as it does not add any value.

david56:
@cbrune,

Agree.  Thank you for your explanation.

paulmenzel:
@cbrune, sorry, just because a certain part of a system is insecure, doesn’t mean you shouldn’t care about the rest. It’s always good to ship with configured best practices, despite the weakest issue still remain. Besides, that the password can always be changed, and then it’d be great, if the next weakest point was fixed.

Second, it’s about the host keys.

Lastly, this commit needs to be rebased. If not, @david56, could you please close this pull request?

david56:
Hello @paulmenzel ,
This PR was created by @gdisneyleugers that I cannot close it.

ehdoyle:
Closing out as this will need to be re-evaluated with the updates introduced in to the master branch.

PR #386, opened 2016-06-23
Fix onie discover stopped while console XOFF is triggered

audihsu-qci:
This patch print logs to files, which was printed to /dev/console directly
    - tee_log_file to /var/log/onie-tee.log
    - _log_console_msg to /var/log/onie-console.log
    - /etc/rcS.d/S30console-logger.sh will tail these logs to /dev/console

Cause:
    While XOFF is trigger with console port somehow, onie discover will stop
searching installer after a while.

Reproduce:
    1. In console terminal, type CTRL+S to send XOFF
    2. telnet/ssh to DUT, and tail -f /var/log/onie.log
    3. The discover loop will stop searching after a while,
       since linux kernel console buffer is full.

### DIFF
--- a/rootconf/default/etc/init.d/console-logger.sh
+++ b/rootconf/default/etc/init.d/console-logger.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+#  Copyright (C) 2016 Audi Hsu <audi.hsu@quantatw.com>
+#
+#  SPDX-License-Identifier:     GPL-2.0
+
+##
+## redirect console_log_file and tee_log_file to /dev/console
+##
+
+. /lib/onie/functions
+
+touch $console_log_file && /usr/bin/tail -f $console_log_file > /dev/console &
+touch $tee_log_file && /usr/bin/tail -f $tee_log_file > /dev/console &
--- a/rootconf/default/lib/onie/functions
+++ b/rootconf/default/lib/onie/functions
@@ -15,7 +15,9 @@ export onie_installer="/var/tmp/installer"
 
 ONIE_RUN_DIR="/var/run/onie"
 
-tee_log_file=/dev/console
+tee_log_file="/var/log/onie-tee.log"
+console_log_file="/var/log/onie-console.log"
+
 if [ "$onie_boot_reason" = "update" -o "$onie_boot_reason" = "embed" ] ; then
     filename_prefix="onie-updater"
     onie_operation="onie-update"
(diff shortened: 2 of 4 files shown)


cbrune:
Interesting. Good find.

rootconf/default/etc/init.d/console-logger.sh
```diff
@@ -0,0 +1,14 @@
...
+#
+#  SPDX-License-Identifier:     GPL-2.0
+
+##
+## redirect console_log_file and tee_log_file to /dev/console
+##
+
+. /lib/onie/functions
+
+touch $console_log_file && /usr/bin/tail -f $console_log_file > /dev/console &
```
cbrune: Looking for this whole patch, it seems we can simplify it a little bit.

Do we really need both `$console_log_file` and `$tee_log_file`?

It seems we only need `$tee_log_file`.

rootconf/default/lib/onie/functions
```diff
@@ -44,33 +46,33 @@ syslog_tag="onie"
...
 _log_msg()
 {
-	printf "$@" | tee $tee_log_file | logger -t $syslog_tag -p ${syslog_onie}.info
+	printf "$@" | tee -a $tee_log_file | logger -t $syslog_tag -p ${syslog_onie}.info
 }
 
 _log_console_msg()
 {
-	printf "$@" | tee /dev/console | logger -t $syslog_tag -p ${syslog_onie}.info
+	printf "$@" | tee -a $console_log_file | logger -t $syslog_tag -p ${syslog_onie}.info
```
cbrune: It seems we could just use `$tee_log_file` here.  Then we do not need `$console_log_file`.

Thinking a bit about names.

The name `tee_log_file`, which I wrote initially, is not a very good name.  It does not tell the programmer what the file is really for.  If we do consolidate on one log file, I think `$console_log_file` is a better name.   I would search and replace `$tee_log_file` with `$console_log_file`.  That name more clearly indicates what the file is for.

rootconf/default/etc/init.d/console-logger.sh
audihsu-qci: Since _log_console_msg and other different log message levels, I shall keep the console_log_file and tee_log_file.
Unless the logging message definition is changed.

cbrune:
Thinking about this a bit more.... Are we sure this is really a problem worth solving?

Why would someone type ctrl-S on the serial console in the first place?  This *might* be a problem for a developer who is working on ONIE, even that would rare.  I don't see how this would ever impact an end user.

My big concern with the  new implementation is that we are creating *two* new log files that will grow without bound.  We already have /var/log/onie.log that captures all the output  Imagine this scenario:

1. a new switch is plugged in a data center on a Friday afternoon
2. the operator has problems configuring DHCP or whatever and goes home for the weekend
3. the switch keeps running, going through the image discovery waterfall
4. the two new files keep growing, along with syslog file /var/log/onie.log
5. all of this is in a ramdisk
6. by Monday morning when the operator comes back the system has run out of memory
7. the operator fixes the DHCP problem, but switch will not install an image because it is out of memory

That sounds bad to me.

I think this is a "won't fix".  What we should do, however, is make a note of the problem.  Instead of a pull request with a fix, I will create an "issue" to track this problem.

cbrune:
I created #387 to track this issue.  I do not think we will fix this problem, however.

audihsu-qci:
The ``XOFF`` may not be sent by human. With console port, the ``XOFF`` is just some signals. That means, if someone plug ethernet port into console port, which is also RJ45, the console may receive the ``XOFF``.

PR #390, opened 2016-07-12
Accton platforms: update u-boot patch

david56:
After enabling CONFIG_API option in the `common_config.h`, some Accton platforms need to update patch to fix conflict problem.

The platforms are including:

- AS5700_96X
- AS5710_54X
- AS6700_32X
- AS6701_32X
- AS6710_32X


cbrune:
This looks like some good clean-up of the include/configs/common_configs.h file.  Thanks for catching this.

Pushed to master:

- f458410 Accton platforms: update u-boot patch

Issue #336, opened 2016-03-02
IPv6 neighbors are not showed after ping ff02::1 successfully


david56:
Hello @vidyaravipati, @cobracmder,

The issue should happen only in the Broadcom armv7a platforms (AS4610_54 and AS4610_30).

When doing neighbor discovery in `/bin/discover`, the IPv6 neighbors can't be showed after pinging ff02::1 successfully.

I created a simple function to reproduce issue:

```
ONIE:/ # neigh_discovery()
> {
>     ping6 -I eth0 -c 2 ff02::1
>     ping -I eth0 -w 2 -c 2 -q 255.255.255.255
>     ip -4 neigh show
>     ip -6 neigh show
> }
ONIE:/ # neigh_discovery
PING ff02::1 (ff02::1): 56 data bytes
64 bytes from fe80::ce37:abff:fe81:bcc6: seq=0 ttl=64 time=0.273 ms
64 bytes from fe80::ce37:abff:fe81:bcc6: seq=1 ttl=64 time=0.190 ms

--- ff02::1 ping statistics ---
2 packets transmitted, 2 packets received, 0% packet loss
round-trip min/avg/max = 0.190/0.231/0.273 ms
PING 255.255.255.255 (255.255.255.255): 56 data bytes

--- 255.255.255.255 ping statistics ---
2 packets transmitted, 0 packets received, 100% packet loss
ONIE:/ #
```

I performed `ip -6 neigh show` manually again, and the neighbor addr was showed.

```
ONIE:/ # ip -6 neigh show
fe80::ce37:abff:fe81:bcc6 dev eth0 lladdr cc:37:ab:81:bc:c6 ref 1 used 0/0/0 probes 1 REACHABLE
ONIE:/ #
```

This addr will be aged out around 30 seconds.  So that ONIE never gathers the neighbors and adds them to the server list to discover images.

The behavior looks strange.   But it works fine in the x86_64 platforms such as AS7716_32X.  Do you have any idea about the issue?

david56:
BTW, the ONIE version is 2015.11.

cbrune:
Are you saying the neighbors eventually show up in the table?

The neigh_discovery() function does have a `sleep 4` (which I hate doing) after the pings and before the `ip neigh show` commands:
```
    # Allow neighbor table to populate -- better way?
    sleep 4
```

Are you saying that 4 seconds is not enough?  But a larger number does work?

I would like to better understand why it takes so long for the neighbor table to be populated....

[...]

PR #392, opened 2016-07-14
Accton platforms: update u-boot patch

david56:
In commit 65bcc5f, some chunks was moved to common patches.  So that the platform's patches would never include them.

- AS4610_54
- AS4610_30

### DIFF
--- a/machine/accton/accton_as4610_30/u-boot/platform-as4610_30.patch
+++ b/machine/accton/accton_as4610_30/u-boot/platform-as4610_30.patch
@@ -2432,36 +2432,10 @@ index 78e85b7..fefca71 100755
  extern void save_shmoo_to_flash(void);
  
 diff --git a/arch/arm/lib/board.c b/arch/arm/lib/board.c
-index 33e7098..747d9bf 100755
+index 466de29..7e042ea 100755
 --- a/arch/arm/lib/board.c
 +++ b/arch/arm/lib/board.c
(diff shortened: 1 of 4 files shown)


vidyaravipati:
Looks good

david56:
@vidyaravipati ,

Yes, both updater and installer work fine on these platforms.  May I know the reason you asked the test item?

BTW, do you have any idea about the problem of IPv6 Neighbors discovery on AS4610?

https://github.com/opencomputeproject/onie/issues/336

We are still tracking the issue.  It seems that caused by the iProc driver.  i.e., present armv7 machines are having the common issue.

vidyaravipati:
@david56 ,  Making sure that there is no other regression w.r.t NOS with changes in  the common and platform specific patches.

I was not aware of IPv6 neighbors discovery problem and will try to look at iproc gmac driver and let you know if there is any update.  Did you get a chance to investigate the iproc gmac driver

david56:
@vidyaravipati ,

I see.  Thank you.

For the IPv6 neighbors discovery issue, we had done several experiments regarding to shell instructions, different neighbors (Windows 2000, Windows XP, and Ubuntu), and different ONIE ARCH (armv7, powerpc, and x86_64).  But not investigated the driver yet.

cbrune:
Pushed to master:

- d32641f Accton platforms: update u-boot patch

PR #394, opened 2016-07-18
UBI related fixes

mrhpearson:
Hi,

I've been working on bringing up ONIE on another of our boards. It's an older design and uses NAND flash with a UBIFS partitioning scheme. I had some issues with the ONIE build and how it had implemented UBI tools and nodes so made some changes to common ONIE build code.

I think the changes are correct - but it's possible they could impact other vendors (centec and nxp look like they may have boards that are using UBI) though I'd expect them to have issues with the latest build too...

The changes made:
1) The make-devices.pl creates some UBI nodes but the major node numbers are wrong for the UBI driver. I think any installer should be running mdev to get these nodes created correctly so I removed them from the script. 
2) I increased the number of mtd nodes created because on this board if I account for our partitioning we make it to 10. 
3) I added some more of the ubi utilites to the MTDBINS so these are available instead of the busybox versions. The busybox version of ubimkvol doesn't support the "-m" option. It seemed to me that as we are building the utilities and they are more functional than the busybox version we should include them.

I hope it's OK bundling these changes together - they seemed all related. Let me know if any concerns or questions

Thanks
Mark

### DIFF
--- a/build-config/make/mtdutils.make
+++ b/build-config/make/mtdutils.make
@@ -24,7 +24,7 @@ MTDUTILS_STAMP		= $(MTDUTILS_SOURCE_STAMP) \
 			  $(MTDUTILS_BUILD_STAMP) \
 			  $(MTDUTILS_INSTALL_STAMP)
 
-MTDBINS = mkfs.jffs2 mkfs.ubifs ubinize ubiformat ubinfo mtdinfo
+MTDBINS = mkfs.jffs2 mkfs.ubifs ubinize ubiformat ubinfo mtdinfo ubiattach ubimkvol ubidetach ubirmvol
 
 PHONY += mtdutils mtdutils-download mtdutils-source mtdutils-build \
 	 mtdutils-install mtdutils-clean mtdutils-download-clean
--- a/build-config/make/mtdutils.make
+++ b/build-config/make/mtdutils.make
@@ -24,7 +24,8 @@ MTDUTILS_STAMP		= $(MTDUTILS_SOURCE_STAMP) \
 			  $(MTDUTILS_BUILD_STAMP) \
 			  $(MTDUTILS_INSTALL_STAMP)
 
-MTDBINS = mkfs.jffs2 mkfs.ubifs ubinize ubiformat ubinfo mtdinfo ubiattach ubimkvol ubidetach ubirmvol
+MTDBINS = mkfs.jffs2 mkfs.ubifs ubinize ubiformat ubinfo mtdinfo 
+UBIBINS = ubiattach ubimkvol ubidetach ubirmvol
 
 PHONY += mtdutils mtdutils-download mtdutils-source mtdutils-build \
 	 mtdutils-install mtdutils-clean mtdutils-download-clean
@@ -80,6 +81,13 @@ $(MTDUTILS_INSTALL_STAMP): $(SYSROOT_INIT_STAMP) $(MTDUTILS_BUILD_STAMP)
                 install
 	$(Q) for file in $(MTDBINS) ; do \
 		cp -av $(DEV_SYSROOT)/usr/sbin/$$file $(SYSROOTDIR)/usr/sbin/ ; \
+	done
+        #if UBI utils from busybox are not installed, use the mtdtools versions
+	$(Q) for file in $(UBIBINS) ; do \
+		if [ ! -f $(SYSROOTDIR)/usr/sbin/$$file ] ; \
+		then \
+			cp -av $(DEV_SYSROOT)/usr/sbin/$$file $(SYSROOTDIR)/usr/sbin/ ; \
+		fi; \
 	done
 	$(Q) touch $@
 
--- a/build-config/scripts/make-devices.pl
+++ b/build-config/scripts/make-devices.pl
@@ -168,16 +168,9 @@ sub devices
     mkdir "${devdir}pts", 0755;
 
     print "NOR flash access: mtd mtdblock\n";
-    foreach my $i (0..7) {
+    foreach my $i (0..12) {
 	makedev("mtdblock${i}", "b", 31, ${i},     ${root}, ${root}, 0660);
 	makedev("mtd${i}",      "c", 90, (2*${i}), ${root}, ${root}, 0660);
-    }
-
-    print "UBI device access: ubi\n";
-    makedev("ubi_ctrl", "c", 10, 63, ${root}, ${root}, 0660);
-    makedev("ubi0"    , "c", 253, 0, ${root}, ${root}, 0660);
-    foreach my $i (0..7) {
-	makedev("ubi0_${i}", "c", 253, ${i}+1, ${root}, ${root}, 0660);
     }
 
     print "SD flash access: mmcblk0 mmcblk0p{1..7}\n";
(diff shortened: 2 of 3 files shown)


build-config/scripts/make-devices.pl
```diff
@@ -168,18 +168,11 @@ sub devices
     mkdir "${devdir}pts", 0755;
 
     print "NOR flash access: mtd mtdblock\n";
-    foreach my $i (0..7) {
+    foreach my $i (0..12) {
```
cbrune: this change seems fine.

build-config/scripts/make-devices.pl
```diff
@@ -168,18 +168,11 @@ sub devices
...
 	makedev("mtd${i}",      "c", 90, (2*${i}), ${root}, ${root}, 0660);
     }
 
-    print "UBI device access: ubi\n";
-    makedev("ubi_ctrl", "c", 10, 63, ${root}, ${root}, 0660);
-    makedev("ubi0"    , "c", 253, 0, ${root}, ${root}, 0660);
-    foreach my $i (0..7) {
-	makedev("ubi0_${i}", "c", 253, ${i}+1, ${root}, ${root}, 0660);
-    }
-
```
cbrune: I think you are generally correct here also.

It would be cool if ONIE platforms that have NAND would do this on the fly for OS installers.  Somewhat discussed here:  http://lists.opencompute.org/pipermail/opencompute-onie/2015-April/000580.html

Something similar to how virtual block devices /dev/vd[a-z] are created:

https://github.com/opencomputeproject/onie/blob/master/rootconf/default/etc/init.d/makedev.sh#L93-112

You mentioned `mdev` -- maybe that is a better way to go.  I am not familiar with it.

cbrune: My problem with taking this out now is that for some legacy PowerPC platforms the above numbers are correct and are some OS installers are relying on it.

mrhpearson: Yeah - I was wondering about that. I figured that legacy PPC platforms will also be broken but that thought may be flawed.
I'll remove (or reinsert) that part and look into a more flexible way of doing it generally. I followed those links above but I'm not sure they will work at build time - I think you need to create the nodes once the switch has booted. Been a while since I looked at the ONIE boot scripts.....time to go refresh my memory

build-config/make/mtdutils.make
```diff
@@ -24,7 +24,7 @@ MTDUTILS_STAMP		= $(MTDUTILS_SOURCE_STAMP) \
 			  $(MTDUTILS_BUILD_STAMP) \
 			  $(MTDUTILS_INSTALL_STAMP)
 
-MTDBINS = mkfs.jffs2 mkfs.ubifs ubinize ubiformat ubinfo mtdinfo
+MTDBINS = mkfs.jffs2 mkfs.ubifs ubinize ubiformat ubinfo mtdinfo ubiattach ubimkvol ubidetach ubirmvol
```
cbrune: Are you sure this plays nice with the mtdutils-install target, later in this file?  It just does a blind copy.
ubimkvol starts as a symlink to /bin/busybox, so I wonder if this will over write busybox?

Also if the mtdutils version of the ubi stuff is better and we need it, then can we remove the ubi tools from busybox all together?

mrhpearson: The busybox install comes after so what I found was if I have UBI enabled in the busybox config then the busybox utilities will still "rule" and be installed in the sysroot by preference (it would overwrite the mtdutils version).
I removed the UBI config from my boards busybox config to allow the mtdutils version to be the only ones. 
I can remove UBI config from the default config if you want - I was assuming that would potentially impact more boards so wouldn't be a good plan :)

build-config/scripts/make-devices.pl
cbrune: Yes, at boot time is when they need to be created.  The link above about the virtual block devices is what you want.  It is not that bad.

mrhpearson: Doh - I was being dense. Agreed that should work. 
I'll put it together and confirm and then update this commit assuming it all works out. Thanks :)

build-config/make/mtdutils.make
cbrune: Hmmm.  So the proposal here is if one wants the full implementation of the tools they should remove the config from busybox locally.  A little goofy.

However, a lot of the old powerpc machines are really resource constrained, so it is best to not increase their image size.

I'm OK with this proposal.

mrhpearson: Actually - my apologies but turns out I was wrong. I was looking at the contents of sysroot in the build directory and there busybox is being used but the final sysroot has only the mtdtools version. I didn't realise until I installed the image on switch.

What I propose is I'll modify the script so only if the binary is not present will it install the mtdtools version. That should mean that if a platform disables the busybox ubi tools they will get the mtdtools version, and otherwise they will just get the default busybox versions. That should maintain backwards compatibility but let folk like us that want the full versions have them.

mrhpearson:
OK - hopefully I've cleaned up a bit the points that had issues.

I've added mdev -s to the init.d makedev.sh script. That correctly creates the ubi_ctrl device node on our switch. I did check if I could leave the manually added ones in made-devices.pl but mdev will not overwrite those unfortunately. 

I added a check in mtdutils.make so the mtdutils ubi binaries are only installed if the busybox ones aren't present. Tested on my switch with busybox config both enabled and disabled. I think this should ensure backwards compatibility for existing switches.

Thanks
Mark

build-config/make/mtdutils.make
```diff
@@ -81,6 +82,13 @@ $(MTDUTILS_INSTALL_STAMP): $(SYSROOT_INIT_STAMP) $(MTDUTILS_BUILD_STAMP)
 	$(Q) for file in $(MTDBINS) ; do \
 		cp -av $(DEV_SYSROOT)/usr/sbin/$$file $(SYSROOTDIR)/usr/sbin/ ; \
 	done
+        #if UBI utils from busybox are not installed, use the mtdtools versions
+	$(Q) for file in $(UBIBINS) ; do \
+		if [ ! -f $(SYSROOTDIR)/usr/sbin/$$file ] ; \
+		then \
+			cp -av $(DEV_SYSROOT)/usr/sbin/$$file $(SYSROOTDIR)/usr/sbin/ ; \
+		fi; \
+	done
```
cbrune: this part looks good.

rootconf/default/etc/init.d/makedev.sh
```diff
@@ -117,3 +117,7 @@ vdevs=$(ls -d /sys/block/vd[a-z] 2&> /dev/null) && {
 }
 
 mkdir -p $ONIE_RUN_DIR
+
+#Create initial device nodes that use dynamic major node numbers. e.g ubi
+mdev -s 
+
```
cbrune: This looks pretty cool.  Thanks.

cbrune:
Looks good, thanks for your help.  Pushed to master:

- f847af8 Run mdev to populate dynamic nodes, including ubi_ctrl
- 8f097b5 Install UBI utilities if busybox versions have not been installed
- f22804e Increase mtd nodes and remove incorrectly created UBI nodes
- b2d5cb5 Adding extra UBI tools to list of binaries

PR #397, opened 2016-07-19
arm-iproc: invalidate d-cache before booting a kernel

cbrune:
U-Boot, which loads an operating system kernel, disables all the
caches and MMU before jumping to a kernel.  The L1 d-cache, even
though disabled, still contains valid entries.  Let's call these
"stale" entries.

In the particular case of a Linux kernel, a problem comes when the
kernel "flushes" then "invalidates" the d-cache.  The "stale" entries
are written out to main memory during the "flush".  Problems soon
develop.

This patch imports a fix-up originally from the Linux kernel, first
added by the iProc platform team. This fix-up "invalidates" the
d-cache without flushing.

The fix-up and comments are added verbatim from the Linux kernel,
modulo a little function entry/exit code.  This problem is best solved
in U-Boot, instead of the Linux kernel.  Every kernel (or any other
operating system) would have to know about this fix and apply it
manually.  Solving the problem here in U-Boot once for all operating
systems.


patches/u-boot/2012.10/platform-broadcom-iproc-helix4.patch
```diff
@@ -13860,6 +13927,13 @@ index 0000000..b96e61a
 +	v7_outer_cache_enable();
 +}
 +#endif
++
++extern void v7_iproc_invalidate_dcache_all(void);
++
++void cpu_cache_initialization(void)
```
cbrune: `cpu_cache_initialization()` is declared as a "weak" symbol in arch/arm/cpu/armv7/cpu.c.  This function is called just before booting into Linux from the `cleanup_before_linux()` function.

cbrune:
Pushed to master:

- 4465465 arm-iproc: invalidate d-cache before booting a kernel

PR #398, opened 2016-07-19
dni_3048up: Fix up usable SDRAM starting address

cbrune:
The iProc U-Boot maps physical SDRAM to begin at 0x6000_0000.

The initial commit for this platform,
65bcc5f5d2027f7fcc929959770a028c46d6352a, defines the first 16MB
(0x0100_0000) as "reserved for custom use" and excludes it from the
available free memory.

  commit 65bcc5f5d2027f7fcc929959770a028c46d6352a
  Author: vincent0083 <Vincent.Liu@agemasystems.com>
  Date:   Thu May 26 11:02:59 2016 +0800

      Add support for DNI 3048UP

None of the PowerPC ONIE U-Boot implementations do this.

This is just a waste of 16MB.

This patch sets the reserved SDRAM size to 0x0.

This also has the nice property of aligning the start of RAM to a
512MB boundary, which simplifies the memory layout for loading kernels.

### DIFF
--- a/machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
+++ b/machine/dni/dni_3048up/kernel/platform-dni_3048up.patch
@@ -2,7 +2,7 @@
 
 diff --git a/arch/arm/boot/dts/dni_3048up.dts b/arch/arm/boot/dts/dni_3048up.dts
 new file mode 100755
-index 0000000..3215058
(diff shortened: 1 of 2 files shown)


cbrune:
I fear this change is will break existing OS deployments.
We will just have to live with the lost 16MB.
Withdrawing this proposal.

PR #399, opened 2016-07-20
accton_as4610: Fix up usable SDRAM starting address

cbrune:
The iProc U-Boot maps physical SDRAM to begin at 0x6000_0000.

The initial commits for the as4610_XX platforms defined the first 16MB
(0x0100_0000) as "reserved for custom use" and excluded it from the
available free memory:

  commit 021ea69a39106006f70080f47cc82404294e18c6
  Author: Vidya Sagar Ravipati <vidya@cumulusnetworks.com>
  Date:   Wed Oct 28 12:15:10 2015 -0700
   
      arm: accton AS4610-54: ONIE support for accton as4610-30 board

  commit 17fbb7adc50e514f7f5b8373dfded362fa34a026
  Author: Vidya Sagar Ravipati <vidya@cumulusnetworks.com>
  Date:   Mon Oct 26 14:51:45 2015 -0700
   
      arm: accton AS4610-54: ONIE support for accton as4610-54 board

None of the PowerPC ONIE U-Boot implementations do this.

This is just a waste of 16MB.

This patch sets the reserved SDRAM size to 0x0.

This also has the nice property of aligning the start of RAM to a
512MB boundary, which simplifies the memory layout for loading
kernels.

### DIFF
--- a/machine/accton/accton_as4610_30/u-boot/platform-as4610_30.patch
+++ b/machine/accton/accton_as4610_30/u-boot/platform-as4610_30.patch
@@ -9746,7 +9746,7 @@ index 0000000..ccccd65
 +#define CONFIG_L2_CACHE_SIZE			0x80000
 +#define CONFIG_PHYS_SDRAM_1				0x60000000
 +#define CONFIG_LOADADDR					0x70000000 /* default destination location for tftp file (tftpboot cmd) */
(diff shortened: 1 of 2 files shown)


cbrune:
I fear this change is will break existing OS deployments.
We will just have to live with the lost 16MB.
Withdrawing this proposal.

PR #401, opened 2016-07-22
DNI3048UP update

vincent0083:
1. Modify command "onie-sysinfo" option "-v" display u-boot dni version.
2. Update dni version
3. Fix u-boot patch conflict issue:
    patches/u-boot/common/feature-sys-eeprom-tlv-common.patch cause import
    machine/dni/dni3048up/u-boot/platform-arm-dni_3048up.patch error


machine/dni/dni_3048up/rootconf/sysroot-bin/onie-sysinfo
```diff
@@ -0,0 +1,271 @@
...
+    print_val $val
+fi
+
+for v in version vendor_id machine machine_rev platform arch config_version partition_type build_date switch_asic ; do
+    eval dump='$'$v
+    if [ "$dump" = "yes" ] || [ "$all" = "yes" ] ; then
+        eval val='$onie'_$v
+        print_val $val
+        if [ "$version" = "yes" ] || [ "$v" = "version" ] ; then
+        	uboot_v="U-Boot-""$(cat /dev/mtd0 | grep '3048UP V' | awk '{print $13}')"
```
cbrune: This is an awkward way of getting the u-boot version into the output of "onie-sysinfo -v".  This will be a nightmare to maintain going forward.

What happens when the core ONIE project updates the onie-sysinfo command?  Who will make sure to update this copy of the onie-sysinfo command?

There must be a better way of doing this.

The version string should come from the `/etc/machine.conf` file and the `onie_version` variable.  See these comments for how the version string is constructed:

https://github.com/opencomputeproject/onie/blob/master/build-config/Makefile#L197-L239

From those instructions you can add an arbitrary string to the onie version by setting `VENDOR_VERSION` in the machine.make file.  In this case you can add the u-boot version to that variable.

That way you do not need to create a copy of `onie-sysinfo`, which will be very difficult to maintain.

vincent0083: We had considered about setting the  variable "VENDOR_VERSION"  in machine.make 
shown as below : 

VENDOR_VERSION = -K_1.0.7-$(UBOOT_VERNOR_VERSION)
UBOOT_VERNOR_VERSION = U_1.0.8
...
"onie-sysinfo -v" will display like that:
ONIE:/ # onie-sysinfo -v
2016.05-K_1.0.7-U_1.0.8

But we have three questions:

1. When we update the u-boot version ( modify UBOOT_VERNOR_VERSION  ), we need to              rebuild ONIE kernel image and update ONIE kernel image.
 Will that be OK ?

If we only modify the u-boot next time. We need to modify
 UBOOT_VERNOR_VERSION =  U_1.0.9.
And should  VENDOR_VERSION = -K_1.0.7-$(UBOOT_VERNOR_VERSION) or
                     VENDOR_VERSION = -K_1.0.8-$(UBOOT_VERNOR_VERSION)   ?


2. if we update the u-boot version to other version with u-boot bin file in u-boot CLI.
The command  "onie-sysinfo -v" (VENDOR_VERSION = -K_1.0.7-$(UBOOT_VERNOR_VERSION) may not show the real u-boot version.
Will that be OK ?
 
3. Or should we merge the ONIE kernel version and u-boot version to the same version ?
    such as VENDOR_VERSION = -DNI_V2.0.0 
    once we modify u-boot or kernel, the VENDOR_VERSION will be updated.

cbrune:
Excellent questions!

My preference is for option 2 you list above -- merge them into a single `VENDOR_VERSION` string.  Update that version string when *anything* changes.

The end user does not really care what u-boot version is, or what the ONIE kernel version is.  What the end user cares about is what you, the hardware vendor, tells them is the correct "version" of ONIE to use.

It is the hardware vendors' responsibility to keep track of the underlying software versions -- ONIE is a system, composed of many different software projects, each of which has a "version".  When you say the ONIE version is "2016-05-DNI_V2.0.0" you are putting a label on all those underlying versions.  When the end customer says I have ONIE version "2016-05-DNI_V2.0.0" installed you know exactly what software they have.

So when *anything* changes, please update the `VENDOR_VERSION` string.

cbrune:
Thanks looks good.

I merge two of the commits together and tidied up the commit message a bit, but no code changes.

Pushed to master:
- 384a254 Update DNI version to V2.0.0
- da6a825 Fix u-boot patch conflict issue

PR #406, opened 2016-08-10
Accton AS4610: correct mass storage for demo OS installation

david56:


cbrune:
Looks good, thanks.  Pushed to master:

- d1baaaa Accton AS4610: correct mass storage for demo OS installation

PR #409, opened 2016-08-11
Alphanetworks Platforms: Add SNC60A0-486F to support ONIE.

willkuo103:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on Broadcom Tomahawk switch ASIC. CPU module is Intel Rangeley C2558.


machine/alphanetworks/alphanetworks_snc60a0_486f/rootconf/sysroot-lib-onie/sysinfo-platform
```diff
@@ -0,0 +1,25 @@
+
+get_serial_num()
```
cbrune: This override of get_serial_num() is not necessary.

The definition here is the same as the CPU architectural default found in rootconf/x86_64/sysroot-lib-onie/sysinfo-arch.

cbrune:
Looks good, except for the minor get_serial_num() comment.

willkuo103:
Thanks, I will create new pull request with the get_serial_num() removing.

PR #411, opened 2016-08-19
add to support Accton new platforms

david56:
- AS4510_52T
- AS_28T_OPS


demo/installer/u-boot-arch/install.sh
```diff
@@ -21,17 +21,23 @@ hw_load() {
     echo "cp.b $img_start \$loadaddr $img_sz"
 }
 
+hw_bootargs() {
+    echo "setenv bootargs quiet console=\$consoledev,\$baudrate"
```
cbrune: why do you need this?

machine/accton/accton_as4510_52t/demo/platform.conf
```diff
@@ -0,0 +1,17 @@
...
+    nandwrite -p /dev/mtd-open demo-${platform}.itb
+    sync ; sync
+}
+
+hw_load() {
+    echo "nand read \$loadaddr open 0x500000"
+}
+
+hw_bootargs() {
+    echo "setenv bootargs quiet console=\$consoledev,\$baudrate mem=\$mem"
```
cbrune: why do you need to set mem= ?

machine/accton/accton_as4510_52t/kernel/platform-as4510_52t.patch
```diff
@@ -0,0 +1,481 @@
...
++/ {
++    model = "accton,as4510_52t";
++    compatible = "accton,as4510_52t";
++
++    aliases {
++        serial0 = &uart0;
++    };
++
++    memory {
++        reg = <0x61000000 0x7f000000>;
```
cbrune: this definition does not match what is in the u-boot patch below

david56: Thank you for your reminder.

The SDRAM base configs were inconsistent between u-boot and kernel that caused kernel boot issue.  It was used `mem` bootargs to fix the issue.  While I made SDRAM base be consistent, the issue is also fixed.  So that I have removed the code related to assign `mem` argument.

machine/accton/accton_as_28t_ops/u-boot/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,2795 @@
...
++#ifdef CONFIG_L2C_AS_RAM
++#define CONFIG_PHYS_SDRAM_0				0x50000000
++#define CONFIG_PHYS_SDRAM_1				0x60000000
++#define CONFIG_LOADADDR					0x70000000 /* default destination location for tftp file (tftpboot cmd) */
++#define CONFIG_PHYS_SDRAM_RSVD_SIZE		0x01000000 /* bytes reserved from CONFIG_PHYS_SDRAM_1 for custom use */
++#else
++#define CONFIG_PHYS_SDRAM_0				0x1b000000
++#define CONFIG_PHYS_SDRAM_1				0x60000000
++#define CONFIG_LOADADDR					0x70000000 /* default destination location for tftp file (tftpboot cmd) */
++#define CONFIG_PHYS_SDRAM_RSVD_SIZE		0x01000000 /* bytes reserved from CONFIG_PHYS_SDRAM_1 for custom use */
```
cbrune: Why are you reserving 16MB of space?  This is just wasted space.

machine/accton/accton_as4510_52t/kernel/platform-as4510_52t.patch
cbrune: I think you fixed this the wrong way round -- I would have corrected the .dts file.  I would not artificially reserve 0x0100_0000 in the u-boot configuration.  That is just wasting 16MB of SDRAM.

david56: In fact, I had tried to correct the .dts file yesterday.  It hanged at starting kernel so that I turned to reserve 0x0100_0000 like the way in AS4610_54 and AS4610_30.

       Verifying Hash Integrity ... crc32+ OK
       Booting using the fdt blob at 0x704bb094
       Uncompressing Kernel Image ... OK
    boot_prep_linux commandline: quiet console=ttyS0,115200 mem=2032M serial_num=EC1612001765 eth_addr=CC:37:AB:C2:C8:80 boot_re on=rescue
       Loading Device Tree to 63ffb000, end 63fff017 ... OK
    
    Starting kernel ...

I try revising `CONFIG_PHYS_SDRAM_RSVD_SIZE` in AS4610_54 to `0x0` and get the same issue.  At the situation, I specify `mem=2032M` to kernel bootargs that can boot to linux shell.  If the specified size larger than 2032M, issue would still occur.

cbrune: I think the AS4610_54 was done "wrong" -- it should have never reserved any space.  It might be too late to fix those old platforms, but for new platforms we should not make the same mistake.

I think your problem may be the kernel load address.   Check how your uImage is built and see if it is being loaded to `0x6000_8000` or to `0x6100_8000`.  When `CONFIG_PHYS_SDRAM_RSVD_SIZE` is set to `0x0` I think the load address needs to change to `0x6000_8000`.  Then you won't need the `mem=` parameter.

david56:
Thank you for your guidance.  I have revised the commits as you mentioned.  Besides, an additional commit is added for making build system remove machine self-installed u-boot image only.

build-config/arch/armv7a.make
```diff
@@ -13,6 +14,8 @@ CROSSPREFIX ?= $(TARGET)-
 CROSSBIN    ?= $(XTOOLS_INSTALL_DIR)/$(TARGET)/bin
 
 KERNEL_ARCH		= arm
+KERNEL_LOAD_ADDRESS	?= 0x61008000
+KERNEL_ENTRY_POINT	?= 0x61008000
```
cbrune: We are almost there.  I would make the default here be 0x6000_8000.  That is value we want people to use by default.

The handful of legacy platforms (3 machines I think) would need to set these to 0x6100_8000 in their machine.make:
```
accton/accton_as4610_30/machine.make:ONIE_ARCH ?= armv7a
accton/accton_as4610_54/machine.make:ONIE_ARCH ?= armv7a
dni/dni_3048up/machine.make:ONIE_ARCH ?= armv7a
```

Then all new machines going forward do not have to do anything special.

The way it is written here, hardware vendors have to go out of their way to "do the right thing".  I want to the "right thing" to be the default.

cbrune:
This looks *really* good.  Thanks for your work.

Just the small matter of having the default be 0x6000_8000.

david56:
The default kernel load address and entry point have been revised to 0x6000_8000.  Accton and DNI armv7a platforms also keep at 0x6100_8000.  Please help to review again.  Thank you.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,479 @@
...
+ };
+diff --git a/drivers/bcmdrivers/gmac/src/et/sys/etcgmac.c b/drivers/bcmdrivers/gmac/src/et/sys/etcgmac.c
+index cf16543..071c991 100755
+--- a/drivers/bcmdrivers/gmac/src/et/sys/etcgmac.c
++++ b/drivers/bcmdrivers/gmac/src/et/sys/etcgmac.c
+@@ -428,6 +428,7 @@ chipattach(etc_info_t *etc, void *osh, void *regsva)
+ 	} else {
+ 		etc->phyaddr = bcm_atoi(var) & EPHY_MASK;
+ 	}
++	etc->phyaddr = 0x18;
```
vidyaravipati: Is this change specific to accton 4510_52

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,479 @@
...
+ 	printf("et%d: chipattach: phyaddr(0x%x)\n", etc->unit, etc->phyaddr);
+ 
+ 	/* nvram says no phy is present */
+@@ -2273,7 +2274,7 @@ chipphyrd(ch_t *ch, uint phyaddr, uint reg)
+ 	uint32 addr, bank;
+ 	uint16 tmp16;
+ 
+-	addr = phyaddr&0xf;
++	addr = phyaddr&0x1f;
```
vidyaravipati: Can you please elaborate more on root cause for this change

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,479 @@
...
+ #endif
+diff --git a/drivers/mtd/devices/m25p80.c b/drivers/mtd/devices/m25p80.c
+index 3d6beb7..f2d9f46 100644
+--- a/drivers/mtd/devices/m25p80.c
++++ b/drivers/mtd/devices/m25p80.c
+@@ -47,6 +47,15 @@
+ #define	OPCODE_CHIP_ERASE	0xc7	/* Erase whole flash chip */
+ #define	OPCODE_SE		0xd8	/* Sector erase (usually 64KiB) */
+ #define	OPCODE_RDID		0x9f	/* Read JEDEC ID */
++#define	OPCODE_RDFSR		0x70	/* Read Flag Status Register */
```
vidyaravipati: Changes for m25p80.c looks like adding more functionality for this driver. 
Can  this patch is maintained as a separate kernel patch and iproc related changes to support this board as separate patches.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,479 @@
...
++        .size = 64 * 1024,
+     },
+     {
+-        .name = "rootfs",
++        .name = "onie",
++        .offset = MTDPART_OFS_APPEND,
++        .size = 7 * 1024 * 1024,
++    },
++    {
++        .name = "hw-info",
```
vidyaravipati: Is this partition related to board_eeprom. If so, can we name it as board_eeprom for consistency across 4610 and 4510 platforms.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
david56: Yes.  It is specific to the machine.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
david56: The PHY of mgmt port locates at 0x18.  To extract the correct address, the bit mask has to be revised 0x1f.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
david56: The changes should be specific to the machine.  I am not sure whether it will affect to other platforms or not.  So that I just make them keeping in the platforms' patch.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
david56: Sure.  I'll revise it to board_eeprom.

machine/accton/accton_as_28t_ops/kernel/platform-as_28t_ops.patch
vidyaravipati: From the changes, m25p80 driver seems to be enhanced to support couple more opcodes.
It  would be better to make these changes as a separate patch and leave under the machine folder to avoid impacting other platforms (if there is a doubt).

david56: I see.  I'll change it.

machine/accton/accton_as_28t_ops/u-boot/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,2795 @@
...
++			goto no_fdt;
+ 		}
+ 
+ 		printf("   Booting using the fdt blob at 0x%p\n", fdt_blob);
+@@ -1678,8 +1682,8 @@ int boot_get_fdt(int flag, int argc, char * const argv[],
+ 			printf("   Booting using the fdt at 0x%p\n", fdt_blob);
+ 
+ 			if (fdt_check_header(fdt_blob) != 0) {
+-				fdt_error("image is not a fdt");
+-				goto error;
```
vidyaravipati: Looks like changes to support no_fdt seems to be common across all iproc platforms. 
Making this patch as common iproc patch will be good future cleanup

machine/accton/accton_as_28t_ops/u-boot/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,2795 @@
...
++ */
++
++#include <common.h>
++#include <asm/io.h>
++
++#ifndef CONFIG_SYS_DCACHE_OFF
++void enable_caches(void)
++{
++	/* Enable D-cache. I-cache is already enabled in start.S */
++	dcache_enable();
```
vidyaravipati: Any issues observed without this change. 
Not observing any of these changes in bcm95615x.c in iproc uboot, so would be good to understand.

david56: We decide to remove the code by internal discussion.

david56:
@vidyaravipati , I have made changes mentioned above.

machine/accton/accton_as4510_52t/busybox/patches/platform-as4510_52t.patch
```diff
@@ -0,0 +1,38 @@
...
++++ b/libbb/xfuncs_printf.c
+@@ -515,7 +515,13 @@ int FAST_FUNC bb_xioctl(int fd, unsigned request, void *argp, const char *ioctl_
+ 
+ 	ret = ioctl(fd, request, argp);
+ 	if (ret < 0)
+-		bb_simple_perror_msg_and_die(ioctl_name);
++	{
++#include <mtd/mtd-user.h>
++		if (request == MEMERASE)
++			bb_simple_perror_msg(ioctl_name);
```
cbrune: Why is it OK for MEMERASE to fail?  Seems strange to treat that request different from the rest.

machine/accton/accton_as_28t_ops/busybox/patches/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,38 @@
...
+ 
++#ifndef DEBUG
++    FILE *fp;
++    fp = fopen(SYS_EEPROM_MTD_DEVICE, "r");
++    fgetc(fp);
++    fclose(fp);
++#endif
++
+     return ret;
+ }
```
cbrune: Two things:

1. is this change really necessary?  or is it left over debug / bring-up code?

2. if it is really necessary it looks identical to the patch in the  AS4510_52T.  If that is the case suggest moving it to a common accton busybox patch.

machine/accton/accton_as4510_52t/u-boot/platform-as4510_52t.patch
```diff
@@ -0,0 +1,2918 @@
...
+diff --git a/common/cmd_mtdparts.c b/common/cmd_mtdparts.c
+index 2268829..d3abe91 100644
+--- a/common/cmd_mtdparts.c
++++ b/common/cmd_mtdparts.c
+@@ -174,7 +174,7 @@ static int device_del(struct mtd_device *dev);
+  * @param retptr output pointer to next char after parse completes (output)
+  * @return resulting unsigned int
+  */
+-static unsigned long memsize_parse (const char *const ptr, const char **retptr)
++static u64 memsize_parse (const char *const ptr, const char **retptr)
```
cbrune: This change for mtd size moving from u32 to u64 looks like a generally good thing.  It does not look machine specific.

I am guessing this is needed to handle NAND parts that are larger than 4GB.

I think it would be OK to promote this change as a general u-boot patch, i.e. put it in patches/u-boot/2012.10/.

machine/accton/accton_as_28t_ops/u-boot/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,2665 @@
...
++#include <spi_flash.h>
++#include <environment.h>
+ #include <linux/ctype.h>
+ 
+ #include "sys_eeprom.h"
+@@ -926,6 +928,82 @@ int write_sys_eeprom(void *eeprom, int len)
+ 
+ #endif /* CONFIG_SYS_FLASH_HWINFO_ADDR && CONFIG_SYS_FLASH_HWINFO_SECT_SIZE */
+ 
++#ifdef CONFIG_SYS_EEPROM_USE_COMMON_SPI_IO
```
cbrune: This SPI IO patch looks generally like a good thing.  It appears to be the same as the AS4510_52T code.

Can this be factored out into the common sys_eeprom patch?  Can we make it part of patches/u-boot/common/feature-sys-eeprom-tlv-common.patch , like `CONFIG_SYS_EEPROM_USE_COMMON_FLASH_IO` and `CONFIG_SYS_EEPROM_USE_COMMON_I2C_IO` ?

machine/accton/accton_as4510_52t/busybox/patches/platform-as4510_52t.patch
david56: I added the code to make behavior between kernel and u-boot be consistent.

machine/accton/accton_as4510_52t/u-boot/platform-as4510_52t.patch
david56: Yes, you are right.  I'll do it.

machine/accton/accton_as_28t_ops/u-boot/platform-as_28t_ops.patch
david56: OK.  I'll do it.

machine/accton/accton_as4510_52t/busybox/patches/platform-as4510_52t.patch
david56: Should I also move this chunk to be a common patch?

david56:
The u-boot features discussed above are moved into the global/common u-boot patch files.

machine/accton/accton_as_28t_ops/u-boot/platform-as_28t_ops.patch
```diff
@@ -0,0 +1,2573 @@
...
++#include <spi_flash.h>
+ 
+ DECLARE_GLOBAL_DATA_PTR;
+ 
+@@ -328,10 +329,35 @@ int env_complete(char *var, int maxv, char *cmdv[], int bufsz, char *buf)
+ */
+ void env_write_default_if_crc_bad(void)
+ {
+-	env_t *env_ptr = (env_t *)CONFIG_ENV_ADDR;
++	int ret;
```
cbrune: I think we can move this SPI env code to the common patch also.

Just protect the code with:
```
#ifdef CONFIG_ENV_IS_IN_SPI_FLASH
/* new code here */
#else
/* old code here */
#endif

cbrune:
@david56 -- WOW!  You have done a really great job here.

One small refactor on the env CRC code for SPI and then we are done.

I *really* appreciate all the work you have done here.

david56:
@cbrune , The patchset has been updated.  Please review it again.  Thank you.

cbrune:
This is fantastic.  Great work here.  Pushed to master:

- dfc8611 u-boot.make: remove machine self-installed u-boot image only
- 03e2fb6 Accton AS_28T_OPS: add to support ONIE
- 95be455 Accton AS4510_52T: add to support ONIE
- 74b6a9f Quanta platforms: refresh u-boot patch to use common features
- 100648d DNI platforms: refresh u-boot patch to use common features
- cb2c0ef Centec platforms: refresh u-boot patch to use common features
- f986478 AlphaNetworks platforms: refresh u-boot patch to use common features
- a00379a Accton platforms: refresh u-boot patch to use common features
- 9618a04 u-boot: add features separated from AS4510_52T and AS4610_54
- 731b454 u-boot-2012.10: remove execute permission from patch files
- 69bc485 Accton armv7a platforms: specify kernel load address and entry point
- 83a45c8 DNI armv7a platforms: specify kernel load address and entry point
- a4610ed onie-mk-itb: allow to specify kernel load address and entry point

PR #412, opened 2016-08-19
Accton AS7716_32X: configure for mgmt port LED

david56:


cbrune:

- 68ae83b Accton AS7716_32X: configure for mgmt port LED

PR #414, opened 2016-08-22
Alphanetworks Platforms: Add SNH61A0-320F to support ONIE.

alvinyang07334:
The machine has 32x100GbE QSFP28 ports on Cavium Xpliant switch ASIC. The CPU module used here is Intel Rangeley C2558.


machine/alphanetworks/alphanetworks_snh61a0_320f/machine.make
```diff
@@ -0,0 +1,47 @@
+# Alpha Networks SNH61A0-320F
+
+ONIE_ARCH ?= x86_64
+# Actually the ASIC vendor is Cavium here. However, there are no matching strings for now. 
+# That is why "unknown" is used for the vendor.
+SWITCH_ASIC_VENDOR = unknown
```
cbrune: Well, let's come up with a good string here and add it to the infrastructure.

The guideline is to use the stock ticker symbol, which in this case would be `cavm`.

The name is so short, however, that using `cavium` would also be fine.  Let's use `cavium`

cbrune:
Generally looks good, thanks.  Let's use `cavium` for the SWITCH_ASIC_VENDOR.

cbrune:
Sorry, accidentally closed this.

alvinyang07334:
Thanks for the reminder. Then I will only modify the SWITCH_ASIC_VENDOR to cavium in my branch and create a new pull request. This pull request will be closed.

PR #416, opened 2016-08-22
X86 set default before uninstall

trapier:
ONIE can take a long time to erase NOS storage during uninstall, and reinstalls
itself to grub only after NOS erasure completes. If power is interrupted to the
system while ONIE is erasing NOS storage, it is likely to come up in grub
rescue, because the NOS grub partition has been wiped. If the symbols for
NOS-grub in the MBR do not match those in ONIE (likely), it may become
necessary to recover via ISO, which may not be available, thereby necessitating
RMA.

This patches addresses the issue by changing the order of the uninstall process
to first installing ONIE to grub then erase NOS partitions.  With this order of
operations, even if the system loses power, it should still come up to an
working environment (ONIE) where further action can be taken.


trapier:
closing.   submitted from wrong branch.  will resubmit shortly.

PR #418, opened 2016-08-23
Alphanetworks Platforms: Add SNH61A0-320F.

alvinyang07334:
The machine has 32x100GbE QSFP28 ports on Cavium Xpliant switch ASIC. The CPU module used here is Intel Rangeley C2558. SWITCH_ASIC_VENDOR set to cavium.


cbrune:

- eada96f Alphanetworks Platforms: Add SNH61A0-320F

PR #419, opened 2016-08-31
Accton AS4610: remove NOS from mass storage in uninstall mode

david56:


rootconf/default/lib/onie/onie-uninstaller-common
```diff
@@ -0,0 +1,174 @@
+# common uninstall routine
+
+#  Copyright (C) 2014,2015 Curt Brune <curt@cumulusnetworks.com>
+#  Copyright (C) 2016 david_yang <david_yang@accton.com>
+#
+#  SPDX-License-Identifier:     GPL-2.0
+
+# for backward compatibility if running partition_type is empty assume it
+# is msdos.
+true ${onie_partition_type="msdos"}
```
cbrune: Why do we need this to default to msdos?

Line 21 below checks this and errors out if it is not set correctly.

Otherwise this patch seems fine.

david56: In powerpc platforms, `onie_partition_type` is empty.  And most platforms do not have GPT tools.  So that it should be msdos by default.

cbrune: I see, sounds good.

cbrune:

- 101296e Accton AS4610: remove NOS from mass storage in uninstall mode
- 7ba95f0 onie-uninstaller: move common code to onie-uninstaller-common

PR #422, opened 2016-09-07
kernel-4.1.28: back port severe memory leak fix

cbrune:
Back port fix of bad memory leak introduced in 4.1.28.  This fix comes
from 4.1.29.

For ONIE this would manifest itself as running out of memory when
trying to install ONIE via an .ISO image on a blank system.

The start up code searches for the ONIE disk label using the 'blkid'
inside a while loop for 70 iterations.  Each invocation of 'blkid'
would leak ~30MB inside the kernel, eventually running out of memory
with a kernel panic.


cbrune:
Pushed to master:

- 05c6011 kernel-4.1.28: back port severe memory leak fix

PR #424, opened 2016-09-22
Added 'pkg-config' to the install list for preparing debian build host

jasonguy:
Included the apt package 'pkg-config' to the list of packages installed when preparing the debian host for ONIE builds. Apparently this package is required for building some machines, and was not previously installed as a dependency.


cbrune:
Thanks for finding this.  Pushed to master:

- 84ccc5f Add package 'pkg-config' to DEBIAN_BUILD_HOST_PACKAGES

PR #425, opened 2016-09-22
Ability to build e2fsprogs without btrfs enabled

mrhpearson:
Hi Curt,

I was just checking our target built against the latest ONIE and hit an issue where the image had gotten too big to fit in the space allocated to it. It seemed the reason for the size increase was the adding of the btrfs tools.
Not a problem - I disabled btrfs on our platform as we don't want it, and then found that the e2fsprogs component doesn't get built any more, which we do want (need resize2fs). It gets pulled in by btrfs but doesn't work when btrfs is disabled.

The small change below allows it to be built separately. I don't think it should have any repercussions on anything else - let me know if you think it might cause any side effects that I haven't considered :)

Thanks
Mark


cbrune:
Looks good, thanks.  Pushed to master:

- df9010d Ability to build e2fsprogs without btrfs enabled

PR #429, opened 2016-10-14
Dropbear 2016.74

nshopik:
Update dropbear to latest version.
Fixes CVE-2016-7406 CVE-2016-7407 CVE-2016-7408 CVE-2016-7409
Plus small memory leaks and performance improvments


cbrune:
Generally sounds good to move forward.

rootconf/default/etc/init.d/dropbear.sh
```diff
@@ -14,7 +14,7 @@ daemon=/usr/sbin/$name
 
 [ -x $daemon ] || exit 0
 
-ARGS="-m -B -P"
+ARGS="-m -B"
```
nshopik: `-P` suppose to have valid pid filename previous version seems accept it empty but new one won't start with empty and by default it will use `/var/run/dropbear.pid` if not `-P` is supplied.

nshopik:
I've tested on quanta LB9 didn't notice even increase size of image after rebuild with new dropbear. 

Currently its doesn't generate ECC keys but I we could add that later on as they are smaller and faster

cbrune:
Thanks, pushed to master:

- 5934c1e Dropbear 2016.74

I also uploaded dropbear-2016.74.tar.bz2 to the ONIE mirror.

PR #430, opened 2016-10-21
build: fix lvm2 race condition

cbrune:
The lvm2 build depends on libblkid.so, which comes from the util-linux
build.

The lvm2 build was missing an explicit dependence on the util-linux
build, which sometimes led to a compile time failure for lvm2.  The
problem is more frequent with higher make parallelism.

This problem has been lurking in the build system forever.

This patch adds to the lvm2 build an explicit dependence on the
util-linux.  This insures that util-linux is build first before the
lvm2 build starts.


cbrune:
Applied to master:

- b872ae0 build: fix lvm2 race condition

PR #433, opened 2016-10-25
Accton AS5610_52X u-boot: tune DDR configurations

david56:


cbrune:

- 1c09ea6 Accton AS5610_52X u-boot: tune DDR configurations

PR #434, opened 2016-10-25
Accton platforms: change the order of installation storage

david56:
The platforms are including:
- AS7712_32X
- AS7716_32X

### DIFF
--- a/installer/x86_64/install-arch
+++ b/installer/x86_64/install-arch
@@ -77,6 +77,17 @@ init_onie_install()
     }
     . ./onie-blkdev-common
 
+    # wipe out current onie_dev if ONIE will embed to another
+    # one, or just keep updating to current one
+    curr_onie_dev=$(onie_get_boot_disk)
+    if [ -b "$curr_onie_dev" ] && [ "$curr_onie_dev" != "$onie_dev" ] ; then
+        if [ "$onie_boot_reason" = "embed" ] ; then
+            wipe_curr_onie_dev="yes"
+        else
+            onie_dev="$curr_onie_dev"
+        fi
+    fi
+
     running_firmware="$(onie_get_running_firmware)"
 
     # Determine the desired firmware to install based on the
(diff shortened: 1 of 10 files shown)


installer/x86_64/install-arch
```diff
@@ -77,6 +77,17 @@ init_onie_install()
...
+    # one, or just keep updating to current one
+    curr_onie_dev=$(onie_get_boot_disk)
+    if [ -b "$curr_onie_dev" ] && [ "$curr_onie_dev" != "$onie_dev" ] ; then
+        if [ "$onie_boot_reason" = "embed" ] ; then
+            wipe_curr_onie_dev="yes"
+        else
+            onie_dev="$curr_onie_dev"
+        fi
+    fi
+
```
cbrune: Can you provide a bit more background?  It is not a lot of code, but it is a big change to way things work.

I think I understand, but can you please describe the intended use for this on the mailing list.  Not asking for a lot, but a short, high level description of the "requirements" and "expected use cases".

installer/x86_64/install-arch
```diff
@@ -57,7 +57,7 @@ init_onie_install()
     # Find target install block device.
     # - Default value from burned in install.conf
     # - Overridden with -d CLI arg
-    onie_dev="$(install_device_platform)"
+    [ -b "$onie_dev" ] || onie_dev="$(install_device_platform)"
```
cbrune: I don't understand this one.  Where would `$onie_dev` be set previous to this?

david56: This should be an incomplete feature descripted above:

    # Find target install block device.
    # - Default value from burned in install.conf
    # - Overridden with -d CLI arg

https://github.com/opencomputeproject/onie/blob/master/installer/x86_64/install-arch#L26-L43 catches `$onie_dev` passed via `-d` argument.

rootconf/default/bin/onie-self-update
```diff
@@ -79,8 +80,12 @@ cli_static_update_url=$1
     exit 1
 }
 
+shift 1
+cli_static_update_parms="$@"
+
 onie_disco=
 onie_disco=$(add_parm "$onie_disco" onie_cli_static_update_url $cli_static_update_url)
+onie_disco=$(add_parm "$onie_disco" onie_cli_static_update_parms "$cli_static_update_parms")
```
david56: I have updated the commit:

7d5352a allow onie-nos-install and onie-self-update to pass parameters to installer and updater

installer/x86_64/install-arch
cbrune: Ah, OK.

david56:
@cbrune , is there any further comment on the PR?  Thanks.

cbrune:
Sorry, this has fallen off my stack.  Will look at it once I get the aarch64 patches squared away.

david56:
I see.  Thank you.

cbrune:
Thanks, pushed to master:

- d21fcfc Accton platforms: change the order of installation storage
- dac3088 allow onie-nos-install and onie-self-update to pass parameters to installer and updater
- 47cbea8 x86_64 install-arch: correct function of specifying block device
- a705007 x86_64 install-arch: support to embed to another mass storage

PR #436, opened 2016-10-31
Add additional ONIE information into onie-support

pichuang:
The additional support information Include
- onie-syseeprom
- onie-sysinfo
- lsb-release
- blkid
- sgdisk

These informations will help switch hardware vendor to debug and fix.

### DIFF
--- a/rootconf/default/bin/onie-support
+++ b/rootconf/default/bin/onie-support
@@ -30,6 +30,7 @@ support_platform()
 
 [ -r /lib/onie/support-arch ]     && . /lib/onie/support-arch
 [ -r /lib/onie/support-platform ] && . /lib/onie/support-platform
+[ -r /lib/onie/functions ] && . /lib/onie/functions
 
 cleanup() {
     umount $tmpdir > /dev/null 2>&1
@@ -43,7 +44,8 @@ mount -t tmpfs tmpfs $tmpdir || {
     cleanup 1
 }
 
-tarfile="$output_dir/onie-support.tar.bz2"
+onie_support_name="onie-support-${onie_machine}.tar.bz2"
+tarfile="$output_dir/$onie_support_name"
 
 tar_dir="onie-support"
 save_dir="$tmpdir/$tar_dir"
--- a/rootconf/default/bin/onie-support
+++ b/rootconf/default/bin/onie-support
@@ -58,6 +58,11 @@ export > $save_dir/runtime-export-env.txt
 set > $save_dir/runtime-set-env.txt
 ps w > $save_dir/runtime-process.txt
 dmesg > $save_dir/dmesg.txt
+onie-syseeprom > $save_dir/onie-syseeprom.txt
+onie-sysinfo -a > $save_dir/onie-sysinfo.txt
+cp /etc/machine.conf > $save_dir/machine.txt
+blkid > $save_dir/blkid.txt
+fdisk -l > $save_dir/fdisk.txt
 support_arch
 support_platform
 tar -C $tmpdir -cjf $tarfile $tar_dir || {


rootconf/default/bin/onie-support
```diff
@@ -58,6 +58,11 @@ export > $save_dir/runtime-export-env.txt
 set > $save_dir/runtime-set-env.txt
 ps w > $save_dir/runtime-process.txt
 dmesg > $save_dir/dmesg.txt
+onie-syseeprom > $save_dir/onie-syseeprom.txt
+onie-sysinfo > $save_dir/onie-sysinfo.txt
+cp /etc/lsb-release > $save_dir/lsb-release.txt
+blkid > $save_dir/blkid.txt
+sgdisk -p /dev/sda > $save_dir/sgdisk.txt
```
cobracmder: This line will only work for ONIE devices that have /dev/sda.  We need to make this line applicable to all ONIE machines.  Take a look at https://github.com/opencomputeproject/onie/blob/master/contrib/oce/test-nos.sh#L99 .  We'll have to do something similar here to get all machines.

cobracmder:
Hi @pichuang,

Thanks for the PR.  Take a look at my comment on the commit about targeting all ONIE machines to print out their disk partitioning layout.

rootconf/default/bin/onie-support
```diff
@@ -58,6 +58,11 @@ export > $save_dir/runtime-export-env.txt
 set > $save_dir/runtime-set-env.txt
 ps w > $save_dir/runtime-process.txt
 dmesg > $save_dir/dmesg.txt
+onie-syseeprom > $save_dir/onie-syseeprom.txt
+onie-sysinfo > $save_dir/onie-sysinfo.txt
```
cbrune: It would be better to use the `-a` option to `onie-sysinfo` as that dumps all the info.

rootconf/default/bin/onie-support
```diff
@@ -58,6 +58,11 @@ export > $save_dir/runtime-export-env.txt
...
+cp /etc/lsb-release > $save_dir/lsb-release.txt
```
cbrune: since we're copying stuff, it would be good to copy `/etc/machine.conf` to the support area.

cbrune:
Generally this is a really great idea as long as we do not go overboard.

Keep in mind the support output is built from 3 pieces:

- a common piece, which is the file you are editing here
- a CPU architecture specific piece, `rootconf/x86_64/sysroot-lib-onie/support-arch` and `rootconf/u-boot-arch/sysroot-lib-onie/support-arch`
- a platform specific piece

Only the `x86_64` CPU arch uses `sgdisk`.  Also like @cobracmder says, the disk is not always /dev/sda.  Some x86_64 machines use /dev/sdb or even /dev/mmcblk0.

I would put the sgdisk piece in `rootconf/x86_64/sysroot-lib-onie/support-arch`.  If you limit that to x86 then you can source `rootconf/x86_64/sysroot-lib-onie/onie-blkdev-common` and call `onie_get_boot_dev` to get the disk device, something like:
```
. rootconf/x86_64/sysroot-lib-onie/onie-blkdev-common
sgdisk -p $(onie_get_boot_dev)
```

As an alternative you could rummage around `/sys/block/*` and run sgdisk on the block devices you find there.  I would like the dump to block devices that match these patterns:
- `[sv]d[a-z]`
- `mmcblk[0-9]`

rootconf/default/bin/onie-support
pichuang: Thanks. Fixed.

rootconf/default/bin/onie-support
pichuang: OK. I change the copying target from `/etc/lsb-release` to `/etc/machine.conf`

pichuang:
Hi @cbrune @cobracmder  Could you please help me to review code? I have change other way to dump all of disk information, not specific partition infomation. It is work on x86 and u-boot. Thanks.

cbrune:
Looks good.  Thanks for the contribution.

Pushed to master:

- 7118374 Change onie-support default filename
- 18d46f7 Add additional ONIE information into onie-support

PR #437, opened 2016-11-03
Accton AS7412_32X: add to support ONIE

david56:
The machine has 32x100GbE QSFP28 ports on Mellanox switch ASIC.
CPU module is Intel Rangeley C2538.


machine/accton/accton_as7412_32x/busybox/conf/config
```diff
@@ -0,0 +1,7 @@
+CONFIG_SYS_EEPROM=y
+CONFIG_SYS_EEPROM_DEVICE_I2C=y
+CONFIG_SYS_EEPROM_I2C_DEVICE="/dev/i2c-1"
+CONFIG_SYS_EEPROM_I2C_ADDR=0x57
+CONFIG_SYS_EEPROM_OFFSET=0
+CONFIG_SYS_EEPROM_SIZE=256
+CONFIG_SYS_EEPROM_MAX_SIZE=256
```
cbrune: I see all 3 of these new switches have 256 for the EEPROM size.  That is pretty small.  Is that correct?

machine/accton/accton_as7412_32x/installer.conf
```diff
@@ -0,0 +1,31 @@
...
+        fi
+    done
+    return 1
+}
+
+# Default ONIE block device
+install_device_platform()
+{
+    discover_mass_storage "1-1.4:1.0" ||
+    discover_mass_storage "target4:0:0"
```
cbrune: The order here is opposite that in the patches for AS7712_16X and AS5822_54X.  Is that correct?  I figured they would all 3 be the same.

machine/accton/accton_as7412_32x/busybox/conf/config
david56: Yes, the three switches have standalone EEPROMs with 256 bytes capacity.

cbrune:
Thanks for the contribution!

- f8a7b92 Accton AS7412_32X: add to support ONIE

PR #439, opened 2016-11-03
Accton AS7712_16X: add to support ONIE

david56:
The machine has 16x100GbE QSFP28 ports on Broadcom Tomahawk
switch ASIC.  CPU module is Intel Rangeley C2538.


cbrune:
Thanks again, good work.

Pushed to master:
- 48224da Accton AS7712_16X: add to support ONIE

PR #441, opened 2016-11-10
linux-3.10.0: back port upstream kernel build system fix

cbrune:
This patch backports a 3.10.y kernel build system fix:

  From: Nicolas Dichtel <nicolas.dichtel@6wind.com>
  Date: Mon, 29 Apr 2013 14:15:51 +0200
  Subject: [PATCH 1/1] kbuild: fix make headers_install when path is too long

  commit c0ff68f1611d6855a06d672989ad5cfea160a4eb upstream.

  If headers_install is executed from a deep/long directory structure, the
  shell's maximum argument length can be execeeded, which breaks the operation
  with:

  | make[2]: execvp: /bin/sh: Argument list too long
  | make[2]: ***

  Instead of passing each files name with the entire path, I give only the file
  name without the source path and give this path as a new argument to
  headers_install.pl.

  Because there is three possible paths, I have tree input-files list, one per
  path.

  Signed-off-by: Nicolas Dichtel <nicolas.dichtel@6wind.com>
  Tested-by: Bruce Ashfield <bruce.ashfield@windriver.com>
  Signed-off-by: Michal Marek <mmarek@suse.cz>
  Cc: Wang Nan <wangnan0@huawei.com>
  Signed-off-by: Greg Kroah-Hartman <gregkh@linuxfoundation.org>

Without this patch building the Mellanox mlnx_x86 machine will fail if
the build directory has deep nesting.

### DIFF
--- a/patches/crosstool-NG/linux-3-10-0-54-0-1-el7-x86_64
+++ b/patches/crosstool-NG/linux-3-10-0-54-0-1-el7-x86_64
@@ -1,0 +1,145 @@
+linux-3.10.0-54.0.1.el7.x86_64-install-headers-argument-too-long.patch
+
+Backport upstream 3.10.y patch to handle build directories that have a
+deep hierarchy.
+
+This patch backports a 3.10.y kernel build system fix:
+
+  From: Nicolas Dichtel <nicolas.dichtel@6wind.com>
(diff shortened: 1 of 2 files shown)


cbrune:
Applied to master:

- c075631 linux-3.10.0: back port upstream kernel build system fix

PR #442, opened 2016-11-15
Alphanetworks SNC60A0-486F: Follow EEPROM specification to modify the…

willkuo103:
… write cycle time to 5ms.


cbrune:

- b1c0445 Alphanetworks SNC60A0-486F: Follow EEPROM specification to modify the write cycle time to 5ms.

Thanks!

PR #443, opened 2016-11-18
x86_64 demo installer: force to create file system

david56:
When performing `mkfs.ext4` on some mass storage, it will raise a hint
messege:

    ONIE:/ # mkfs.ext4 /dev/sda3
    mke2fs 1.42.13 (17-May-2015)
    /dev/sda3 contains a ext4 file system
            created on Mon Jan  1 00:01:14 2001
    Proceed anyway? (y,n)

The patch addes `-F` to make it proceed automatically.


cbrune:
Looks reasonable -- thanks.  Pushed to master:

- 1bea3e4 x86_64 demo installer: force to create file system

PR #446, opened 2016-11-29
Alphanetworks STX60A0-486F : add to support ONIE

Mingxikao:
The machine has 48x10GbE SFP+ ports and 6x100GbE QSFP28 ports on Broadcom Qumran switch ASIC. CPU module is Intel Rangeley C2558.


cbrune:
Thanks for your contributions!

Pushed to master:
- c4b1262 Alphanetworks STX60A0-486F : add to support ONIE

PR #447, opened 2016-11-29
Alphanetworks STX60B0-486F : add to support ONIE

Mingxikao:
The machine has 48x10GbE SFP+ ports and 6x100GbE QSFP28 ports on Broadcom Qumran switch ASIC. CPU module is Intel Broadwell DE D1518.


cbrune:
Thanks, pushed to master:
- 9db6ce4 Alphanetworks STX60B0-486F : add to support ONIE

PR #450, opened 2016-12-01
Alphanetworks STX60A0-486F : fix the console device number

Mingxikao:
The machine has 48x10GbE SFP+ ports and 6x100GbE QSFP28 ports on Broadcom Qumran switch ASIC. CPU module is Intel Rangeley C2558.
Change the console device from ttyS0 to ttyS1.


cbrune:
pushed to master:
- 1616e33 Alphanetworks STX60A0-486F : fix the console device number

PR #451, opened 2016-12-01
Alphanetworks STX60B0-486F : fix the console device number

Mingxikao:
The machine has 48x10GbE SFP+ ports and 6x100GbE QSFP28 ports on Broadcom Qumran switch ASIC. CPU module is Intel Broadwell DE D1518.
Change the console device from ttyS1 to ttyS0.


cbrune:
pushed to master:
- 217f833 Alphanetworks STX60B0-486F : fix the console device number

PR #453, opened 2016-12-14
Remove the 'local' keyword from global variables

### LINKED ISSUE #452: new busybox busybox-1.25.1 enforces 'local' keyword syntax
Seeing this on the master branch:
```
/etc/rcS.d/S02init-arch.sh: local: line 17: not in a function
/etc/rcS.d/S02init-arch.sh: local: line 18: not in a function
sh: -lt: argument expected
ERROR: Getting ONIE boot device timeout
```

This does not occur on the 2016.11 release.

The main difference is that busybox is now version 1.25.1, previously 1.20.2.
The new version has stricter checking of the 'local' ash keyword.

cbrune:
This bug has been lurking in the code for two years.  Now the current
busybox version 1.25.1 catches the syntax error.

Previously the older busybox 1.20.2 ignored the syntax error.

This patch removes the 'local' keyword from the global variable
definitions.

Closes #452

### DIFF
--- a/rootconf/grub-arch/sysroot-lib-onie/init-arch
+++ b/rootconf/grub-arch/sysroot-lib-onie/init-arch
@@ -14,8 +14,8 @@
 
 mkdir -p $onie_boot_mnt
 
-local delay=70
-local cnt=0
+delay=70
+cnt=0
 while [ $cnt -lt $delay ] ; do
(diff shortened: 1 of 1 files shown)


cbrune:
pushed to master:

- f047d3e Remove the 'local' keyword from global variables

PR #457, opened 2016-12-21
Installer runtime code sharing

cbrune:


cbrune:
Pushed to master:

- f4fc17f kvm_x86_64: test the installer/runtime code sharing
- c29b230 installer: share files with ONIE run-time

PR #459, opened 2016-12-23
Fix Compilation Errors in armv8a and fix issue with NOS image discovery

pankajbansal3073:


build-config/make/images.make
```diff
@@ -196,11 +196,17 @@ endif
 sysroot-check: $(SYSROOT_CHECK_STAMP)
 $(SYSROOT_CHECK_STAMP): $(PACKAGES_INSTALL_STAMPS)
 	$(Q) for file in $(SYSROOT_LIBS) ; do \
-		[ -r "$(DEV_SYSROOT)/lib/$$file" ] || { \
-			echo "ERROR: Missing SYSROOT_LIB: $$file" ; \
-			exit 1; } ; \
-		find $(DEV_SYSROOT)/lib -name $$file | xargs -i \
-			cp -av {} $(SYSROOTDIR)/lib/ || exit 1 ; \
+		if [ "$(ARCH)" == "arm64" ] || [ "$(ARCH)" == "x86_64" ] ; then \
```
cbrune: Why is x86_64 included here?  x86_64 is currently building fine as it is with the new crosstool-ng.

It seems like this is only an arm64 issue.  I'm probably missing something.

build-config/scripts/onie-mk-iso.sh
```diff
@@ -64,10 +64,12 @@ RECOVERY_ISO_IMAGE=${12}
...
-[ -r "${GRUB_TARGET_LIB_I386_DIR}/biosdisk.mod" ] || {
-    echo "ERROR: Does not look like valid GRUB i386-pc library directory: $GRUB_TARGET_LIB_I386_DIR"
-    exit 1
-}
+if [ "$FIRMWARE_TYPE" = "auto" ] || [ "$FIRMWARE_TYPE" = "bios" ] ; then
+    [ -r "${GRUB_TARGET_LIB_I386_DIR}/biosdisk.mod" ] || {
+        echo "ERROR: Does not look like valid GRUB i386-pc library directory: $GRUB_TARGET_LIB_I386_DIR"
+        exit 1
+    }
+fi
```
cbrune: The onie-mk-iso.sh change looks OK to me.  This is orthogonal to the above bit about lib32 / lib64.  Let's make this a separate patch.

rootconf/default/bin/discover
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
```
cbrune: I think this issue needs some more thought.  Also this issue should be in a separate pull request from the lib32/lib64 issues as they are unrelated.

Correct me if I'm wrong, but here is my take on what is happening:

The `sd_dhcp4()` function attempts DHCPv4 on all interfaces, but stops when it gets a reply from one.  Let's say you have two interfaces `eth0` and `eth1`, both of which have a DHCP server (possibility different servers, or different configurations) - lets call the DHCP severs dhcp0 and dhcp1 respectively.

The `sd_dhcp4()` loop will stop after `eth0` gets a DHCP reply from dhcp0.  `eth1` never gets a chance to talk with dhcp1.

In your case the desired DHCP configuration (the tftp server info) is only valid from dhcp1 -- for example the reply from dhcp0 does not contain the tftp server info.

It seems like the ONIE discovery needs to be per interface and the image waterfall driven all the way through before moving on to the next interface.  Today the code is not quite right, where it first tries to gather all the DHCP info from all interfaces and then drive the waterfall.

The reason the code stops after the first interface has a valid reply is that what if both dhcp0 and dhcp1 specify a tftp server?  Who wins?

All this code needs a bit of a rework and you have done most that here -- thanks!

A few comments:

1. `sd_static` and `sd_localfs` should only happen once per discovery loop.  These should not happen once per network interface.  I think once these discoveries are complete we could launch `exec_installer` just for them.

2. `sd_dhcp6`, `sd_dhcp4`, `sd_mdns`, `sd_fallback` and `neigh_discovery` happen once per *interface*.  After all of these network discoveries are complete for an interface, launch `exec_installer`.  This would allow each interface to receive DHCP replies from different servers, with potentially different image URL configuration.

It looks like you have done most of this already.  The only change I would make is that the static and local file system discovery only happen once per "discovery loop", instead of once per interface.  Otherwise looks pretty reasonable.

rootconf/default/bin/discover
```diff
@@ -162,30 +158,33 @@ service_discovery()
...
-    service_discovery
-    log_debug_msg "onie_disco: $onie_disco"
-    neigh_discovery
+    intf_list=$(net_intf)
+    for i in $intf_list ; do
+        /etc/init.d/networking.sh "$i" discover
+        /etc/init.d/syslogd.sh discover
+        service_discovery "$i"
+        log_debug_msg "onie_disco: $onie_disco"
+        neigh_discovery
```
cbrune: looks like `neigh_discovery` should be passed `$i`

PR #463, opened 2017-01-11
Lenovo g8272

mrhpearson:
Hi,

This is the support for the Lenovo G8272 platform. 

It should all be pretty standard with the minor exception of the change made to the onie-mk-bin.sh where I added a new format of image. Our uboot env is miles away in flash from uboot and onie (this is a legacy platform and rejigging the flash map would cause significant headaches). This change helps us have sane ONIE images. Let me know if you see any problems

Thanks
Mark


machine/lenovo/lenovo_g8272/kernel/platform-lenovo-g8272.patch
```diff
@@ -0,0 +1,764 @@
...
++                           label ="onie";
++                           reg = <0xba0000 0x3e0000>;
++                       };
++                       u-boot@f80000 {
++                           label = "uboot";
++                           reg = <0xf80000 0x80000>;
++                       };
++                       open@40000 {
++                           label = "open";
++                           reg = <0x40000 0x800000>;
```
cbrune: does it work OK having an overlapping region?

I'm guessing the `linux-kernel` and `rfs` regions are for supporting a legacy OS.  Where the new, overlapping `open` partition allows a new OS to carve up those sections as it sees fit.

mrhpearson: Exactly. If you install, for example. the demoOS it trashes our (legacy) OS kernel/RFS but that's OK because it would be reinstalled if you ran the Lenovo NOS installer again.

cbrune:
Looks, great.  Thanks for the contribution.

Pushed to master:

- d118a32 Lenovo G8272 platform
- 56a204b Add format for platforms where uboot env is in hard to reach location

PR #465, opened 2017-01-13
Fix model description

VicLin66:
Just fix description to right model name with installer.conf.


cbrune:
I believe b965742480c9b5a538b4d14ee15233bbace3871d is already applied to master.

Please make a new pull request, based on the current master branch, that only includes b9eea83.

VicLin66:
OK. Will make a new pull request. Close it.

PR #469, opened 2017-01-26
Accton AS5912_54X: add to support ONIE

david56:
on Broadcom Qumran switch ASIC.  CPU module is Intel Rangeley C2538.


cbrune:
Thanks, pushed to master:

- 5e8cb89 Accton AS5912_54X: add to support ONIE

PR #470, opened 2017-01-26
Accton AS7512_32X: change the order of installation storage

david56:


cbrune:
Pushed to master:

- f3eb5d7 Accton AS7512_32X: change the order of installation storage

PR #471, opened 2017-01-26
Accton AS7312_54X: enable busybox swap instructions

david56:


machine/accton/accton_as7312_54x/busybox/conf/config
```diff
@@ -6,3 +6,7 @@ CONFIG_SYS_EEPROM_I2C_MEM_ADDR_BITS=8
 CONFIG_SYS_EEPROM_OFFSET=0
 CONFIG_SYS_EEPROM_SIZE=256
 CONFIG_SYS_EEPROM_MAX_SIZE=256
+CONFIG_MKSWAP=y
+CONFIG_FEATURE_MKSWAP_UUID=y
+CONFIG_SWAPONOFF=y
+CONFIG_FEATURE_SWAPON_PRI=y
```
cbrune: Is this something we should make available to every platform?

Seems a little strange to add it to just one platform.

david56: @cbrune, Happy Lunar New Year!
I'm not sure whether most NOS installers need these commands or not.  Currently, we have only two machines for the same customer that needs the commands to create swap partition.  I think it can to be available to every platform, too.

cbrune:
I think adding this globally will be OK.  See PR #484.

My only real concern is the size of the image will increase a little.

I tried compiling the accton_as5610_52x, which has tight space concerns, and that was fine.

david56:
Thank you.  I will close PR #471 and revise PR #472.

david56:
PR #484 has enabled global config

PR #474, opened 2017-01-26
Accton AS5916_54X: add to support ONIE

david56:


machine/accton/accton_as5916_54x/busybox/patches/series
```diff
@@ -0,0 +1 @@
+onie-syseeprom-write-2b-delay-4000.patch
```
cbrune: maybe not for this platform, but I see this type of patch so often.....  perhaps we should make a busybox CONFIG_SYS_EEPROM_DELAY parameter that platforms can put in busybox/conf/config...

david56: Agree.  Generally speaking, it needs to generate a patch for each new platform.  Fortunately, it has only a few patch up to now that sharing between platforms.  For long-term support, using busybox config to tune the write delay would be better then using a patch.  I'll create an additional patchset for the change.  Thank you.

cbrune:
Pushed to master.  Thanks!

PR #478, opened 2017-02-01
busybox_1.25.1: updated config to restore stat filesystem features

pjferrell:
ONL installer was failing due to stat cmd not having --f option (filesystem features). Build options needed to be updated to account for behavioral change in upstream busybox (new feature flags added between 1.20.2 to 1.25.1).


cbrune:
Thanks, looks good.  Pushed to master:

- 5e2bfdd busybox_1.25.1: updated config to restore stat features

PR #472, opened 2017-01-26
Accton AS7316_54X: add to support ONIE 

david56:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on Broadcom
Tomahawk switch ASIC.  CPU module is Intel Xeon D-1518.

PR #479, opened 2017-02-07
busybox sys_eeprom: make i2c write delay be configurable

david56:
The patchset intends to move i2c write delay from patch file to busybox config.

Add options to make tuning i2c write delay more easily.

The options with default values are:

- CONFIG_SYS_EEPROM_I2C_WRITE_1B_DELAY=10
- CONFIG_SYS_EEPROM_I2C_WRITE_2B_DELAY=1000
- CONFIG_SYS_EEPROM_I2C_WRITE_3B_DELAY=10

These options can be ovrridden by $MACHINE/busybox/conf/config.

The patch has been tested in Accton AS7716_32X.

PR #472 and #474 are also updated to follow the new way.


cbrune:
I think this is a wonderful patch.  Good idea.

However, I am going to hold off on applying this patch until after the 2017.02 release goes out.  We are currently in the quiet period for that release.

patches/busybox/feature-onie-syseeprom-command.patch
```diff
@@ -58,6 +35,27 @@ index 0000000..c304273
...
++/*
++ * SYS_EEPROM_I2C_WRITE_1B_DELAY: delay number of microseconds after writing 1 byte
++ * SYS_EEPROM_I2C_WRITE_2B_DELAY: delay number of microseconds after writing 2 bytes
++ * SYS_EEPROM_I2C_WRITE_3B_DELAY: delay number of microseconds after writing 3 bytes
++ */
++#if CONFIG_SYS_EEPROM_I2C_WRITE_1B_DELAY == 0
++  #define SYS_EEPROM_I2C_WRITE_1B_DELAY   10
++#else
++  #define SYS_EEPROM_I2C_WRITE_1B_DELAY   CONFIG_SYS_EEPROM_I2C_WRITE_1B_DELAY
++#endif
```
cbrune: Is the check for 0 necessary?  The miscutils/Config.src file (down below) defines the default already.

The only way it could be zero is if the user explicitly set it to zero.

build-config/conf/busybox.config
```diff
@@ -777,6 +777,9 @@ CONFIG_WATCHDOG=y
 CONFIG_SYS_EEPROM_I2C_DEVICE=""
 CONFIG_SYS_EEPROM_I2C_ADDR=
 CONFIG_SYS_EEPROM_I2C_MEM_ADDR_BITS=0
+CONFIG_SYS_EEPROM_I2C_WRITE_1B_DELAY=0
+CONFIG_SYS_EEPROM_I2C_WRITE_2B_DELAY=0
+CONFIG_SYS_EEPROM_I2C_WRITE_3B_DELAY=0
```
cbrune: I would set these to the default values.

david56: > Is the check for 0 necessary? The miscutils/Config.src file (down below) defines the default already.
> 
> The only way it could be zero is if the user explicitly set it to zero.

Although the default values have been defined in `miscutils/Config.src`, the `busybox.config` generated by busybox-1.25.1 automatically still not has these values.  This deficiency did not appear in busybox-1.20.2.

The same symptom occurred recently is on `CONFIG_SYS_EEPROM_I2C_MEM_ADDR_BITS`:

https://github.com/opencomputeproject/onie/blob/master/patches/busybox/feature-onie-syseeprom-command.patch#L863-L870

I did not notice the issue before.  It added the option to all `CONFIG_SYS_EEPROM_DEVICE_I2C=y` platforms (see commits 977abaf and 0cea415) to fix compilation errors.

In the patchset, to prevent from these values resetting by re-generating `busybox.config` in the future, I would add check for 0 to the source rather than modify values in `busybox.config` manually.  Furthermore, setting them to zero explicitly should not make sense to access i2c EEPROM (the original value, 10ms, in `24cXX.c` should be the minimum delay).

What do you think?  Please let me know.  Thank you very much.

cbrune: I'm sorry, but I don't understand.  Under what circumstances do these values become "0" ?

I tried a couple of experiments:

1.  I tried with no busybox config.  I just unpacked the tarball, applied the patches and typed "make menuconfig" the busybox build directory with no pre-existing .config file present.  In this case I had to navigate the config menu and turn on ONIE system eeprom, which was expected.  Then I quit the menuconfig and saved the config.  The saved config had the default `[123]B_DELAY` values from misc/Config.src.  -- that all seems fine to me

2. another experiment -- I removed the `[123]B_DELAY` settings from our default config in build-config/conf/busybox.config.  I just removed the lines.  Then I "built" a machine that only enabled the ONIE syseeprom, without setting the `[123]B_DELAY` values.  That created a .config file that was just wrong -- the `[123]B_DELAY` values were not set at all.  That would lead to compile failures.  I think we should #error if these macros are unset in include/24cXX.h

3. another experiment -- the `[123]B_DELAY` values are set in our baseline config to their default values, like what I suggested up above.  This resulted in a build time busybox config that used the default values.  That seems correct to me.

One thing that came to mind during all this is perhaps we should set the `range` property on these configuration values.  For example:
```
config SYS_EEPROM_I2C_WRITE_2B_DELAY
        int "delay time for i2c writing 2 bytes"
        range 1 100000
        default 1000
        depends on SYS_EEPROM_DEVICE_I2C
        help
          Delay number of microseconds after writing 2 bytes.
```

This enforces the value is always greater than zero.

david56:
Thank you for your explanation.  The patchset is updated for applying changes mentioned above.  Besides, I also add a patch to revert `CONFIG_SYS_EEPROM_I2C_MEM_ADDR_BITS` back to 8.  Please review them again.  Thank you.

cbrune:
Pushed to master.  Thanks.

PR #484, opened 2017-02-13
busybox: enable swap commands

cbrune:
Enable the following busybox commands related to swap partitions:

* mkswap - set up a Linux swap area

* swapon, swapoff - enable/disable devices and files for paging and
                    swapping

Some NOS installers need these command available.


cbrune:
pushed to master:
- d96eb71 busybox: enable swap commands

PR #489, opened 2017-02-24
onie-mk-itb.sh: revise $10 to ${10}

david56:
The old version bash regards $10 as ${1}0. This makes onie-mk-itb.sh
creates itb file with ${1}0 name:

    accton_as6710_32x0

The patch revise $10 to ${10} to keep the script work with old version
bash.


cbrune:
Thanks!  Pushed to master:

- 13125b9 onie-mk-itb.sh: revise $10 to ${10}

PR #490, opened 2017-02-24
ls2088ardb: add dpaa2 support

ZhaoQiang-b45475:
add dpaa2 support which is for network processing acceleration.

Signed-off-by: Zhao Qiang <qiang.zhao@nxp.com>


cbrune:
Of course I have no way of testing if any of this works, so I am trusting you tested it :)

Pushed to master:

- 364e46b ls2088ardb: add dpaa2 support

PR #493, opened 2017-03-14
Celestica BigStone-G Support

scott-emery:
The Celestica BigStone-G platform is the open networking version of the Facebook
Backpack. Therefore, most of this patch is a copy of the Facebook Backpack
code with the names changed to refer to Celestica Bistone-G instead of Facebook
Backpack.


cbrune:
Looks great.  Thanks.

Pushed to master:

- a207ffd Celestica BigStone-G Support

PR #494, opened 2017-03-14
Create "resolv.conf" by default

pichuang:
"resolv.conf" is helpful for manually via network if operator want to get installer image

Signed-off-by: Phil Huang <phil_huang@edge-core.com>


cbrune:
What is the use case for this file?

If DHCP is configured then /etc/resolv.conf will be automatically created.

PR #495, opened 2017-03-21
installer: Set bootable flag in the GPT protective MBR partition table

cbrune:
For GPT partition tables, some buggy UEFI firmwares operating in CSM
(legacy BIOS mode) require that the disk have at least one MBR
partition that's marked as bootable/active to boot in BIOS mode.

See http://www.rodsbooks.com/gdisk/bios.html for more discussion.

When creating the GRUB-BIOS GPT partition, indicating that legacy BIOS
GRUB is being installed in a GPT partition, this patch uses parted to
set the bootable/active flag in the disk's protective MBR.

### DIFF
--- a/installer/grub-arch/install-arch
+++ b/installer/grub-arch/install-arch
@@ -336,16 +336,28 @@ create_gpt_partition()
         exit 1
     esac
 
-    local sgdisk_log=$(mktemp)
+    local tmp_log=$(mktemp)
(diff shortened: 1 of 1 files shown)


cbrune:
Pushed to master:
- 2715c63 installer: Set bootable flag in the GPT protective MBR partition table

PR #497, opened 2017-03-31
quanta_common_rangeley: Alias mbdetect from mb_detect for backward co…

audihsu-qci:
…mpatible

mbdetect is the name of formal release 2014.05 version, which is written as shell script
mb_detect is busybox executable for ONIE, alias to mbdetect for backward compatible
And also add it into install-platform to use the existed one


cbrune:
This is great.  Thank you!

Pushed to master:
- edc5dfd quanta_common_rangeley: Alias mbdetect from mb_detect for backward compatible

Issue #498, opened 2017-04-03
unable to ping to localhost or do self ping from onie


karthikbgt:
Hi,

I recently downloaded the onie for backpack to load the ONL. It was built in Debian 8 (Jessie) environment as per the build procedure mentioned for debian in onie wiki.
 
I am able to load the onie in the box. I tried the method of the installation of ONL via network and find that it is unable to configure interface using DHCPv4: eth0 and lands up using link-local IPv4 address. As a result of this, it is unable to fetch the onie-installer via network for ONL. 
On further analysis, I find that local ping or ping to localhost is itself not happening in onie and hence it is unable to communicate to other interfaces for fetching the onie-installer.. Since I am unable to to self ping or ping to localhost from onie, I am unable to fetch the NOS image over network.
 
Kindly let me know if I need to enable anything specifically in onie as part of build procedure ? or once onie is up so that i am able to ping to my localhost or self ping at onie…?
Is there any patch that should applied separately for networking …?  Also, is there any known issues in onie in network driver or other networking components ?

Thanks in advance,
Karthik Balaguru

cbrune:
This is an oversight that has been in ONIE forever.  The general fix is described here:

http://lists.opencompute.org/pipermail/opencompute-onie/2017-April/001383.html

PR #502, opened 2017-04-17
networking: bring up link on the loopback interface

cbrune:
The boot time networking script never brought the link up for the
loopback interface (lo).  This is a historical oversight that never
caused any problems with the ONIE services.

While not fatal, it is confusing that the loopback address, 127.0.0.1,
and the host name "localhost" do not function properly.

This patch brings up the "lo" interface as part of the system
networking initialization.

Closes: #498


cbrune:
Pushed to master:
- a3f47f9 networking: bring up link on the loopback interface

PR #503, opened 2017-04-20
busybox-1.25.1: revise pid file path to /var/run

david56:
This version of busybox has an option to set where pid files are
created.  Current configuration is empty that results in pid files
in the root directory.  The patch revises the path to `/var/run`.

The patch has been tested in Accton AS7716_32X.


cbrune:
- ee181d4 busybox-1.25.1: revise pid file path to /var/run

PR #504, opened 2017-04-20
Accton AS5916_54X: add to support r1 machine

david56:


cbrune:
Looks good.  Thanks for the patches.

Pushed to master:
- 2a234d9 Accton AS5916_54X: add to support r1 machine
- de56894 busybox.make: allow to specify machine's busybox directory

Issue #505, opened 2017-04-24
GRUB fragment 50_onie_grub is not quite right for UEFI


cbrune:
While trying to use installer/grub-arch/grub.d/50_onie_grub from a NOS where UEFI enabled I hit a small snag.

The `onie_uefi_grub` variable definition is a problem when called from a NOS:

```sh
...
# get running machine from conf file                                                                                                        
[ -r /etc/machine.conf ] && . /etc/machine.conf

default_chainloader="+1"
[ "${onie_firmware}" = "coreboot" ] && default_chainloader=/grub/i386-coreboot/core.elf
uefi_root=/boot/efi
onie_uefi_grub=EFI/onie/${onie_grub_image_name}
...
```

The problem is the `onie_grub_image_name` variable is defined in the `/etc/machine.conf` file.  When a NOS runs this fragment the `/etc/machine.conf` is unavailable -- that file is part of the ONIE initramfs.  The upshot is that the UEFI chainloading entry is broken.

Note: this fragment works perfectly fine with the DEMO NOS as it has a /etc/machine.conf file.

cbrune:
The `onie_grub_image_name` only has two possible values:
- `grubx64.efi`
- `grubaa64.efi`

As a work around for current NOS suggest setting `onie_grub_image_name` before running `50_onie_grub`:

```sh
# for x86_64
export onie_grub_image_name=grubx64.efi
50_onie_grub
```
Going forward suggest removing `onie_grub_image_name` from `/etc/machine.conf`.  Instead just embed it directly into `50_onie_grub`, as that is only place it is used.

Note: for Secure Boot with shim the chainload entry would be `shimx64.efi`, not `grubx64.efi`.

PR #507, opened 2017-04-24
Set the UEFI boot loader statically

cbrune:
The 50_onie_grub GRUB fragment is intended to be executed by a NOS as
part of the NOS's /etc/grub.d scripts.

For UEFI systems this fragment has a dependence on the the ONIE file
/etc/machine.conf for the definition of the UEFI boot loader.
However, when the NOS is running ONIE's /etc/machine.conf is
unavailable.

For ONIE today, the UEFI boot loader has two possible values:

- grubx64.efi
- grubaa64.efi

These names are defined in the CPU architecture dependent Makefile
fragments in build-config/arch/<arch>.mk in the $(GRUB_IMAGE_NAME)
variable.

This patch passes the $(GRUB_IMAGE_NAME) value all the way through the
build system and sets the value statically in 50_onie_grub.

Closes: #505


cbrune:
pushed to master:
- da698f4 Set the UEFI boot loader statically

PR #510, opened 2017-04-27
add machine delta_wb2448

louiskao324:
add machine delta_wb2448


cbrune:
Thanks for posting code for a new machine.  That is great!

There are a number of small things with this patch we need to iron out before adding it to the code base.  Most of those I will detail inline in the code, but here I want to talk about the commit message itself.

All the commit message says is: add machine delta_wb2448

That is not a very informative commit message.  I like to see the commit message for a new machine describe the new machine in moderate detail, such as:
- what kind of CPU
- what kind of storage
- what switching ASIC
- how many front panel ports
- what kind of ports?  10G, 25G, 40G, 100G?  copper? SFP+? QSF28? ...

As an example look at this commit from Agema:
b1d7f72bcf85cfec9cc5dfb52890e889fe258195

```
commit b1d7f72bcf85cfec9cc5dfb52890e889fe258195
Author: Dave Hu <dave.hu@agemasystems.com>
Date:   Sat Feb 25 12:44:59 2017 -0500

    AGEMA AG7648C: add to support ONIE
    
        The machine has 48x1GbE/10GbE Copper  and 6x40GbE QSFP ports
        on Broadcom Trident2/Trident2+ switch ASIC.  CPU module is Intel Atom C2338.
```

machine/delta/delta_wb2448/i2ctools/i2ctools-sys-eeprom-platform.patch
```diff
@@ -0,0 +1,17 @@
+Add platform specific stuff for the sys-eeprom
```
cbrune: Using i2ctools to support system EEPROM access has been obsolete since March 2015.  See this email on the mailing list:

Use the busybox implementation instead.

As an example, again consult this Agema commit: b1d7f72bcf85cfec9cc5dfb52890e889fe258195

machine/delta/delta_wb2448/kernel/acpi_enforce.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/drivers/acpi/osl.c b/drivers/acpi/osl.c
+index de03938..fe12708 100644
+--- a/drivers/acpi/osl.c
++++ b/drivers/acpi/osl.c
+@@ -1610,7 +1610,7 @@ __setup("acpi_no_auto_serialize", acpi_no_auto_serialize_setup);
+ #define ENFORCE_RESOURCES_LAX    1
+ #define ENFORCE_RESOURCES_NO     0
+ 
+-static unsigned int acpi_enforce_resources = ENFORCE_RESOURCES_STRICT;
++static unsigned int acpi_enforce_resources = ENFORCE_RESOURCES_NO;
```
cbrune: Why is this needed?  I have not seen any other machine do this.

machine/delta/delta_wb2448/kernel/igb-bcm50210s.patch
```diff
@@ -0,0 +1,193 @@
...
+ #define M88E1000_PHY_SPEC_CTRL     0x10  /* PHY Specific Control Register */
+diff --git a/drivers/net/ethernet/intel/igb/igb_main.c b/drivers/net/ethernet/intel/igb/igb_main.c
+index 7dfbcde..7bd6c9d 100644
+--- a/drivers/net/ethernet/intel/igb/igb_main.c
++++ b/drivers/net/ethernet/intel/igb/igb_main.c
+@@ -68,6 +68,10 @@ static const char igb_driver_string[] =
+ static const char igb_copyright[] =
+ 				"Copyright (c) 2007-2014 Intel Corporation.";
+ 
++#if 1 /* add BCM B50210S phy combo mode initialize */
```
cbrune: Get rid of all these `#if 1 / #endif`.

If the code being bracketed needs to be there then put it in the patch -- no need to bracket it with `#if 1 / #endif`.  That is just cruft and does not belong in production code.

machine/delta/delta_wb2448/machine.make
```diff
@@ -0,0 +1,38 @@
...
+# Vendor ID -- IANA Private Enterprise Number:
+# http://www.iana.org/assignments/enterprise-numbers
+# Delta Electronics, Inc. IANA number
+VENDOR_ID = 2254 
+
+# This platform requires the PXE_EFI64 installer
+PXE_EFI64_ENABLE = yes
+
+# Default to msdos disk label for this platform
+PARTITION_TYPE = gpt
```
cbrune: The default partition type is `gpt`.  I would remove these lines.

machine/delta/delta_wb2448/machine.make
```diff
@@ -0,0 +1,38 @@
...
+
+# Console parameters
+CONSOLE_DEV = 0
+CONSOLE_FLAG = 1
```
cbrune: `CONSOLE_FLAG` is no longer a parameter -- it has no meaning.  Please remove this line.

louiskao324:
@cbrune thanks for your advice, I will check and update.

louiskao324:
@cbrune  I have updated the pull request , please help to check.  Thanks

machine/delta/delta_wb2448/kernel/acpi_enforce.patch
louiskao324: If we use ENFORCE_RESOURCES_STRICT, we will face below ACPI warning.
It will case i2c driver i801 init fail.

ACPI Warning: SystemIO range 0x000000000000E000-0x000000000000E01F conflicts with OpRegion 0x000000000000E000-0x000000000000E00F (\_SB_.PCI0.SBUS.SMBI) (20150410/utaddress-254)
ACPI: If an ACPI driver is available for this device, you should use it instead of the native driver

cbrune: Is that perhaps then a bug with your ACPI tables for this box?  I would not make this change without understanding what that error message is telling me.  Why is there a conflict?

cbrune:
Thanks for all the clean up.  It looks pretty good now.

My only concern is the ACPI work around -- are you sure you want that in the code forever?

louiskao324:
@cbrune I think ACPI work around is a temporarily method, and will still trace it side effect.

louiskao324:
@cbrune 
We reference kernel-parameters.txt to set the acpi_enforce_resources=no in our box.
	acpi_enforce_resources=	[ACPI]
			{ strict | lax | no }
			Check for resource conflicts between native drivers
			and ACPI OperationRegions (SystemIO and SystemMemory
			only). IO ports and memory declared in ACPI might be
			used by the ACPI subsystem in arbitrary AML code and
			can interfere with legacy drivers.
			strict (default): access to resources claimed by ACPI
			is denied; legacy drivers trying to access reserved
			resources will fail to bind to device using them.
			lax: access to resources claimed by ACPI is allowed;
			legacy drivers trying to access reserved resources
			will bind successfully but a warning message is logged.
			no: ACPI OperationRegions are not marked as reserved,
			no further checks are performed.

please help to submit the pull request.

cbrune:
I am trying to understand why this is necessary.

This will be something *every* NOS that runs on this box will have to do.  How will a NOS vendor know they need to put `acpi_enforce_resources=no` on the kernel command line for their NOS on this box?

This really points to a problem with the ACPI tables on this box.  The ACPI tables come from the BIOS firmware.  I would fix this problem there and not here in the kernel.  That would be better for everyone.

At a minimum can you document what "goes wrong" when a NOS does not set `acpi_enforce_resources=no`?  Which drivers affected?

louiskao324:
@cbrune Thanks for your advice, I have asked for our BIOS engineer to support this issue.

louiskao324:
@cbrune Our BIOS engineer ask for our BIOS vendor AMI
AMI says it is not a true "conflict", as both I/O ranges pertain to the same device (SMBUS controller). 
And suggested us use acpi_enforce_resources=no in KERNEL command line too.

cbrune:
OK, pushed to master:

- f94412c Delta WB2448: add to support ONIE

I updated the commit message to include a note about the NOS needing to specify `acpi_enforce_resources=no`.

PR #511, opened 2017-04-28
Change the platform string for Bigstone-G

scott-emery:
The platform string for the Celestica Bigstone-G (aka Facebook Backpack) is determined by reading the Part Number (0x22) and Label Revision (0x27) fields from the EEPROM and then using those fields to determine if the platform is a Fabric Card, Left Line Card, or Right Line Card.

This patch changes the platform string to simply be the values read from the EEPROM: <part number>[_<label revision>]. The fields are lowercased and hyphens are converted to underscores. ONIE will no longer be interpreting the EEPROM contents to determine the platform type. That responsibility is now placed on the OS installer.


cbrune:
Looks good.  Thanks.  Pushed to master:
- c460ed1 Change the platform string for Bigstone-G

PR #513, opened 2017-05-10
BigStone-G: Allow ONIE install when there are no EEPROM contents

scott-emery:
When the EEPROM is not initialized the platform was being misidentified on the BigStone-G platform. This prevented ONIE from being able to be installed because of a platform mismatch. Which, in turn, prevented the EEPROM from being programmed because ONIE was being used to program the EEPROM.

This patch corrects the problem by checking to see if the Part Number (0x22) field of the EEPROM is valid, instead of blindly using it. This prevents an invalid platform string from being created.


cbrune:
Please make a PR that only contains the one commit, without the merge commit.  The easiest way to do that is:

- close this PR
- create a new branch for this issue, based off of upstream master
- create a new PR using the new branch

This quickly boils down to "how do I keep my fork's master branch in sync with upstream?".  github has some docs or try google:  https://www.google.com/search?q=github+keep+fork+up+to+date

Personally, I follow the method described here:
https://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository

scott-emery:
I didn't realize that this PR had multiple commits. I'll close this PR and submit another. Thanks Curt.

PR #514, opened 2017-05-11
BigStone-G: Allow ONIE install when there are no EEPROM contents

scott-emery:
When the EEPROM is not initialized the platform was being misidentified on the BigStone-G platform. This prevented ONIE from being able to be installed because of a platform mismatch. Which, in turn, prevented the EEPROM from being programmed because ONIE was being used to program the EEPROM.

This patch corrects the problem by checking to see if the Part Number (0x22) field of the EEPROM is valid, instead of blindly using it. This prevents an invalid platform string from being created.


cbrune:
- a70f61d BigStone-G: Allow ONIE install when there are no EEPROM contents

PR #517, opened 2017-05-17
build: compile common utilities once per toolchain

cbrune:
Today ONIE re-compiles *everything*, except the toolchain, for each
machine.  This is really wasteful of time, CPU and disk space when
building multiple machines that share a toolchain.

Most ONIE components are platform independent.  For example the common
utilities, like e2fstools, dropbear, etc., only need to be built once
for each toolchain.

Some components are machine specific and must be built for each
machine, such as:

- Linux kernel
- U-Boot
- busybox  (contains machine specific i2c EEPROM access)
- i2ctools (contains machine specific i2c EEPROM access)

This patch introduces the notion of common "user" components (short
for user space) that only need to be built once per toolchain.  This
significantly improves the build times when building multiple
machines.

With this change the build directory structure now contains a common
user build directory, called "build/user/<toolchain>".  All the common
components are built here.

The other change is the behavior of "make clean".  Historically "make
clean" cleaned up the build space for a single machine.  Now that
common component build products are shared across machine builds, it
would be bad to remove the common components as part of "make clean".

For cleaning, this patch introduces two new clean related targets:

- machine-clean -- only cleans components that are specific to the
  machine, like the kernel, u-boot, busybox, i2ctools, etc.

- user-clean -- a super set including machine-clean, plus all the
  common user components

The old "clean" target is now a synonym for "machine-clean".

### DIFF
--- a/build-config/Makefile
+++ b/build-config/Makefile
@@ -1,7 +1,7 @@
 #-------------------------------------------------------------------------------
 #>
 #
-#  Copyright (C) 2013,2014,2015,2016 Curt Brune <curt@cumulusnetworks.com>
+#  Copyright (C) 2013,2014,2015,2016,2017 Curt Brune <curt@cumulusnetworks.com>
 #  Copyright (C) 2014,2015 Carlos Cardenas <carlos@cumulusnetworks.com>
 #  Copyright (C) 2014 david_yang <david_yang@accton.com>
 #  Copyright (C) 2014 Mandeep Sandhu <mandeep.sandhu@cyaninc.com>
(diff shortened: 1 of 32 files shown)


cbrune:
I tested by building a number of machines in a row.  See this report for the build time results:

http://mirror.opencompute.org/onie/docs/ONIE-common-build-time.pdf

The server used to build all these ONIE machines was in use for other purposes during the course of this test, so the numbers should not be interpreted literally.  However, the general trend is clear:  the first machine using a toolchain "takes a long while" (just like before), but subsequent machines within that toolchain family build much, much faster.

The horizontal axis of the graph did not have enough room to print all the machine labels, but the raw data for all the machines in test is included in the report.

david56:
Great!  This patch will help us to save time and disk space.  Everything looks good to me, but `user-clean` and `machine-clean` might be lost in the PHONY list.

cbrune:
@david56 - good catch.  I updated the PR after your comments.

david56:
@cbrune Thank you for your update.

cbrune:
Applied to master:

- a48f8ad231f

PR #518, opened 2017-05-17
Support host gcc-6.3.0

cbrune:
A couple of updates to the build infrastructure when the host version of GCC is 6.3.0.  GCC6 will soon be common across many Linux distributions.

Prior to this ONIE was building fine with host GCC version 4.9.2, which is the default of Debian 8 (jessie).

The goal is to have ONIE building on Debian 9 (stretch) when it is formally released.

The patches here were tested on a VM running Debian 9 RC3 with GCC version 6.3.0.  The patches were also verified on Debian 8 with GCC version 4.9.2.


cbrune:
Pushed to master:
- aec3633 grub2: update configure to work with gcc-6.3.0
- 7aba9af crosstool-ng: backport fix to compile gcc 4.9.2 using host gcc 6.3.0

PR #521, opened 2017-05-24
U boot version info

cbrune:
A few changes for U-Boot platforms during the ONIE update process.  This is for arm-32 and powerpc.

The changes involve the U-Boot variables that are created/updated when ONIE is installed/updated.


david56:
The variables, `ver`, `onie_uboot_version` and `onie_build_date`,  will disappear once U-Boot env restored to default (when checking CRC error, making install mode sticky or other possible operations).  So I think it would be better to make them be also in the default env.

cbrune:
That seems like a good idea.  Let me see what is involved with that.

cbrune:
I added a build time step that compiles `onie_build_date`, `onie_uboot_version` and `ver` into the default U-Boot environment block.

cbrune:
Pushed to master:

- 4352887 U-Boot: store version and build date in default environment
- faa80ca updater: store ONIE build date in U-Boot env
- f667653 updater: store U-Boot version string reliably
- 7b43509 updater: rename U-Boot env variable 'ver' to 'onie_uboot_version'

PR #522, opened 2017-05-25
u-boot: fix save-default-env-on-bad-crc issue

david56:
While reading u-boot-env from flash, it specifies `ENV_SIZE` as data
length.  `ENV_SIZE` is not equal to size of `env_t` so that it is
possible to read partial data from flash.  That will make crc error.

The patch has been tested in Accton AS4610_54.


cbrune:
Nice catch!  Thanks.

Pushed to master:
- 133a290 u-boot: fix save-default-env-on-bad-crc issue

PR #523, opened 2017-05-26
dropbear enable ECDSA keys

nshopik:
This is enable ECDSA keys which is shorter and faster than RSA

`-R` option in init generate keys as they needed, if they don't exist (as soon someone connects to SSH) makes faster startup times and better `urandom` for them

I didn't test this yet, creating pull for comments if any

### DIFF
--- a/rootconf/default/etc/init.d/dropbear.sh
+++ b/rootconf/default/etc/init.d/dropbear.sh
@@ -2,6 +2,7 @@
 
 #  Copyright (C) 2013 Dustin Byford <dustin@cumulusnetworks.com>
 #  Copyright (C) 2014 Curt Brune <curt@cumulusnetworks.com>
+#  Copyright (C) 2017 Nikolay Shopik <shopik@nvcube.net>
 #
 #  SPDX-License-Identifier:     GPL-2.0
 
@@ -18,15 +19,18 @@ ARGS="-m -B"
 
 RSA_KEY=/etc/dropbear/dropbear_rsa_host_key
 DSS_KEY=/etc/dropbear/dropbear_dss_host_key
+ECDSA_KEY=/etc/dropbear/dropbear_ecdsa_host_key
 
 [ -r /lib/onie/dropbear-arch ] && . /lib/onie/dropbear-arch
 
 get_keys() {
     # If keys are already present just return
-    [ -r "$RSA_KEY" ] && [ -r "$DSS_KEY" ] && return 0
+    [ -r "$ECDSA_KEY" ] && [ -r "$RSA_KEY" ] && [ -r "$DSS_KEY" ] && return 0
 
     get_keys_arch || {
         # If problems just make new keys in ramdisk
+        # genereate ecdsa key
+        dropbearkey -t ecdsa -s 256 -f $ECDSA_KEY > /dev/null 2>&1
         # genereate rsa key
         dropbearkey -t rsa -s 1024 -f $RSA_KEY > /dev/null 2>&1
         # genereate dss key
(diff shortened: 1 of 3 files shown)


nshopik:
I've tested this patch on KVM_x64 machine for now

rootconf/default/etc/init.d/dropbear.sh
```diff
@@ -14,31 +15,18 @@ daemon=/usr/sbin/$name
 
 [ -x $daemon ] || exit 0
 
-ARGS="-m -B"
+ARGS="-m -B -R"
```
cbrune: I don't think we need the -R, as our `get_keys()` implementation makes sure the keys exist.

cbrune:
I think this is generally OK.  Does not seem to heavy weight.

Can you please squash this down to a single commit, to make reviewing easier?

rootconf/default/etc/init.d/dropbear.sh
nshopik: Yeah I'll remove that, initially I though i could rework a bit around it instead using get_keys() and since we must generate keys on first start and save them to env

nshopik:
I'll incorporate requested change and squash it down to one commit. I was planning also update dropbear, but I found that recent release fixes only security issues that doesn't affect us (we use empty password by default)

nshopik:
This should be clean now

rootconf/u-boot-arch/sysroot-lib-onie/dropbear-arch
```diff
@@ -29,9 +40,13 @@ get_keys_arch() {
         dropbearkey -t dss -s 1024 -f $DSS_KEY > /dev/null 2>&1
     fi
 
-    if [ -z "$rsa_val" ] || [ -z "$dss_val" ] ; then
+    if [ -z "$ecdsa_var" ] || [ -z "$rsa_val" ] || [ -z "$dss_val" ] ; then
```
cbrune: typo --- should be "val" not "var"

rootconf/u-boot-arch/sysroot-lib-onie/dropbear-arch
```diff
@@ -29,9 +40,13 @@ get_keys_arch() {
...
         tmp_env=$(mktemp)
         # encode key values
+        if [ -z "$ecdsa_var" ] ; then
```
cbrune: same here, val not var

rootconf/u-boot-arch/sysroot-lib-onie/dropbear-arch
```diff
@@ -29,9 +40,13 @@ get_keys_arch() {
...
+            ecdsa_var=$(uuencode -m $ECDSA_KEY r | tr ' \n' '@#')
```
cbrune: should be 'val'

rootconf/u-boot-arch/sysroot-lib-onie/dropbear-arch
```diff
@@ -29,9 +40,13 @@ get_keys_arch() {
...
+            echo "$ecdsa_var $ecdsa_var" >> $tmp_env
```
cbrune: should be:
```
echo "$ecdsa_var $ecdsa_val" >> $tmp_env
```

nshopik:
fixed var/val typos

cbrune:
Looks good, thanks.

I tested it on the kvm_x86_64 and on a PowerPC/u-boot machine (nxp_p2020rdbpca).  Works OK.

Pushed to master:
- 8e11076 dropbear enable ECDSA keys

PR #524, opened 2017-05-27
Support Netberg Platform

raymondhuey:
Add Netberg Aurora 620 and Aurora 720 Platform Support


cbrune:
Hi @raymondhuey , thanks for contributing!

It looks like some of this code was cloned from the Quanta "common rangeley" code.  I am not a huge fan of the name `netberg_common_rangeley`.  I was never a fan of the Quanta "common rangeley" name either.

If it really is a common CPU module, then it probably has a part number. I would use that part number, or a short version of that part number, in the ONIE machine name.  It is much more descriptive. "Rangeley" does really mean anything.

Also I would drop "common" from the name.  The fact that it is common will be obvious from the code.

For example see the facebook_backpack_b632vg, where the `b632vg` refers to a common CPU module:
http://portwell.com/products/detail.php?CUSTCHAR1=PCOM-B632VG

More comments in the code.

machine/netberg/netberg_common_rangeley/INSTALL
```diff
@@ -0,0 +1,224 @@
...
+
+  Specification
+===============
+ - Modularized CPU board design
+ - 1 RJ-45 out-of-band management port (10/100/1000M)
+ - 1 console port (RJ-45 on Aurora 720, Mini-USB on Aurora 620)
+ - 1 8GB DDR3 SDRAM
+ - 1 SSD 32GB
+ - 1 USB 2.0 port
+
```
cbrune: Some of this detail about the CPU module should also go into the commit message.  It is very handy to find this info in the `git log`.

machine/netberg/netberg_common_rangeley/INSTALL
```diff
@@ -0,0 +1,224 @@
...
+|                                                    |+/-: Change Opt.         |
+|  Access Level            Administrator             |F1: General Help         |
+|                                                    |F2: Previous Values      |
+|                                                    |F3: Optimized Defaults   |
+|                                                    |F4: Save & Exit          |
+|                                                    |ESC: Exit                |
+\----------------------------------------------------+-------------------------/
+        Version 2.17.1245. Copyright (C) 2015 American Megatrends, Inc.        
+                                                                             AB
+
```
cbrune: These "screen shots" of the BIOS menus are *super* helpful.  Thanks so much for adding that.

machine/netberg/netberg_common_rangeley/busybox/patches/series
```diff
@@ -0,0 +1,3 @@
+onie-sys-eeprom-platform.patch
+mb_detect_c.patch
+mb_detect_h.patch
```
cbrune: All these "patches" logically belong in a single patch.

A patch should stand on its own, meaning it can be applied by itself and the code will still compile.  What you have here requires that all 3 patches be applied.

machine/netberg/netberg_common_rangeley/busybox/patches/onie-sys-eeprom-platform.patch
```diff
@@ -0,0 +1,278 @@
+diff --git a/include/applets.src.h b/include/applets.src.h
+index 71b8cbd..891fb67 100644
+--- a/include/applets.src.h
++++ b/include/applets.src.h
+@@ -345,6 +345,7 @@ IF_WHOAMI(APPLET_NOFORK(whoami, whoami, BB_DIR_USR_BIN, BB_SUID_DROP, whoami))
+ IF_YES(APPLET_NOFORK(yes, yes, BB_DIR_USR_BIN, BB_SUID_DROP, yes))
+ IF_ZCIP(APPLET(zcip, BB_DIR_SBIN, BB_SUID_DROP))
+ IF_SYS_EEPROM(APPLET_ODDNAME(onie-syseeprom, onie_syseeprom, BB_DIR_USR_BIN, BB_SUID_DROP, onie_syseeprom))
++APPLET_ODDNAME(mb_detect, mbdetect, BB_DIR_USR_BIN, BB_SUID_DROP, mbdetect)
```
cbrune: why define both names?

I know why Quanta did, because they had legacy platforms to support.

You have no legacy to support here.  Just use the one name from the start.

machine/netberg/netberg_common_rangeley/busybox/patches/onie-sys-eeprom-platform.patch
```diff
@@ -0,0 +1,278 @@
...
++    return i2c_write(bus, chip, 0, 1, &data, 1);
++}
++
++/**
++ * i2c_mux_release - release channel for i2c mux
++ */
++static inline __s32 i2c_mux_release(__s32 bus, __u8 chip)
++{
++    return i2c_mux_select(bus, chip, 0xff);
++}
```
cbrune: why is the i2c access on this platform so complex?

Can it be simpler?  We tried pretty hard to make the i2c access configurable via the busybox config.

For example see:
https://github.com/opencomputeproject/onie/blob/master/machine/accton/busybox/onie-syseeprom-select-slave-device.patch
https://github.com/opencomputeproject/onie/blob/master/machine/accton/accton_as7712_16x/busybox/conf/config

machine/netberg/netberg_common_rangeley/installer/install-platform
```diff
@@ -0,0 +1,25 @@
...
+
+# Overrides install.sh
+check_machine_image()
+{
+    [ -x "/usr/bin/mb_detect" ] && {
+        platform=$(/usr/bin/mb_detect -p)
+    }
+    [ -x "/usr/bin/mbdetect" ] && {
+        platform=$(/usr/bin/mbdetect -p)
+    }
```
cbrune: Again, why support both names?  Since this is the first I would only support one name.

machine/netberg/netberg_common_rangeley/kernel/driver-i2c-busses-ismt-set-bus_speed-100kHz.patch
```diff
@@ -0,0 +1,18 @@
...
+old mode 100644
+new mode 100755
+index 09a3e03..a82c962
+--- a/drivers/i2c/busses/i2c-ismt.c
++++ b/drivers/i2c/busses/i2c-ismt.c
+@@ -192,7 +192,11 @@ static DEFINE_PCI_DEVICE_TABLE(ismt_ids) = {
+ MODULE_DEVICE_TABLE(pci, ismt_ids);
+ 
+ /* Bus speed control bits for slow debuggers - refer to the docs for usage */
++#if 1 /* Foxconn: Set i2c bus_speed to 100kHZ */
```
cbrune: The `#if 1` does not belong in the patch.  The patch should only contain the actual code necessary.

Why is it necessary to set the bus speed to 100kHz?

machine/netberg/netberg_common_rangeley/kernel/series
```diff
@@ -0,0 +1,4 @@
+driver-igb-1.patch
+driver-igb-2.patch
```
cbrune: What do the patches igb-1 and igb-2 do?

machine/netberg/netberg_common_rangeley/machine.make
```diff
@@ -0,0 +1,39 @@
...
+# http://www.iana.org/assignments/enterprise-numbers
+# Netberg IANA number
+VENDOR_ID = 47294
+
+I2CTOOLS_ENABLE = yes
+I2CTOOLS_SYSEEPROM = no
+
+CONSOLE_SPEED = 115200
+CONSOLE_DEV = 1
+CONSOLE_FLAG = 0
```
cbrune: The `CONSOLE_FLAG` setting is no longer used.  Please remove this line.

machine/netberg/netberg_common_rangeley/rootconf/sysroot-init/rtc-sync.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+RTC_DS1339_I2C_SLAVE=0x69
+
+. /lib/onie/functions
+
+/usr/bin/i2cget -y 0 $RTC_DS1339_I2C_SLAVE &> /dev/null && {
+	log_info_msg "Initializing RTC ..."
+	/sbin/hwclock --hctosys
+	log_info_msg "Initializing RTC Done ..."
+}
```
cbrune: What does this fragment do?  Why is it necessary?

machine/netberg/netberg_common_rangeley/busybox/patches/onie-sys-eeprom-platform.patch
raymondhuey: The "mbdetect" will be removed.

machine/netberg/netberg_common_rangeley/installer/install-platform
raymondhuey: The "mbdetect" will be removed.

machine/netberg/netberg_common_rangeley/machine.make
raymondhuey: The "CONSOLE_FLAG" will be removed.

machine/netberg/netberg_common_rangeley/kernel/driver-i2c-busses-ismt-set-bus_speed-100kHz.patch
raymondhuey: 1. The "#if 1" will be removed.
2. Because the "CPLD" on Netberg's platform supports 100kHz bus speed only.

machine/netberg/netberg_common_rangeley/kernel/series
raymondhuey: The patch igb-1 is to add Broadcom phy bcm5461s support and patch igb-2 is the driver functions of bcm5461s. These two patches will be combined to a single patch.

machine/netberg/netberg_common_rangeley/busybox/patches/series
raymondhuey: These three patches will be combined to a single patch.

machine/netberg/netberg_common_rangeley/rootconf/sysroot-init/rtc-sync.sh
raymondhuey: This fragment seems not necessary and will be removed.

machine/netberg/netberg_common_rangeley/busybox/patches/onie-sys-eeprom-platform.patch
raymondhuey: It cannot be simpler. Because before reading or writing EEPROM, the channel on the I2C-bus switch must be selected.

cbrune: It seems like a lot of code to me.  Setting the i2c switch should be easy.  All you have to do write a write a byte to a i2c-switch config register before trying to read/write the EEPROM.  It should not take this much code to do that.

cbrune:
OK, we are making progress.  Getting rid of unnecessary code.

cbrune:
When all the changes are ready, you will need to squash these commits down into a single commit.

raymondhuey:
Hi @cbrune,
All the changes are ready, please review this commit.
Thanks.

cbrune:
Looks good, thanks!  Pushed to master:

- 88304f9 Add Netberg Aurora 620 and Aurora 720 Platform Support

PR #527, opened 2017-06-22
Accton AS7212_54X: add to support ONIE

david56:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 on Marvell 98CX8420 switch ASIC.  CPU module is Intel Rangeley C2538.


cbrune:
Pushed to master, thanks!

- a7e5dcb Accton AS7212_54X: add to support ONIE
- ebbc9de add 'mvl' to KNOWN_ASIC_VENDORS list

PR #531, opened 2017-06-23
Reduce verbosity during boot

cbrune:
This series reduces the boot time chattiness.  Everything import is still logged to /var/log/messages.


cbrune:
pushed to master:

- 3e654b9 networking: reduce the verbosity of udhcpc
- a3e9632 console: reduce printk() default verbosity

PR #532, opened 2017-06-23
klogd: add and enable daemon

cbrune:
Start the klogd daemon at boot time, which logs the kernel dmesg
output to /var/log/messages via syslog.

This ensures the dmesg information is logged into /var/log/messages.


cbrune:
pushed to master: 

- f159f32 klogd: add and enable daemon

PR #533, opened 2017-06-23
qemu_armv7a: update demo OS install parameters

cbrune:
Correct a little bit rot.

The offset and size of the ONIE image have changed slightly over
time.  This patch corrects that.


cbrune:
pushed to master:

- 9bced2a qemu_armv7a: update demo OS install parameters

PR #535, opened 2017-06-26
Update toolchain to gcc-6.30 and uClibc-ng-1.0.22

cbrune:
At a high level this pull request moves the toolchain forward to modern version of the components.  Specifically the new versions are:

- crosstool-NG 1.23.0
- gcc 6.3.0
- uClibc, is now uClibc-ng-1.0.22

Also the default Linux kernel version moved from the 4.1.x series to the 4.9.x series.  The only machine currently using this kernel version is the kvm_x86_64 machine.  All other machine continue to use the same kernel version as before.

To keep machines as they were before I went through and set the versions of Linux and GCC to what they were before as necessary.

For the Linux kernel, where machines were using the default, 4.1.38, I edited the corresponding machine.make to *explicitly* select kernel 4.1.38.  Thus these machines continue to build with their current patch sets.

For older machines that use Linux kernel 3.2.69 the new gcc-6.3.0 will not work.  For these machine I edited the corresponding machine.make to *explicitly* specify gcc-4.9.4, which is the same family as before (old was 4.9.2) but a slightly updated version that crosstools-NG 1.23.0 supports directly.

GCC-6.3.0 does not work directly with the older u-boot versions either, however the patches to make it work are trivial and already upstream in the u-boot project.  Most of the ONIE machines that use the old u-boot are also using kernel 3.2.69, which from the previous paragraph are using the old compiler anyway and no changes are needed.

For machines under my control, like the nxp_p2020 and the armv7a virtual machine I went ahead and patched u-boot 2013.01.01 to work with gcc-6.3.0.  If you are interested in using gcc-6.3.0 with older u-boot consult the u-boot 2013.01.01 patchset.

After all these changes I have verified the following:

- kvm_x86_64  -- using kernel 4.9.33. Compiles and runs in legacy mode and UEFI modes
- dell_s6000_s1220 (x86_64) -- compiles and runs in legacy BIOS mode
- nxp_p2020rdbpca (a PowerPC P2020 eval board, u-boot) -- compiles and runs OK
- accton_as5600_52x (a PowerPC P2020, u-boot, with storage limitations) -- compiles and runs OK
- qemu_armv7a (armv7a virtual machine, u-boot) -- compiles and runs OK
- accton_as4610_54 (dual-core armv7a, u-boot) -- compiles and runs OK
- nxp_ls1043ardb (armv8a, 64-bit ARM) -- compiles. Unable to run as I don't have one

PR #537, opened 2017-06-28
gcc-4.9.2: enable to support powerpc64 target

david56:
This option was removed accidentally in commit 44a1f46.

The patch enables the option again.

The patch has been tested in Accton AS7710_32X.

@cbrune , the patch should also sync to gcc-4.9.4 in PR #535


cbrune:
@david56 - great, thanks for finding that.

I will also make sure it gets into PR #535.  Thanks again.

Pushed to master:

- 77483d4 gcc-4.9.2: enable to support powerpc64 target

PR #538, opened 2017-06-30
Add support for QuantaMesh Broadwell-DE with BMC CPU board

audihsu-qci:
This pull request conains two commits.
1. file system: Add support for NVMe SSD
2. Add support for QuantaMesh Broadwell-DE with BMC CPU board

Due to QuantaMesh Broadwell-DE with BMC CPU board platform uses
M.2 NVMe SSD as primary storage, I seperated this into two commits.
The CPU board support depends on the NVMe SSD support.


build-config/scripts/make-devices.pl
```diff
@@ -179,6 +179,12 @@ sub devices
 	makedev("mmcblk0p${i}", "b", 179, ${i}, ${root}, ${root}, 0660);
     }
 
+    print "NVMe SSD access: nvme0n1 nvme0n1p{1..7}\n";
+    makedev("nvme0n1", "b", 259, 0, ${root}, ${root}, 0660);
+    foreach my $i (1..7) {
+    makedev("nvme0n1p${i}", "b", 259, ${i}, ${root}, ${root}, 0660);
+    }
```
cbrune: I do not think creating these device nodes at compile time is going to work in general.
The major number used here, 259, is not necessarily the number the Linux kernel will choose when the nvme driver is registered.  259 might be the correct number on your system, but in general it will not be correct.

I have two thoughts one this:

1. It might be time to move away from the static dev names.  Use a helper program, like `mdev`, to populate /dev/ dynamically, similar to how udev works.

2. use a hybrid approach, like how the VIRT_IO disk devices are create for the virtual machine.   These are the `/dev/vd[a-z]` devices created in /etc/init.d/makedev.sh. https://github.com/opencomputeproject/onie/blob/master/rootconf/default/etc/init.d/makedev.sh#L93-L117

Can you try using `mdev` and see if helps here.  Here's the help on `mdev`:
```
ONIE:/ # mdev --help
BusyBox v1.25.1 (2017-06-22 15:39:41 PDT) multi-call binary.

Usage: mdev [-s]

mdev -s is to be run during boot to scan /sys and populate /dev.

Bare mdev is a kernel hotplug helper. To activate it:
        echo /sbin/mdev >/proc/sys/kernel/hotplug

It uses /etc/mdev.conf with lines
        [-][ENV=regex;]...DEVNAME UID:GID PERM [>|=PATH]|[!] [@|$|*PROG]
where DEVNAME is device name regex, @major,minor[-minor2], or
environment variable regex. A common use of the latter is
to load modules for hotplugged devices:
        $MODALIAS=.* 0:0 660 @modprobe "$MODALIAS"

If /dev/mdev.seq file exists, mdev will wait for its value
to match $SEQNUM variable. This prevents plug/unplug races.
To activate this feature, create empty /dev/mdev.seq at boot.

If /dev/mdev.log file exists, debug log will be appended to it.
```

I do not think we need to use the config file.  I think we just need to activate it.

The mdev helper could be activated in /etc/init.d/makedev.sh.

You could test it by first activating mdev, then create a couple partitions using `sgdisk`.  Then check if the device nodes were created in /dev/.  Something like this:
```
# echo /sbin/mdev >/proc/sys/kernel/hotplug
# sgdisk -n 4:: /dev/nvme0n1
# ls -l /dev/nvme0n1*
```
If that is working we should consider activating mdev for hotplug by default.

installer/grub-arch/install-arch
```diff
@@ -459,6 +461,21 @@ init_uefi_disk()
...
+        dev=$(basename $nd)
+        major=$(sed -e 's/:.*$//' $nd/dev)
+        minor=$(sed -e 's/^.*://' $nd/dev)
+        rm -f /dev/$dev
+        mknod /dev/$dev b $major $minor || {
+            echo "Problems creating /dev/$dev block device."
+            continue
+        }
+        done
+    }
```
cbrune: If the mdev hotplug helper is working properly we should not need this code here.

rootconf/default/lib/onie/onie-uninstaller-common
```diff
@@ -80,8 +80,13 @@ erase_part()
...
     local blk_dev="$1"
     local part="$2"
+    # check if we have an mmcblk device
+    local blk_suffix=
+    echo ${blk_dev} | grep -q mmcblk && blk_suffix="p"
+    # check if we have an nvme device
+    echo ${blk_dev} | grep -q nvme && blk_suffix="p"
 
-    erase_block_device "${blk_dev}$part"
+    erase_block_device "${blk_dev}$blk_suffix$part"
```
cbrune: Thanks so much for fixing this!  Looks like this is a long standing bug for mmcblk devices.

machine/quanta/quanta_bwde/rootconf/sysroot-lib-onie/sysinfo-platform
```diff
@@ -0,0 +1,21 @@
+
+#  Copyright (C) 2014 Curt Brune <curt@cumulusnetworks.com>
+#
+#  SPDX-License-Identifier:     GPL-2.0
+
+get_serial_num()
+{
+    [ -x /usr/bin/onie-syseeprom ] && /usr/bin/onie-syseeprom -g 0x23
+}
```
cbrune: This function is not needed, as the grub-arch default implementation does the same thing:
https://github.com/opencomputeproject/onie/blob/master/rootconf/grub-arch/sysroot-lib-onie/sysinfo-arch#L11

machine/quanta/quanta_bwde/rootconf/sysroot-lib-onie/init-platform
```diff
@@ -0,0 +1,33 @@
...
+# Use this function to perform any initializations required by the
+# architecture specific initialization.  This function executes before
+# the architecture initializations.
+init_platform_pre_arch()
+{
+    echo "Running init_platform_pre_arch..."
+    echo "Mount devtmpfs..."
+    mount -t devtmpfs devtmpfs /dev
+    mkdir /dev/pts
+    mount -t devpts devpts /dev/pts
```
cbrune: Did mounting devtmpfs solve a particular problem?

It looks useful.  Also combined with mdev could be very interesting.

If it is useful, we should considering making it the default for machines.

cbrune:
This is really great, adding nvme support.  Really nice.  Thanks.

The new machine definition looks good too.

We have an opportunity here to straighten out the dynamic device node handling, which we should do.

machine/quanta/quanta_bwde/kernel/config
```diff
@@ -0,0 +1,18 @@
+CONFIG_MODULES=y
+CONFIG_DEVTMPFS=y
+CONFIG_DEVTMPFS_MOUNT=y
```
cbrune: according to the kernel docs, with DEVTMPFS_MOUNT=y, the devtmpfs filesystem is automounted by the kernel on /dev right after the initramfs rootfs is mounted.  there should be no need to mount it manually, like what you have below.

machine/quanta/quanta_bwde/rootconf/sysroot-lib-onie/init-platform
cbrune: I played around a bit enabling devtmpfs by default in the kernel and mounting it at boot time.  This ignores all the devices that the make-devices.pl script creates at build time.

It looks promising.  I was also able to remove a bunch of the dynamic device creation code from /etc/init.d/makedev.sh.

This was only with the kvm_x86_64, so it would need quite a bit more testing on other CPU architectures and devices to become the ONIE default.  Kernel configs would also need to be updated to enable the devtmpfs feature.

I won't have a chance to look into this for a while.  It is a long holiday weekend in the US.

cbrune: Created #539 to capture this idea.

audihsu-qci: Actually, we found that CONFIG_DEVTMPFS_MOUNT does not work with cpio.xz format. It just works with initramfs in previous kernel version. Therefore, the common way to make this work is to add ``mount -t devtmpfs none /dev`` in the very first part of system init, such as inittab or init script.

audihsu-qci: The way to change /dev nodes creation to CONFIG_DEVTMPFS_MOUNT may take quite a lot of efforts to try and test. I think I will have a slightly modification here for this platform, and then try to move forward for /dev nodes auto mount mechanism.

PR #540, opened 2017-07-03
Add support for QuantaMesh Broadwell-DE with BMC CPU board

audihsu-qci:
This pull request contains two commits.

    file system: Add support for NVMe SSD
    Add support for QuantaMesh Broadwell-DE with BMC CPU board

Due to QuantaMesh Broadwell-DE with BMC CPU board platform uses
M.2 NVMe SSD as primary storage, I seperated this into two commits.
The CPU board support depends on the NVMe SSD support.

Note that NVMe device nodes in /dev is not pre-created.
QuantaMesh uses ``CONFIG_DEVTMPFS=y`` in kernel config, and mounts devtmpfs
in the very first script of ``/etc/rcS/S01mount-devtmpfs.sh``,
that ensures NVMe device nodes work properly.


demo/installer/grub-arch/install.sh
```diff
@@ -296,6 +298,9 @@ demo_install_uefi_grub()
 
 eval $create_demo_partition $blk_dev
 demo_dev=$(echo $blk_dev | sed -e 's/\(mmcblk[0-9]\)/\1p/')$demo_part
+echo $blk_dev | grep -q nvme && {
```
cbrune: have you tried this on a non-NVME box?

I think the above test should use `if .... ; then ; fi`.  Without it the failure of the `grep` will trigger an overall script failure, as the script is run with `set -e`.

cbrune:
I added one more commit that moves uClibc ahead to the default version, as the old version is no longer supported.

Pushed to master:
- ed5108d file system: Add support for NVMe SSD
- 006451d Add support for QuantaMesh Broadwell-DE with BMC CPU board
- 1eecdd1 quanta_bwde: use updated uClibc

PR #541, opened 2017-07-05
Support of new Lenovo G8296 switch

psarker:
This is to support ONIE on Lenovo G8296 switch.


cbrune:
The file `build-config/conf/onie-release` should never be committed to the master branch.

That file is only used in release branches.

What are you trying to do?

psarker:
I have not tried to commit anything yet.. Cant find any clue where from "build-config/conf/onie-release" has come.. Im surprised !! ..

cbrune:
Closing this, as it appears to be a mistake.

PR #543, opened 2017-07-12
Update Kernel Version to 4.9.35

jonathantsai-qci:
Please review and merge code. Thanks a lot.


cbrune:
Nice to see you could drop the big PHY kernel patch with this update.  Great.

Pushed to master:
- 540a106 quanta_ly4r: use gcc-6.3.0 and kernel-4.9.35
- 9664c46 quanta_common_rglbmc: use gcc-6.3.0 and kernel-4.9.35

PR #544, opened 2017-07-14
gcc-4.9:  go back to version gcc-4.9.2 instead of gcc-4.9.4

cbrune:
Commit a613240 updated to a new version of crosstool-NG and moved gcc-4.9.2 forward to gcc-4.9.4:

  commit a613240
  Date:   Fri Jun 23 14:02:09 2017 -0700

      gcc-4.9.x: upgrade to gcc-4.9.4

Using gcc-4.9.4 all machines and architectures were able to compile,
however some architectures experience run time crashes.  In particular
the powerpc P2020 machines experience kernel crashes using Linux
kernel 3.2.69.

The decision here is to go back to using gcc-4.9.2.  This also entails
moving all the gcc "companion libraries" and binutils back to the
versions used in the ONIE build system prior to commit
a613240.

With these changes the old machines, for all CPU architectures,
compile and run Linux kernel 3.2.69.

The versions of components used, which matches the previous gcc-4.9.2:

- gcc-4.9.2
- binutils-2.24
- gdb-7.11.1
- gmp-6.0.0a
- mpfr-3.1.2
- isl-0.12.2
- mpc-1.0.2
- expat-2.1.1
- strace-4.9


cbrune:
Pushed to master.

Issue #539, opened 2017-06-30
use devtmpfs to populate /dev


cbrune:
Move away from the static device creation script, make-devices.pl.  Let the kernel manage the /dev nodes for us.

In the relevant kernel configs need to enable:
```
CONFIG_DEVTMPFS=y
```

Maybe consider enabling this one as well, which will auto mount the devtmpfs on /dev:
```
# CONFIG_DEVTMPFS_MOUNT is not set
```
After that is done clean up /etc/init.d/makedev.sh to remove all the dynamic device node creation.

It may still be necessary to use `mdev` to manage nodes.  Some R&D required here.

PR #545, opened 2017-07-14
sysroot: use devtmpfs instead of static /dev node

cbrune:
For the longest time ONIE used static /dev nodes, which were created
at image build time.  Initially this worked well enough, but over time
as more devices and storage systems came into wide spread use the
static /dev nodes became painful.

This problem was worked around for a few specific /dev nodes, like
/dev/vda* and /dev/btrfs-control, but ultimately that solution is too
painful.

This patch removes all the static /dev node creation and instead uses
the kernel CONFIG_DEVTMPFS feature.  Quoting the kernel documentation
on this feature:

  This creates a tmpfs/ramfs filesystem instance early at bootup.  In
  this filesystem, the kernel driver core maintains device nodes with
  their default names and permissions for all registered devices with
  an assigned major/minor number.  Userspace can modify the filesystem
  content as needed, add symlinks, and apply needed permissions.  It
  provides a fully functional /dev directory, where usually udev runs
  on top, managing permissions and adding meaningful symlinks.  In
  very limited environments, it may provide a sufficient functional
  /dev without any further help. It also allows simple rescue systems,
  and reliably handles dynamic major/minor numbers.

Very early during system boot up, the devtmpfs file system is mounted
on /dev and off we go.

To enable this feature across all platforms, the base kernel
configuration files are modified to enable CONFIG_DEVTMPFS and
CONFIG_DEVTMPFS_MOUNT.

Closes: #539


cbrune:
Pushed to master:

- a047959 sysroot: use devtmpfs instead of static /dev node
- 46c2f91 quanta_ly4r: remove explicit mounting of devtmpfs
- 65cf993 quanta_common_rglbmc: remove explicit mounting of devtmpfs
- a8e85fe quanta_bwde: remove explicit mounting of devtmpfs

PR #547, opened 2017-07-17
grub: explicitly disable build time grub-mkfont

cbrune:
The GRUB autoconf configure script checks the host build system for
the presence of the freetype development libraries.  If found, the
configure script enables building and running grub-mkfont at compile
time.

This causes problems depending on what fonts are installed on the
build system and the build can fail.

Historically this has never been a problem as the freetype development
libraries were never installed on the build host.  Recently, however,
on a build system I use the freetype development libraries were
present.  Hence I hit this problem.

This patch explicitly disables the building and running of the
grub-mkfont utility.

We have already disabled the building of grub-mkfont for run time
previously.  This patch make sure we do not compile the "build time"
utility as well.


cbrune:
Pushed to master.

PR #548, opened 2017-07-17
Boot time network driver initialization hook

cbrune:
This patch series is the first step in implementing the proposal to enable the switching silicon ASIC in ONIE.  This would allow the front panel ports to participate in the ONIE image discovery mechanisms.

The proposal is from the mailing list:
http://lists.opencompute.org/pipermail/opencompute-onie/2017-April/001387.html


cbrune:
pushed to master

PR #553, opened 2017-07-19
Add ONIE support on a new Lenovo NE1032 platform 

psarker:
Adding a new Lenovo NE1032 platform


cbrune:
Thanks for the new platform.

Very minor thing, which I just noticed after pushing the code -- the commit message title is wrong.  It is identical to dc7d830.  It probably should have been `Adding ONIE support on a new Lenovo NE1032 platform` without the `T`.  Oh well, :)

Pushed to master:
- 55d4757

PR #556, opened 2017-07-20
Add ONIE support on a new Lenovo NE10032 platform 

psarker:
Adding ONIE support on a new X86 based Lenovo platform(NE10032)


machine/lenovo/lenovo_ne10032/rootconf/sysroot-lib-onie/sysinfo-platform
```diff
@@ -0,0 +1,25 @@
+
+get_serial_num()
```
cbrune: you don't need this function, as this is the default behavior:

https://github.com/opencomputeproject/onie/blob/master/rootconf/grub-arch/sysroot-lib-onie/sysinfo-arch#L11-L21

psarker: Thank you Curt. yes, this is not required.. I have removed it and placed another push ..

psarker:
machine/lenovo/lenovo_ne10032/rootconf/sysroot-lib-onie/sysinfo-platform is removed .. thank you for pointing this. Cheers!.

cbrune:
Thanks.

I squashed 80960ed into 5960c61 and pushed the result to master:

- 9940029 Add ONIE support to a new Lenovo X86 platform NE10032

PR #557, opened 2017-07-20
Accton AS5610_52X u-boot: tune DDR configurations

david56:


cbrune:
Applied to master:
- 2778fa0

Thanks!

PR #558, opened 2017-07-20
Enable UEFI features in some Accton X86_64 platforms

david56:


cbrune:
Very nice, pushed to master:
- d631baa
- 2b67afb
- a7108c0

machine/accton/accton_as7712_32x/machine.make
```diff
@@ -23,6 +23,9 @@ endif
 # Accton Technology Corporation IANA number
 VENDOR_ID = 259
 
+UEFI_ENABLE = yes
+PXE_EFI64_ENABLE = yes
```
cbrune: I know I merged this already, but I could not get this machine to compile.

With UEFI enabled the build of `efivars` fails, like this:
```
linux.c:29:24: fatal error: linux/nvme.h: No such file or directory
 #include <linux/nvme.h>
                        ^
compilation terminated.
/work/marvel-06/curt/onie/build/user/x86_64-g4.9.2-lnx3.2.69-uClibc-ng-1.0.22/efivar/efivar-0.20/Make.rules:14: recipe for target 'linux.o' failed
```

The problem is this machine is using kernel 3.2.69, which is too old.  It does not provide nvme.h.

UEFI (efivar) requires using kernel 4.1.38 or later.

The as7312_54x likely has the same problem.

david56: Thank you for your friendly reminder.  I will send another PR to fix the problem.

PR #559, opened 2017-07-21
Accton X86_64 Platforms: use linux-4.1.38 and gcc-4.9.2

david56:


cbrune:
What is the reason to use gcc-4.9.2?  Was there a problem using gcc-6.3.0 and kernel 4.1.38 for x86_64?

I had to fix a minor thing to use gcc-6.3.0 with kernel 4.1.38 for powerpc-softfloat, but it was trivial:
https://github.com/opencomputeproject/onie/blob/master/patches/kernel/4.1.38/arch-powerpc-fix-out-of-bounds-array.patch

Anyway, I pushed all this master to move forward for now.

PR #561, opened 2017-07-26
Accton P9532X: add to support ONIE

david56:
The machine has 32x100GbE QSFP ports on Broadcom Tomahawk switch
ASIC.  CPU module is Intel Xeon D-1557.


cbrune:
Pretty cool, pushed to master:

- c94dace Accton P9532X: add to support ONIE

PR #562, opened 2017-07-26
Accton Wedge100S_32X: add to support ONIE

david56:
This machine has 32x100GbE QSFP ports on Broadcom Tomahawk switch


machine/accton/accton_wedge100s_32x/rootconf/sysroot-lib-onie/init-platform
```diff
@@ -0,0 +1,36 @@
...
+	# wait for the cp2112 USB-to-I2C bridge to come up.
+	# USB device initialization can happen asynchronously
+	# and late.  We want the EEPROM to be available before
+	# the rest of user-level initialization.
+	#
+	sleep 1.5
+	i=100
+	while true; do
+		if [ -r /sys/bus/i2c/devices/i2c-0/new_device ] ; then
+			echo pca9548 0x74 > /sys/bus/i2c/devices/i2c-0/new_device
```
cbrune: are you sure you want these echo lines executed every iteration of the loop?

machine/accton/accton_wedge100s_32x/rootconf/sysroot-lib-onie/init-platform
```diff
@@ -0,0 +1,36 @@
...
+			echo 24c64 0x50 > /sys/bus/i2c/devices/i2c-7/new_device
```
cbrune: do you need to check if `/sys/bus/i2c/devices/i2c-7/new_device` exists before trying to write to it?

david56: Hello @cbrune ,
The key point here would be waiting for `i2c-0` ready whose hardware's init is done by other system (BMC).  Once the device is ready, everything behind it would go on smoothly.  So the write expressions are executed only one time and the EEPROM comes up and exists from the loop.  The expressions were tested several times and looks OK.

BTW I've revised the expressions slightly to be more reasonable.

cbrune:
Looks good - pushed to master:

- 2cc147d Accton Wedge100S_32X: add to support ONIE

PR #564, opened 2017-07-26
Accton AS7116_54X: add to support ONIE

david56:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on
MediaTek/Nephos NP8365 switch ASIC.  CPU module is Intel Xeon
D-1518.


cbrune:
Pushed to master:

- cb63c85 Accton AS7116_54X: add to support ONIE

Issue #565, opened 2017-07-30
clean: make `I2CTOOLS_SYSEEPROM = no` the default


cbrune:
An old legacy default for x86_64 machines is `I2CTOOLS_SYSEEPROM = yes`.  See this:
https://github.com/opencomputeproject/onie/blob/2017.05br/build-config/arch/x86_64.make#L126

This is dumb.  We moved to using busybox for syseeprom access long ago.

The time come to change the default of this option to `no`.

The cleanup would be to:

1. change the default to `no`
2. visit all the x86_64 machine.make files that are relying on the old default and specifically add `I2CTOOLS_SYSEEPROM = yes`

That list is found with this grep:
```
boson:~/src/onie-ocp$ find machine -name machine.make | xargs grep -L 'I2CTOOLS_SYSEEPROM = no' | xargs grep -l x86_64
machine/alphanetworks/alphanetworks_snh60a0_320f/machine.make
machine/alphanetworks/alphanetworks_snq60a0_320f/machine.make
machine/alphanetworks/alphanetworks_snx60a0_486f/machine.make
machine/celestica/cel_redstone_xp/machine.make
machine/celestica/cel_smallstone_xp/machine.make
machine/dell/dell_s6000_s1220/machine.make
machine/imt/im_n29xx_t40n/machine.make
machine/mellanox/mlnx_x86/machine.make
machine/qemu_armv8a/machine.make
```

That is not much work.

PR #569, opened 2017-08-01
i2ctools: set I2CTOOLS_SYSEEPROM=no as the default

cbrune:
Using the system EEPROM support in i2ctools has been deprecated since
2014.  The accepted method is to use the support added to busybox with
this commit:

  commit fc76004dca594a5f4f12a7905474f77c67e5effb
  Author: david_yang <david_yang@accton.com>
  Date:   Thu Sep 18 09:33:01 2014 +0800
   
      Add to support onie-syseeprom command in busybox

The legacy default for x86_64 machines is I2CTOOLS_SYSEEPROM=yes,
which is painful as all new machines have to explicitly set it to "no"
as they are using the busybox implementation.

This is dumb. The time come to change the default of this option to
no.

This patch also goes through all the legacy machines using the
i2ctools syseeprom implementation, which were relying on the old
default value of "yes", and explicitly sets the option to "yes" in the
corresponding machine.make file.

Closes: #565

### DIFF
--- a/build-config/arch/x86_64.make
+++ b/build-config/arch/x86_64.make
@@ -125,11 +125,6 @@ GRUB_IMAGE_NAME = grubx64.efi
 # override this.
 I2CTOOLS_ENABLE ?= yes
 
-# The onie-syseeprom in i2ctools is deprecated from now on.
-# It is recommended to migrate the support from i2ctools to busybox.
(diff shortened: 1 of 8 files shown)


cbrune:
Pushed to master:
- e633c56 i2ctools: set I2CTOOLS_SYSEEPROM=no as the default

Issue #567, opened 2017-07-31
Adding ".bin" file name extensions to Waterfall


ericpulvino:
It is somewhat common for users to leave ".bin" file extensions attached when renaming files. The error is so trivially simple that most will miss it. I've seen multiple users leave the .bin file extensions in place which prevents ONIE from finding installers, this is especially common when performing USB-based installations. 

When the file extension is not detected, the USB stick then has to be mounted manually and manually installed. This process can make Open Networking look harder than it is.

PR #571, opened 2017-08-01
Add ".bin" extension to default waterfall file names

cbrune:
In the field a number of user errors have been reported where the
setting of the default ONIE waterfall image is incorrect.  In most
cases the user has mistakenly appended ".bin" to the simple base name
"onie-installer".

The error is so trivially simple that most will miss it, which
prevents ONIE from finding installers.  This is especially common when
performing USB-based installations.

When the file extension is not detected, the USB stick then has to be
mounted manually and manually installed. This process can make Open
Networking look harder than it is.

This patch adds infrastructure for appending any number of suffixes to
the waterfall image names.  This infrastructure is then used to append
".bin" to the default names.

Closes: #567

### DIFF
--- a/rootconf/default/bin/discover
+++ b/rootconf/default/bin/discover
@@ -86,7 +86,7 @@ sd_localfs()
     local_parts=
     for p in $part_list ; do
         mount $p $mp > /dev/null 2>&1 && {
-            for f in $onie_default_filenames ; do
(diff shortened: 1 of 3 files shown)


cbrune:
Pushed to master:

- b61fcba Add ".bin" extension to default waterfall file names

PR #573, opened 2017-08-03
Set correct dns server ip in dhcp config

rick-yhchen1013:
Sync dhcp_dns_server with dhcp_dns_server, this makes dhcp server
issue correct DNS ip to clients.

The client always get 192.168.1.1 as DNS ip and the test cases
32-43 and 92-103 will fail once user choose another ip address
for OCE test host rather than 192.168.1.1.

Please check below discussion on forum:
http://lists.opencompute.org/pipermail/opencompute-onie/2017-June/001440.html


cbrune:
Thanks for this.  Pushed to master:

- 66319cc oce: Set correct dns server ip in dhcp config

PR #575, opened 2017-08-03
Linux 4.9.40

cbrune:
Minor bump of default kernel to 4.9.40

Update a couple machines that were using 4.9.35.


cbrune:
Pushed to master:
- deb2f83 quanta_ly4r: track default kernel version 4.9.40
- b184e5e quanta_common_rglbmc: track default kernel version 4.9.40
- e201e60 kernel: bump default kernel from 4.9.35 to 4.9.40

PR #577, opened 2017-08-08
Boot message cleanup for Lenovo NE1032, NE1032T, NE1072T platforms

psarker:
This is to cleanup some boot up messages on Lenovo NE1032, NE1032T, NE1072T platforms.


cbrune:
Looks OK, pushed to master:

- f19bb01 lenovo-NE10xx: Clean up U-Boot bootup messages

PR #579, opened 2017-08-16
kvm_x86_64: add disk based onie-syseeprom support

cbrune:
Our busybox patches long ago added a hack to support "EEPROM contents"
on a unused portion of the primary ONIE disk device.  This is a hack
and should not be used for real production hardware.  For the virtual
machine it seems fine.

For the VM this is nice, as it allows testing onie-syseeprom code on a
VM.

This patch enables the busybox config CONFIG_SYS_EEPROM_DISK_DEVICE
parameter for the kvm_x86_64 ONIE machine.

Additionally this patch adds a small "EEPROM init" script to the image
that helps initialize a blank virtual machine disk based EEPROM.

The INSTALL documentation is also updated with info about the disk
based EEPROM.


cbrune:
Pushed to master.

Issue #572, opened 2017-08-02
During ONIE update/install stop setting ONIE  EEPROM TLV 0x28 (platform name)


cbrune:
Referring to mailing list:
http://lists.opencompute.org/pipermail/opencompute-onie/2017-August/001491.html

It is no longer required to set EEPROM TLV 0x28 during ONIE install/update.  That is a legacy hold over that is no longer needed.

PR #580, opened 2017-08-16
During ONIE update, do not modify EEPROM TLV 0x28

cbrune:
A historical artifact in the ONIE installer is to update the platform
name (TLV 0x28) in the system EEPROM as part of the upgrade.

That operation is not needed as the platform name should be static
across upgrades.

This patch removes the code from the ONIE installer that was updating
EEPROM TLV 0x28 during ONIE upgrades.

Closes: #572


cbrune:
Pushed to master.

PR #586, opened 2017-08-18
limit platform install files to installer only

cbrune:
The build-config/scripts/onie-mk-installer.sh script is used for
making both the ONIE installer and any ONIE firmware installers.

The type of installer to build is specified in $update_type.

For firmware installers we do not want to include the
$machine_dir/installer/install-platform file, which is only intended
to be used by the ONIE installer.  The firmware installers do not need
this and it can cause problems.


cbrune:
pushed to master:
- 062bea3 limit platform install files to installer only

PR #587, opened 2017-08-18
Rename /etc/rcS.d scripts - increase spacing

cbrune:
This patch renames the rcS.d scripts, retaining the same order, but
increasing the spacing between stages.  This does not change any
behavior.

This allows for inserting new stages in between some of the original
stages.


cbrune:
pushed to master:
- fec3ba0 Rename /etc/rcS.d scripts - increase spacing

PR #592, opened 2017-08-22
Supporting common CPU modules

cbrune:
This patch series adds support and infrastructure for platforms that use a common CPU module.

Also included is an example of using the infrastructure byway of the kvm_x86_64 virtual machine.

### DIFF
--- a/installer/install.sh
+++ b/installer/install.sh
@@ -7,11 +7,11 @@
 
 cd $(dirname $0)
 
-[ -r ./machine.conf ] || {
-    echo "ERROR: machine.conf file is missing."
+[ -r ./machine-build.conf ] || {
+    echo "ERROR: machine-build.conf file is missing."
     exit 1
 }
-. ./machine.conf
+. ./machine-build.conf
 
 # Default implementation is no additional args
 parse_arg_arch()
@@ -43,6 +43,10 @@ true ${onie_machine_rev=0}
 # for backward compatibility if running onie_conf_version is empty
 # assume it is 0.
 true ${onie_config_version=0}
+
+# for backward compatibility, if running onie_build_machine is empty
+# assign the value from onie_machine.
+onie_build_machine=${onie_build_machine:-$onie_machine}
 
 args="hivfqx${args_arch}"
 
@@ -99,7 +103,7 @@ while getopts "$args" a ; do
             ;;
         i)
             # Dump the image information
-            cat ./machine.conf
+            cat ./machine-build.conf
             exit 0
             ;;
         *)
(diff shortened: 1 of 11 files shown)


rootconf/default/etc/machine.conf
```diff
@@ -0,0 +1,16 @@
...
+# Source build-time machine configuration
+. /etc/machine-build.conf
+
+# Source run-time machine configuration if available
+[ -r /etc/machine-live.conf ] && . /etc/machine-live.conf
+
+# Use onie_machine if set, otherwise use build_machine
+onie_machine=${onie_machine:-$onie_build_machine}
+
+onie_platform="${onie_arch}-${onie_machine}-${onie_machine_rev}"
```
david56: I believe it is missing 'r' before `${onie_machine_rev}` .  Please see the original definition of `onie_platform`:

    RUNTIME_ONIE_PLATFORM     ?= $(ARCH)-$(RUNTIME_ONIE_MACHINE)-r$(MACHINE_REV)

installer/install.sh
```diff
@@ -156,15 +160,15 @@ check_machine_image
 
 if [ "$fail" = "yes" ] && [ "$force" = "no" ] ; then
     echo "ERROR:$update_label: Machine mismatch"
-    echo "Running machine     : ${onie_arch}-${onie_machine}-${onie_machine_rev}"
-    echo "Update Image machine: ${image_arch}-${image_machine}-${image_machine_rev}"
+    echo "Running machine     : ${onie_arch}-${onie_build_machine}-${onie_machine_rev}"
+    echo "Update Image machine: ${image_arch}-${image_build_machine}-${image_machine_rev}"
```
david56: Should we need to add 'r' before `${onie_machine_rev}` and `${image_machine_rev}` here?

david56: It looks that `Running machine` and `Update Image machine` need to be changed to `Running platform` and `Update Image platform` for meeting ONIE's platform definition.

rootconf/default/etc/machine.conf
cbrune: Right.  Great catch. Thanks.

installer/install.sh
cbrune: Indeed, it looks like we've missed that `r` for a long time.

Also, agree that the text should say "platform" instead of "machine".

I'll fix both of these.

cbrune:
Thanks for the review everyone, pushed to master:
- 3cbcb9e Identifying machines with common CPU modules
- b92c261 kvm_x86_64: test of runtime ONIE config generation

PR #598, opened 2017-08-31
revise uefi_boot_first to print error messages precisely

david56:


cbrune:
These commits look helpful.  I like the extra error checking and the more informative message.

Pushed to master:
- e0fe278 uefi_boot_first: correct error message of updating BootOrder
- 60cfc1b uefi_boot_first: check boot_num before updating BootOrder

PR #600, opened 2017-09-06
Merge the kernel patches from different Alpha machines.

Mingxikao:
Merge the kernel patches from different machines.
The "kernel-3-14-drivers-base-dd.patch" is for Linux Kernel 3.14.xx.
The "kernel-3-2-drivers-base-dd.patch" is for Linux Kernel 3.2.xx. 
These two patches are to skip the error messages during probing the specified devices.


cbrune:
Nice clean up, pushed to master:

- 80e2b8f Merge the kernel patches from different Alpha machines.

PR #601, opened 2017-09-07
Add two Alpha Networks machines.

Mingxikao:
Add two Alpha Networks machines : SNH60A0-640F and SNH60B0-640F
These two machines both have 64 x 100 Gigabit Ethernet QSFP28 ports with Broadcom BCM56970 switch ASIC.
The SNH60A0-640F is using Intel Rangeley C2558 Quad Cores/2.4GHz and the default console device is ttyS1.
The SNH60B0-640F is using Intel Broadwell DE D-1518 Quad cores/2.3GHz and the default console device is ttyS0.


cbrune:
This looks fine, but the commit message needs to be better.  Two specific things:

1. please make this two different commits.  one commit for each machine.  do not combine multiple machines into a single commit.

2. for each commit please include an informative commit message.  The current commit message is "Add two Alpha Networks machines", which tells us nothing.

You did make a nice, informative pull request message -- that level of detail needs to be in the commit messages.

Mingxikao:
Thanks, I will close this request and refine my commitments.

PR #613, opened 2017-10-02
changes for ONIE build date

david56:


cbrune:
Looks good, pushed to master:
- 56b7b83 make onie_build_date in U-Boot and machine.conf be consistent
- 443ada9 updater: show build date during installing ONIE

PR #614, opened 2017-10-02
Accton P9572X: add to support ONIE

david56:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on Broadcom
Tomahawk switch ASIC.  CPU module is Intel Xeon D-1557.


cbrune:
Looks good.  Thanks for the contribution.

Pushed to master:
- e50d692 Accton P9572X: add to support ONIE

PR #616, opened 2017-10-11
Accton ASXvOLT16: add to support ONIE

david56:
The machine has 16x10GbE XFP and 4x40GbE/100GbE QSFP28 ports


cbrune:
Thanks for the new machine!  Pushed to master:

- 82cbb36 Accton ASXvOLT16: add to support ONIE

PR #618, opened 2017-10-11
Accton AS5916_54X: change SYS_EEPROM slave address

david56:


cbrune:
Pushed to master:
- dd949e4 Accton AS5916_54X: change SYS_EEPROM slave address

PR #620, opened 2017-10-13
Accton AS7816_64X: add to support ONIE

david56:
The machine has 64x100GbE QSFP28 ports on Broadcom Tomahawk2 switch


cbrune:
I think this is the first Tomahawk2 based switch in ONIE.  Nice!

Pushed to master:

- bfc8312 Accton AS7816_64X: add to support ONIE

PR #623, opened 2017-10-25
2017.08.01br add ONIE of AG5648.

delta-JohnsonLu:
The machine have 48x25GbE SFP28 and 6x100GbE QSFP28 on broadcom BCM56963 Tomahawk switch ASIC. CPU module is Intel Rangeley C2538 and with SSD 8G.

PR #624, opened 2017-10-26
Ag5648

delta-JohnsonLu:
The machine have 48x25GbE SFP28 and 6x100GbE QSFP28 on broadcom BCM56963 Tomahawk switch ASIC. CPU module is Intel Rangeley C2538 and with SSD 8G.


cbrune:
This is progress over #623 in that the PR is made against the master branch.

It still is a bit of a mess though -- there are 15 commits in this PR.  There should only be 1 commit, the commit that adds the AG5648.

Also PRs need to start with the master branch, it looks like this started with the 2017.08.01 branch.

In a PR for the master branch we should never commit the build-config/conf/onie-release file.

build-config/conf/onie-release
```diff
@@ -0,0 +1 @@
+2017.08.01
```
cbrune: A PR for the master branch will never include this file.  This file in only present in release branches.

PR #626, opened 2017-10-30
Delta AG5648

delta-JohnsonLu:
The machine have 48x25GbE SFP28 and 6x100GbE QSFP28 on broadcom BCM56963 Tomahawk switch ASIC. CPU module is Intel Rangeley C2538 and with SSD 8G.


cbrune:
This looks great.  Thanks for getting this patchset into shape.

Pushed to master:
- 3bdb868 delta_ag5648:  Add new platform

PR #627, opened 2017-10-31
Update variables in install-platform revised in 3cbcb9e

david56:
In commit 3cbcb9e, it renamed `onie_machine` and `image_machine` to `onie_build_machine` and `image_build_machine`.  And `image_machine` never exists since the commit.  We need to update these variables to the new ones.

The modification has been tested only in Accton AS7712_32X because it is the same in other machines.


cbrune:
Thank you for fixing these!  Pushed all the commits to master.

PR #628, opened 2017-11-01
Update variables in install-platform revised in 3cbcb9e 

david56:
Sorry for my fault.  Some variables were not discovered and replaced.  I checked them again carefully.  All `onie_machine` and `image_machine` in install-platforms are replaced to `onie_build_machine` and `image_build_machine` now.


cbrune:
Thanks for the fix.  Pushed all to master.

PR #629, opened 2017-11-01
Delta agc7648a

delta-JohnsonLu:
The machine have 48x25GbE SFP28 and 6x100GbE QSFP28 on broadcom BCM88375 Qumran switch ASIC.
CPU module is Intel Rangeley C2538 and with SSD 32G.


cbrune:
This all looks really good.  Thanks for the new platform!

Pushed to master:
- 9b84b74 delta_agc7648a: Add new platform

PR #632, opened 2017-11-09
Add gen-config-platform for Quanta common CPU board

jonathantsai-qci:
Please do validation and merge code. Thanks a lot.


jonathantsai-qci:
Need to do some modification and patch it later.

PR #636, opened 2017-11-13
Delta ag9032v2

delta-JohnsonLu:
The machine have 32x100GbE QSFP28 on broadcom BCM56870 Trident 3 switch ASIC.
CPU module is Intel Denverton C3538 and with SSD 64G.


machine/delta/delta_ag9032v2/machine.make
```diff
@@ -0,0 +1,46 @@
...
+# Specify uClibc version
+#UCLIBC_VERSION = 0.9.32.1
+
+UEFI_ENABLE = no
+
+#-------------------------------------------------------------------------------
+#
+# Local Variables:
+# mode: makefile-gmake
+EXTRA_CMDLINE_LINUX = libata.force=noncq panic=10 hpet=disable
```
cbrune: Why are these 3 variables set?  What problems do they solve?  I have never seen these set in other ONIE machines.

delta-JohnsonLu: Sorry, these 3 variables set is use in debug, I forgot to remove it.

Issue #487, opened 2017-02-23
Coreboot support is broken


cbrune:
I looks like some recent GRUB refactoring has broken the build for coreboot platforms.

See this thread:

http://lists.opencompute.org/pipermail/opencompute-onie/2017-February/001355.html

```
I'm having trouble compiling the Intel Rangeley recovery iso.

I keep getting an error message saying ERROR: Does not look like valid 
GRUB i386-pc library directory: 
onie-master/build/intel_coreboot_rangeley-r0/dev-sysroot/usr/lib/grub/i386-pc.
```

jf6316:
I have same error too. Please help to fix this bug.

reju85:
I also same error too. Please help to fix this bug.

cbrune:
Hi @jf6316  and @reju85  --

Thanks for pointing out the desire to have this fixed.

Can either of you help fix this problem?  We need someone with access to a coreboot setup to fix it.  I do not have one.

jf6316:
I have go back git branch "2017.02br" it's work fine thank you

donnydavis:
I built on Centos 7.4 and had to go back to 2016.11br

cbrune:
I made a PR that fixes the build.

However, I do not have a coreboot system, so have no way of verifying if the resulting image actually works or not.  @donnydavis, @reju85 or @jf6316  can you test it?

donnydavis:
I only have a generic C2758 machine to test with. I'm not sure if that is helpful

PR #646, opened 2017-11-29
Fix build of coreboot platforms

cbrune:
The build of coreboot platforms has not been working for some time, see #487 .

This patch series fixes a few things for coreboot in the areas of GRUB and how the .ISO installer image is built.


cbrune:
pushed to master.

PR #649, opened 2017-12-04
Improve checking for missing libraries at build time

cbrune:
The method for validating that the sysroot contains all the necessary
libraries from the dev-sysroot has a small flaw.  The check only
validates that the file names in the two trees are identical.

In the case of libraries, sometimes the trees would differ because one
file was a symlink, while the other file was a real file.

This patch moves the validation checking out into a stand alone
script.  The logic was becoming too complex to implement inside of a
Makefile.

The new validation algorithm first checks that all the "real files"
are identical and then checks that all the "symlinks" are identical.


cbrune:
pushed to master

PR #650, opened 2017-12-04
Use pkg-config correctly

cbrune:
The ONIE build system was not correctly using automake and pkg-config in all cases.  These patches fix that.


cbrune:
pushed to master

Issue #648, opened 2017-12-04
TFTP image discover attempts "DHCP Server IP" (option 54) before TFTP server name (DHCP option 66)


cbrune:
ONIE commit b760eded3b54c9 inadvertently changed the TFTP server search order:
```
  commit b760eded3b54c9104f5c80c1bed01794546a7a9d
  Author: Curt Brune <curt@cumulusnetworks.com>
  Date:   Mon Apr 24 13:55:59 2017 -0700

    Add $onie_disco_serverid and $onie_server_name to TFTP waterfall
```

In the list of servers to try, that commit inserted "DHCP Server IP" (DHCP option 54) *before* TFTP server name (DHCP option 66).  Since option 54 is pretty much universally set by the DHCP server, we will always hit that before we get to TFTP server name (DHCP option 66).                                                           

1. This is an ONIE bug.  "DHCP Server IP" should definitely come *after* "DHCP Server Name".                                                                                                   

2. This impacts OCE tests 12 and 72.

PR #651, opened 2017-12-04
image-discovery: update TFTP server search order

cbrune:
ONIE commit b760ede inadvertently changed the TFTP server search
order:

  commit b760eded3b54c9104f5c80c1bed01794546a7a9d
  Date:   Mon Apr 24 13:55:59 2017 -0700

    Add $onie_disco_serverid and $onie_server_name to TFTP waterfall

In the list of servers to try, that commit inserted "DHCP Server
IP" (DHCP option 54) before TFTP server name (DHCP option 66). Since
option 54 is pretty much universally set by the DHCP server, we will
always hit that before we get to TFTP server name (DHCP option 66).

The fix here is to put "DHCP Server IP" (DHCP option 54) *last* in the
search order.  Compared to "TFTP server name" and "TFTP server IP",
the IP address of the DHCP server is the least specified.

Closes: #648

### DIFF
--- a/rootconf/default/bin/exec_installer
+++ b/rootconf/default/bin/exec_installer
@@ -314,13 +314,13 @@ tftp_download()
 $onie_server_name
 # BOOTP next-server IP
 $onie_disco_siaddr
-# DHCP server IP (DHCP opt 54)
-$onie_disco_serverid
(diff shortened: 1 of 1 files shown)


cbrune:
pushed to master

PR #654, opened 2017-12-06
Platform support for the DELL EMC S4100

sharma-lakesh:


cbrune:
Looks good.

Thanks for all your efforts.

Pushed to master:
- 640ebe3 Platform support for the DELL EMC S4100

PR #657, opened 2017-12-12
Add build time platform information to GRUB messages

cbrune:
Commit 3cbcb9ef32c changed the way some of the $onie_XXX variables are
created and used, in particular $onie_platform is now computed are
runtime:

commit 3cbcb9ef32cdeffb2e036565271cbd010cf1f005
Date:   Wed Aug 16 16:02:29 2017 -0700

    Identifying machines with common CPU modules

At runtime we now synthesize $onie_platform from runtime information
if available, otherwise we use the build time information.

For GRUB, this had the effect of making $onie_platform blank and GRUB
would print a blank platform string.

For GRUB we have no runtime info, so the best we can do is use the
build time info.

This patch echos the "build time" platform information when GRUB loads
and boots a kernel.


build-config/make/images.make
```diff
@@ -272,6 +272,7 @@ $(SYSROOT_COMPLETE_STAMP): $(SYSROOT_CHECK_STAMP)
 	$(Q) echo "onie_build_machine=$(ONIE_BUILD_MACHINE)" >> $(MACHINE_CONF)
 	$(Q) echo "onie_machine_rev=$(MACHINE_REV)" >> $(MACHINE_CONF)
 	$(Q) echo "onie_arch=$(ARCH)" >> $(MACHINE_CONF)
+	$(Q) echo "onie_build_platform=$(ARCH)-$(ONIE_BUILD_MACHINE)-$(MACHINE_REV)" >> $(MACHINE_CONF)
```
david56: The letter `r` is lost in the string.  i.e. it would be `$(ARCH)-$(ONIE_BUILD_MACHINE)-r$(MACHINE_REV)`.   😃

cbrune: ah ha!  good catch. i will update that.

david56:
Thank you very much.  Everything looks good to me now.

cbrune:
Pushed to master.

PR #658, opened 2017-12-12
intel_coreboot_rangeley: add demo OS files

cbrune:
This platform never had the demo/ sub-directory containing demo OS
specific files.  Without these files we cannot build the demo OS for
this platform.

Building the demo OS is part of the ONIE certification process.

This patch adds a minimal demo/ sub-directory so that we can build the
demo OS for this platform.


cbrune:
pushed to master:
- 5de5768 intel_coreboot_rangeley: add demo OS files

PR #659, opened 2017-12-13
Add a new Lenovo platform NE2572

psarker:
A new Lenovo x86 platform NE2572


machine/lenovo/lenovo_ne2572/i2ctools/001-i2ctools-sys-eeprom-16bits-address.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/sys_eeprom/onie_tlvinfo.c b/sys_eeprom/onie_tlvinfo.c
```
cbrune: For a couple of years now, the ONIE project does not accept the "i2ctools" implementation of onie-syseeprom for new platforms.  We have standardized on the busybox implementation.  Discussed here:

Looking around I see I mistakenly accepted this for your `lenovo_ne10032` platform in July of this year.  That was a mistake on my part.  Missed it.

The busybox implementation is much better, more configurable without the need for platform specific patches to change things like the delay timing.  See this example:
https://github.com/opencomputeproject/onie/blob/master/machine/accton/accton_as5812_54x/busybox/conf/config

Also see these commits that added the busybox implementation:
fc76004dca594a5f4f12a7905474f77c67e5effb
481b975ac4c5e8dbec255a2ffb1af3e383475d3a

machine/lenovo/lenovo_ne2572/machine.make
```diff
@@ -0,0 +1,51 @@
...
+# Set Linux kernel version
+LINUX_VERSION		= 4.1
+LINUX_MINOR_VERSION	= 38
+
+# Specify uClibc version
+#UCLIBC_VERSION = 0.9.32.1
+GCC_VERSION = 4.9.2
+
+#Extra kernel command line
+EXTRA_CMDLINE_LINUX ?= quiet
```
cbrune: quiet is already added to the kernel command line

machine/lenovo/lenovo_ne2572/machine.make
```diff
@@ -0,0 +1,51 @@
...
+VENDOR_VERSION = .0.3
+
+# Vendor ID -- IANA Private Enterprise Number:
+# http://www.iana.org/assignments/enterprise-numbers
+# Lenovo 
+VENDOR_ID = 19046
+
+# Enable the i2ctools and the onie-syseeprom command for this platform
+I2CTOOLS_ENABLE = yes
+I2CTOOLS_SYSEEPROM = no
```
cbrune: `I2CTOOLS_SYSEEPROM = no` is the default already.

Funny, with this set to `no` the i2ctools patches above are not used at all.

Also you have defined the busybox implementation of onie-syseeprom.

I think the thing to do here is completely remove the i2ctools directory from this patch.

cbrune: Same comments for the previous `lenovo_ne10032` platform.  Looks like you can get rid of the `i2ctools` subdirectory.  It is not being used.

machine/lenovo/lenovo_ne2572/i2ctools/001-i2ctools-sys-eeprom-16bits-address.patch
psarker: Yea.. it was my mistake.. let me completely remove i2ctool directory

machine/lenovo/lenovo_ne2572/machine.make
psarker: I just have checked it and im still seeing when demo OS boots up it prints kernel msgs in console. So, I will keep this flag ON.

machine/lenovo/lenovo_ne2572/machine.make
psarker: I will comment out I2CTOOLS_SYSEEPROM = no. thank you. I will also delete i2ctools directory

psarker:
Hi Curt,
I have created a new submit #663 to address your comments. I just kept "EXTRA_CMDLINE_LINUX ?= quiet" as it is in machine.make.

Thanks,
Pinaki

machine/lenovo/lenovo_ne2572/machine.make
cbrune: The demo OS does not use `quiet`, you are right.  That was on purpose to see that things were working.

When ONIE boots, it uses `quiet`.

If you want demo OS to also be quiet, then this is fine.

cbrune:
this is handled by #663.

PR #660, opened 2017-12-13
Use dmidecode for system information

cbrune:
For GRUB based architectures (x86_64 and arm64) fallback to using DMI/SMBIOS information when a system EEPROM is unavailable.


cbrune:
Pushed to master.

PR #665, opened 2017-12-15
Add to support Accton AS7816_64X r1 machine

david56:


cbrune:
It all looks good to me. I like the new GRUB_TIMEOUT addition.

Pushed to master:
- b2d5162 Accton AS7816_64X: add to support r1 ONIE
- cc7a80a grub: be able to set timeout by platform

PR #666, opened 2017-12-15
Accton AS5916_54XM: add to support ONIE

david56:


cbrune:
Looks good.  Thanks for the new machine.

Pushed to master:
- 30a7079 Accton AS5916_54XM: add to support ONIE

PR #668, opened 2017-12-19
Update ONIE support for Mellanox x86 based systems.

MellanoxBSP:
This pull request contains support of Mellanox X86 systems for current "master" ONIE.
Main changes:
Build with new linux kernel 4.9.57 and new gcc version;
Use busybox onie-syseeprom instead i2c-tools utility;
Update dynamic platform configuration;
Clean unnecessary code.
Tested on MSN2700, MSN2410 and MSN2100 systems
All other Mellanox x86 based systems should be supported as well.

Signed-off-by: Michael Shych <michaelsh@mellanox.com>


machine/mellanox/mlnx_x86/rootconf/sysroot-init/mellanox-machine-model.sh
```diff
@@ -1,47 +1,10 @@
 #!/bin/sh
 
-PATH=/usr/bin:/usr/sbin:/bin:/sbin
+cat /etc/machine-build.conf > /etc/machine.conf
```
cbrune: Please do not overwrite `/etc/machine.conf`.  Your `gen-config-platform` should generate everything you need into `/etc/machine-live.conf`.  Then the base `/etc/machine.conf` will take care of the rest.

I think you should just remove this file : `mellanox-machine-model.sh`.  It is not needed anymore with the new common infrastructure.

machine/mellanox/mlnx_x86/i2ctools/series
```diff
@@ -1,4 +1,2 @@
-0001-Increasing-error-flow-verbosity-eeprom_read_byte.patch
-0002-fixing-DEBUG-compilation.patch
-0003-Adding-Mellanox-platform-support.patch
-0004-write_sys_eeprom-taking-eeprom-open-out-of-the-eepro.patch
+0001-Mellanox-i2c-platform-support.patch
+0002-Mellanox-i2c-eeprom_force.patch
```
cbrune: Why do you need to patch i2ctools at all?  Now you are using the busybox implementation of `onie-syseeprom` I would expect all the i2ctools patches to go away.

cbrune:
This looks like a very nice clean up.  Just a couple of minor things noted in the patches.

I like seeing that the machine definition is simpler now.  It looks a lot less complicated.

machine/mellanox/mlnx_x86/rootconf/sysroot-init/mellanox-machine-model.sh
MellanoxBSP: I checked this ONIE with supported NOS's for our systems: Cumulus 3.4.3, Sonic, Mlnx-OS and ONL.
There is a problem with ONL. It looks that ONL installer reads directly /etc/machine.conf
and don't accept new format.
"...
ImportError: No module named "${onie_arch}_${onie_machine}_r${onie_machine_rev}"                              
Install failed. 
"
More than that, it fails also in simply concatenated machine-build.conf and machine-live.conf.
"
...
ValueError: invalid line in /etc/machine.conf: # Runtime ONIE Machine                    
Install failed. 
"
That's why I remove this comment in mellanox-machine-model.sh
Unfortunately we already have systems with ONL. I tested with latest ONL build.
Is it possible that also other NOS rely on old machine.conf format?

machine/mellanox/mlnx_x86/i2ctools/series
MellanoxBSP: I made just in case support also for onie-syseeprom tool and tested it.
I'll remove completely unnecessary i2ctools directory under Mellanox machine.

MellanoxBSP:
Last commit 214ca6c is not connected to Update Mellanox systems pull request. 
I can send it if needed as separate pull request.
This patch enables print BIOS logs (if they supplied) by dmidecode -t15

machine/mellanox/mlnx_x86/rootconf/sysroot-init/mellanox-machine-model.sh
cbrune: NOSes should not try to read/parse `/etc/machine.conf` directly.  That file is not an interface.

The supported method is to use `onie-sysinfo` to obtain system information.  See:
https://opencomputeproject.github.io/onie/cli/index.html#onie-sysinfo

If a NOS cannot get what it needs from `onie-sysinfo` then please file a bug and let's fix it.

I understand for the legacy of ONL you may be stuck with this hybrid.  I would encourage ONL to change.

patches/dmidecode/dmidecode_bios_logs_support.patch
```diff
@@ -0,0 +1,106 @@
+--- a/dmidecode.c	2017-12-20 15:38:13.901624013 +0000
```
cbrune: Please open a separate pull request for this unrelated patch

cbrune:
Let's be pragmatic and keep the hybrid approach for `/etc/machine.conf` for now, in order to support the ONL folks.  However, let's also notify them that the plan is to move to the standard `onie-sysinfo` based method in the 2018.05 release.

I squashed a57958b and 05b240c into a single commit and pushed the result:
- bc25ef2 Update ONIE support for Mellanox x86 based systems.

I am not pushing the dmidecode change at this time.  Open a separate PR for that.

jnealtowns:
@cbrune We are working to update the ONL installer to use ``onie-sysinfo`` now.

PR #669, opened 2017-12-20
Dmidecode additions that enables BIOS logs print.

mshych:
Common code based on SMBIOS specification that enables print of BIOS logs. dmidecode -t15


cbrune:
This pull request is a bit messy.  You should make this PR from a branch that is specific to the problem being fixed.  We should not see any of the older commits here.

A few comments on the patches themselves:
- please squash the commits down to a single commit for the PR
- please add a bit more meat to the commit message, explaining why it is necessary, pointing to relevant specs, etc.

mshych:
New pull request will be done instead this one.

PR #670, opened 2017-12-21
machine.conf: generated automatically during ONIE booting

david56:
In commit 3cbcb9e, it introduced a new machanism to identify machines
with common CPU module.  The machine.conf was changed to source
other configuration files to calculate the real `onie_machine` and
`onie_platform` at ONIE runtime.

Some NOSs such as SONiC would copy machine.conf to its partition and use
it at its runtime.  The new machine.conf needs other files and has to
put in /etc directory or it cannot be sourced correctly.

Besides SONiC other NOSs also has similar issues.

The patch intends to expand all variables from machine-build.conf and
machine-live.conf to machine.conf directly.  That would be like the
previous one:

    ONIE:/ # cat /etc/machine.conf
    # /etc/machine.conf for onie

    #  Copyright (C) 2017 Curt Brune <curt@cumulusnetworks.com>
    #  Copyright (C) 2017 david_yang <david_yang@accton.com>
    #
    #  SPDX-License-Identifier:     GPL-2.0

    onie_version=2017.12.12.20-dirty
    onie_vendor_id=259
    onie_build_machine=accton_as7712_32x
    onie_machine_rev=0
    onie_arch=x86_64
    onie_build_platform=x86_64-accton_as7712_32x-r0
    onie_config_version=1
    onie_build_date="2017-12-20T13:43+0800"
    onie_partition_type=gpt
    onie_kernel_version=4.1.38
    onie_firmware=auto
    onie_switch_asic=bcm
    onie_skip_ethmgmt_macs=no
    onie_grub_image_name=grubx64.efi
    onie_machine=accton_as7712_32x
    onie_platform="x86_64-accton_as7712_32x-r0"
    ONIE:/ #

The patch has been tested on Accton AS7712_32X with SONiC environment.


david56:
I'm thinking whether we need to keep a copy machine.conf in the initrd image.  It seems to be unnecessary to keep it in current design.

cbrune:
I am coming to understand several OSes are reading/including /etc/machine.conf directly.  I think that is a really bad idea and people should be using `onie-sysinfo` to determine things about the system.

However, I do not want to intentionally break things that are working either.  I encourage people to start using `onie-sysinfo` instead of sourcing this file directly.

Let's add this patch to support the legacy for now.  This patch might impact #668, which went in yesterday.  That PR also grandfathered some legacy /etc/machine.conf behavior.  Please look at that.

Regarding this patch I have a few stylistic suggestions, but the logic looks good.

rootconf/default/etc/init.d/gen-config.sh
```diff
@@ -17,9 +18,45 @@ gen_live_config()
...
 
+gen_machine_config()
+{
+    local build_conf=/etc/machine-build.conf
+    local live_conf=/etc/machine-live.conf
+    local machine_conf=/etc/machine.conf
+
+    gen_live_config > $live_conf
+
+    (cat <<EOF
```
cbrune: I don't like the subshell approach when using <<EOF.  I think this syntax is cleaner and easier to understand:
```
cat<<EOF > $machine_conf
# /etc/machine.conf for onie
# ...
EOF
```
You can also append by using `cat<<EOF >> $machine_conf`.

Please use that syntax here and below.

rootconf/default/etc/machine.conf
```diff
@@ -1,16 +0,0 @@
-# /etc/machine.conf for onie
```
cbrune: I think it is OK to remove this file as you have done here.

david56:
Thanks for your explanations.  I have updated the patch.

cbrune:
Thanks for doing this.  Pushed to master:

- fe5b096 machine.conf: generated automatically during ONIE booting

PR #673, opened 2018-01-05
gen-config: remove double quotes from onie_platform

david56:
Some old version of NOSes does not take double quotes into consideration
so that onie_platform extracted from recent machine.conf will have
double quotes that would be taken as an unsupported platform.

The original onie_platform was not surrounded by double quotes.  The
patch changes it back to the original format.


cbrune:
OK, pushed to master:

- 33768d0 gen-config: remove double quotes from onie_platform

PR #674, opened 2018-01-11
im_n29xx_t40n: Handle uninstall of MSDOS partitions.

oleksandrivantsiv:
- Logical partitions should be removed before extended.
- Extended partitions should be removed without data erasing.

Signed-off-by: Oleksandr Ivantsiv <oleksandri@interfacemasters.com>
Signed-off-by: Vitaliy Ivanov <vitaliyi@interfacemasters.com>


cbrune:
I do not fully understand the reasoning, but I trust you IM folks know what you are doing with msdos partition tables.  I believe the im machines are the only ones using msdos partitions.  Pushed to master:

- 3ed8cd6 im_n29xx_t40n: Handle uninstall of MSDOS partitions.

vitalivanov:
@cbrune The reason is that with msdos you can have only 4 primary partitions.
If you need more you can create extended and logical partitions.
Added is the correct way to cleanup extended and logical partitions if you have any.

PR #676, opened 2018-01-12
Accton AS7726_32X: add to support ONIE

david56:
The machine has 32x100GbE QSFP ports on Broadcom Trident3 switch


cbrune:
Thanks for the new machine definition, pushed to master:

- fd2204a Accton AS7726_32X: add to support ONIE

PR #678, opened 2018-01-17
VENDOR_VERSION is updated on Lenovo G8272 and G8296

psarker:
VENDOR_VERSION is updated to .0.3 on G8272 and G8296 to bring all platforms in same version.


cbrune:
Pushed to master:

- 4cb1103 VENDOR_VERSION is updated on Lenovo G8272 and G8296

Issue #679, opened 2018-01-18
onie-fwpkg update records are missing


cbrune:
The onie-fwpkg command is supposed to display the ONIE update (and firmware update) history on the box.  However, it looks like only the most recent record is displayed.  It is supposed to keep all the history.

Digging a bit there is a bug in the installer/updater where it tries to preserve the update results directory:

https://github.com/opencomputeproject/onie/blob/master/installer/grub-arch/install-arch#L783-L786

Instead of keeping the old update folder in /mnt/onie-boot/onie/update, the old update folder is copied to
/mnt/onie-boot/onie/update/preserve-update.  And that happens recursively for each update.  Eventually you get deep paths like: /mnt/onie-boot/onie/update/preserve-update/preserve-update/preserve-update

PR #681, opened 2018-01-18
onie-fwpkg: correct preservation of update records

cbrune:
The onie-fwpkg command is supposed to display the ONIE update (and
firmware update) history on the box.

During ONIE updates the entire ONIE data directory, which includes the
location of the ONIE update records, is reformatted.  During this
process the update records are first preserved and then, after the
reformatting, restored to the previous location.

However, the restore logic has a bug.

This patch fixes the bug, so that the restoration of update records
works as intended.

Closes: #679

### DIFF
--- a/installer/grub-arch/install-arch
+++ b/installer/grub-arch/install-arch
@@ -731,8 +731,8 @@ install_image()
         exit 1
     }
 
-    # Create basic directories for installing and logging
-    mkdir -p $onie_root_dir $onie_update_dir $onie_update_results_dir $onie_update_pending_dir
+    # Create ONIE data directory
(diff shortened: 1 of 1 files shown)


cbrune:
pushed to master:
- cdeb90e onie-fwpkg: correct preservation of update records

lakesh-sharma:
Does it fix the issue of recursive preserve-update folder seen, which could eventually lead to disk space outage ?

/mnt/onie-update/update/preserve-update/preserve-update/

Do we need to remove preserve-update folder in purge option ?

cbrune:
Yes, the issue is fixed.

PR #685, opened 2018-01-25
WNC new platfrom osw1800

YaoTien:
Upload for WNC new platform, osw1800.


machine/wnc/wnc_osw1800/kernel/remove-i354-device-of-igb-driver.patch
```diff
@@ -0,0 +1,17 @@
...
++++ b/drivers/net/ethernet/intel/igb/igb_main.c
+@@ -77,9 +77,11 @@ static const struct e1000_info *igb_info_tbl[] = {
+ };
+ 
+ static DEFINE_PCI_DEVICE_TABLE(igb_pci_tbl) = {
++#if 0 //Remove I354 device of igb driver.
+ 	{ PCI_VDEVICE(INTEL, E1000_DEV_ID_I354_BACKPLANE_1GBPS) },
+ 	{ PCI_VDEVICE(INTEL, E1000_DEV_ID_I354_SGMII) },
+ 	{ PCI_VDEVICE(INTEL, E1000_DEV_ID_I354_BACKPLANE_2_5GBPS) },
++#endif
```
cbrune: Why do these devices need to be removed ?

If they really need to be removed, then just remove them instead of using `#if 0 ....  #endif`.

machine/wnc/wnc_osw1800/machine.make
```diff
@@ -0,0 +1,75 @@
...
+
+# The onie-syseeprom command in i2ctools is deprecated.  It is recommended to
+# use the one implemented in busybox instead.  The option intends to provide a
+# quick way to turn off the feature in i2ctools.  The command will be removed
+# from i2ctools in the future once all machines migrate their support of
+# sys_eeprom to busybox.
+#
+# The option is significant when I2CTOOLS_ENABLE is 'yes'
+#
+I2CTOOLS_SYSEEPROM = no
```
cbrune: you can remove this setting (and the comment above it) as this is default since July 2017.

machine/wnc/wnc_osw1800/kernel/remove-i354-device-of-igb-driver.patch
YaoTien: Our CPU board didn't support this device.
However I reserve it and test, all functions can work well.
So I think we can ignore this modification.

machine/wnc/wnc_osw1800/machine.make
YaoTien: I updated it. Thanks.

cbrune:
Looks alright!  I squashed the commits into a single commit and pushed it:

- 54ce1c3 WNC new platfrom osw1800

Issue #599, opened 2017-08-31
Update EFI libraries and utilities to more recent versions


cbrune:
The versions of `efivars` and `efibootmgr` in the ONIE project are kind of old.  These projects move pretty fast and a number of bug fixes have occurred in recent years.

ONIE versions:
efibootmgr: efibootmgr-0.12
efivars: efivar-0.20

It is time to resync with the upstream projects:
https://github.com/rhboot/efibootmgr
https://github.com/rhboot/efivar

PR #692, opened 2018-01-31
Upgrade efibootmgr and efivar

cbrune:
Address #599 

Upgrade efibootmgr and efivar to later versions.


cbrune:
Pushed to master

Issue #695, opened 2018-02-02
New efivar version 30 breaks with gcc 4.9.x


cbrune:
When building the updated efivar-30 with older gcc-4.9.2 it fails with this:

```
x86_64-onie-linux-uclibc-gcc-ar -cvqs libefivar.a dp.static.o dp-acpi.static.o dp-hw.static.o dp-media.static.o dp-message.static.o efivarfs.static.o error.static.o export.static.o guid.static.o guids.static.o guid-symbols.static.o lib.static.o vars.static.o
sorry - this program has been built without plugin support                                                                   
/work/onie/build/user/x86_64-g4.9.2-lnx4.1.38-uClibc-ng-1.0.22/efivar/efivar-30/Make.rules:8: recipe for target 'libefivar.a' failed
```

Seems related to newer efivars using LTO (Link Time Optimizer) support.  The fix looks to be to enable LTO support only for GCC version > 4.9.x.

See this https://github.com/sudar/Arduino-Makefile/issues/456

cbrune:
On a closer read it seems our 4.9.2 compiler should be OK with LTO.  The problem is in the configuration of our binutils for the 4.9.2 toolchain.

Need to enable "plugin" support for our binutils:
https://lists.openwrt.org/pipermail/openwrt-devel/2014-December/030223.html

Short term work around to not rock the boat too much for the 2018.02  release is just to no use LTO for GCC version < 5.  After 2018.02 is out the door will revisit to modify the binutils configuration.

PR #696, opened 2018-02-02
gcc-4.9.2: enable binutils plugin support

cbrune:
The newly updated efivar-30 component relies on LTO (link time
optimizer) support.  LTO support for the gcc-4.9.2 toolchain requires
that the underlying binutils (ar) be build with PLUGIN support.

This patch enables binutil plugin support for the gcc-4.9.2 x86_64
architecture only.  The x86_64 architecture is the only architecture
we have that uses both gcc-4.9.2 and efivar.

Closes: #695


cbrune:
pushed to master:

- 2c5f0a3 gcc-4.9.2: enable binutils plugin support

PR #700, opened 2018-02-15
Secure boot

cbrune:
Phase 1 of adding secure boot support to ONIE.

This adds support for building shim with a vendor's certificate embedded inside.  Also support for signing grub and the kernel with the vendor keys.

See the README.secureboot file in the kvm_x86_64 machine for more details.


cbrune:
Pushed to master

PR #702, opened 2018-02-28
grub: make setting debug kernel arguments easier

cbrune:
Since we call a GRUB shell function to generate the grub menu entries,
it is difficult to use the GRUB menu "edit" feature to append extra
args to the kernel.

For example, before this change, when editting the "Rescue" entry you
would see this:
```
            GNU GRUB  version 2.02

 +----------------------------------------------+
 |setparams 'ONIE: Rescue'                      |
 |                                              |
 |  onie_rescue                                 |
 +----------------------------------------------+
```
No clue how to add extra args with that definition.

This patch adds an empty "onie_debugargs=" setting to every GRUB menu
item.  To temporarily add additional kernel args, add values to
onie_debugargs.

After this change, the editting screen looks like
```
            GNU GRUB  version 2.02

 +----------------------------------------------+
 |setparams 'ONIE: Rescue'                      |
 |                                              |
 |  onie_debugargs=""                           |
 |  onie_rescue                                 |
```

### DIFF
--- a/installer/grub-arch/grub/grub-common.cfg
+++ b/installer/grub-arch/grub/grub-common.cfg
@@ -22,9 +22,6 @@ export onie_menu_embed
 
 set fallback="${onie_menu_rescue}"
 
-# default ONIE boot reason
-onie_boot_reason="install"
-
 onie_quiet="quiet"
 onie_linux="/onie/vmlinuz-${onie_kernel_version}-onie"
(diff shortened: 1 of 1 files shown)


cbrune:
pushed to master:
- 155c42e grub: make setting debug kernel arguments easier

PR #703, opened 2018-02-28
demo: support UEFI secure boot installs

cbrune:
The recent work for UEFI secure boot broke how the DEMO OS installer
works.  The demo installer needed similar changes as the regular ONIE
installer to accommodate shimx64.efi and the monolithic grubx64.efi
binary.

In particular, the monolithic grubx64.efi binary with a hard coded
--prefix option was problematic to re-use in the DEMO OS context.

The fix here is to embed a grub.cfg into the monolithic image that
looks for another grub.cfg located in the same directory as the
grubx64.efi binary.

Fixes: a7b03d5, 0e56d23, b68ca41

### DIFF
--- a/build-config/scripts/mk-grub-efi-image
+++ b/build-config/scripts/mk-grub-efi-image
@@ -91,6 +91,15 @@ GRUB_MODULES="
 	zfsinfo
 "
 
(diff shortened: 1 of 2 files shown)


cbrune:
pushed to master:

- 4996cd7 demo: support UEFI secure boot installs

Issue #690, opened 2018-01-29
GRUB menu: make installed NOS the top entry


cbrune:
`install mode` is currently the default GRUB menu item when entering the ONIE GRUB.

This can be a surprise for people who are manually navigating the GRUB menus from their NOS and all of a sudden find themselves in ONIE's GRUB and booting into install mode.  From there people get burned since install mode is sticky and now their NOS is gone.

For some discussion see: http://lists.opencompute.org/pipermail/opencompute-onie/2018-January/001617.html

The proposal here is to investigate the possibility of having a NOS be the *default* ONIE GRUB menu entry.  Specifically the behavior falls into two separate scenarios.

Scenario 1 -- Completely fresh system that only has ONIE installed.  No NOS yet.  In this case the default ONIE GRUB entry should be `install mode`.  Also install mode continues to remain sticky. The menu looks like:
```
   +----------------------------------+
   |*ONIE: Install OS                 |  <-- default
   | ONIE: Rescue                     |
   | ONIE: Uninstall OS               |
   | ONIE: Update ONIE                |
   | ONIE: Embed ONIE                 |
   | DIAG: Vendor Diagnostic          |
   +----------------------------------+
```

Scenario 2 -- A NOS is installed.  The user accidentally enters the ONIE GRUB menus.  In this case the default ONIE GRUB menu item would be for the NOS.  The menu would look like:
```
   +----------------------------------+
   |*NOS: Network Operating System    |  <-- default
   | ONIE: Install OS                 |
   | ONIE: Rescue                     |
   | ONIE: Uninstall OS               |
   | ONIE: Update ONIE                |
   | ONIE: Embed ONIE                 |
   | DIAG: Vendor Diagnostic          |
   +----------------------------------+
```
In this scenario if the end user stumbles into the ONIE GRUB menus they will simply boot back into their NOS.

Scenario 3 -- The user requests `install mode` on purpose, knowing that it is sticky.  As part of install mode, ONIE will make itself the new default.  The GRUB menu will now look like this (and remain this way until a NOS is installed):
```
   +----------------------------------+
   |*ONIE: Install OS                 |  <-- default
   | NOS: Network Operating System    |
   | ONIE: Rescue                     |
   | ONIE: Uninstall OS               |
   | ONIE: Update ONIE                |
   | ONIE: Embed ONIE                 |
   | DIAG: Vendor Diagnostic          |
   +----------------------------------+
```

[...]

PR #704, opened 2018-02-28
Grub fallback

cbrune:
Update the `onie-boot-mode` command to allow a NOS installer to persistently change default ONIE boot mode.

The intended use case is a NOS installer would set the fallback to "rescue" at the end of install a NOS.  That way if a user stumbles into the ONIE GRUB menu by accident the default entry will be "rescue", instead of the more damaging "install".

### DIFF
--- a/demo/installer/grub-arch/install.sh
+++ b/demo/installer/grub-arch/install.sh
@@ -427,6 +427,9 @@ EOF
 
 fi
 
+# Set the ONIE fallback to rescue mode
+onie-boot-mode -f -o rescue
+
 # clean up
 umount $demo_mnt || {
     echo "Error: Problems umounting $demo_mnt"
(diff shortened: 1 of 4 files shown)


cbrune:
@lakesh-sharma and @david56 -- do you think this is a good approach for #690 ?  It does not put the NOS installer at the top of the list, but it does make "RESCUE" mode the default entry when a user accidentally stumbles into the ONIE grub menus.

lakesh-sharma:
Looks good. The NOS installer needs to specifically call 'onie-boot-mode' to set this option or is it default case with these changes and would remain so, until some one selects/set any other onie mode.

cbrune:
As it is currently, the NOS installer has to actively set RESCUE as the fallback.  See the DEMO NOS commit 58d83c13dc62b.

I think we want to leave the decision about whether or not to set RESCUE as the fallback in the hands of the NOS installer.  It is hard for ONIE to guess what the NOS wants to do, so better to leave it up to the NOS installer.

cbrune:
Closing this PR as a better approach occurred to me.  Stay tuned...

PR #717, opened 2018-03-20
Add Alpha Networks SNH61D2-320F: 32*QSFP28+2*SFPP/Intel Denverton C3958.

YuChang9565:
SNH61D2-320F have 32x100 Gigabit Ethernet QSFP28 ports and 2x10 Gigabit Ethernet SFPP ports with Broadcom BCM56870(Trident3) switch ASIC. The CPU is Intel Denverton C3958 16-Core/2GHz and the default console device is ttyS0.


machine/alphanetworks/alphanetworks_snh61d2_320f/kernel/config
```diff
@@ -0,0 +1,9 @@
+CONFIG_IGB=y
+CONFIG_I2C_I801=y
+CONFIG_X86_X2APIC=y
+CONFIG_USB_XHCI_HCD=y
+CONFIG_EFI=y
+CONFIG_EFIVAR_FS=y
```
cbrune: Same thing here -- no need to set these EFI config options.  They are already enabled in the base config.

machine/alphanetworks/alphanetworks_snh61d2_320f/machine.make
```diff
@@ -0,0 +1,46 @@
+# Alpha Networks SNH61D2-320F
+
+ONIE_ARCH ?= x86_64
+SWITCH_ASIC_VENDOR = bcm
+FIRMWARE_TYPE = auto
```
cbrune: Same as the other machine -- no need to set auto when UEFI_ENABLE is set.

PR #718, opened 2018-03-20
Add empty dir for mounting USB device

pichuang:
The directory can avoid user mount over issue.
It would be more clear when user want to mount USB.

Signed-off-by: Phil Huang <phil_huang@edge-core.com>

### DIFF
--- a/rootconf/default/etc/init.d/makedev.sh
+++ b/rootconf/default/etc/init.d/makedev.sh
@@ -97,3 +97,5 @@ if grep -q securityfs /proc/filesystems ; then
 fi
 
 mkdir -p $ONIE_RUN_DIR
+
+mkdir -p $ONIE_USB_DIR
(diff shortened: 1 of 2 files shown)


pichuang:
I have met many customers mount USB on /mnt. It cause onie-nos-installer can't find /mnt/onie-boot because personal issue.

pichuang:
Refer the real case - https://github.com/Azure/SONiC/issues/161

cbrune:
This seems like a fine idea.  I would like to see the patch do this for all CPU architectures, not just x86.

To do that I would define the variable in `rootconf/default/lib/onie/function` and perform the mkdir in `rootconf/default/etc/init.d/makedev.sh`.

Follow the model of the `ONIE_RUN_DIR` variable.

pichuang:
@cbrune  Thanks for your suggestion. How about the latest PR?

cbrune:
Looks good, thanks!
Pushed to master:

- a2bd122 Add empty dir for mounting USB device

PR #722, opened 2018-03-26
gen-config: remove comments from machine.conf

orent:
Some NOSes do not accept '#' comments in config files. This causes the
installer to fail with a runtime error when parsing /etc/machine.conf

The original machine.conf did not contain comments. The patch changes
it back to the original format.

### DIFF
--- a/rootconf/default/etc/init.d/gen-config.sh
+++ b/rootconf/default/etc/init.d/gen-config.sh
@@ -26,17 +26,7 @@ gen_machine_config()
 
     gen_live_config > $live_conf
 
-    cat <<EOF > $machine_conf
-# /etc/machine.conf for onie
-
-#  Copyright (C) 2017 Curt Brune <curt@cumulusnetworks.com>
-#  Copyright (C) 2017 david_yang <david_yang@accton.com>
-#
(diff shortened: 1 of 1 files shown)


orent:
Specifically, it breaks the OpenNetworkLinux installer.

cbrune:
Sounds OK to me.

Funny, in that /etc/machine.conf was intended to be a shell script fragment from day 1.  That is how ONIE uses it internally.  Never occurred to me that folks would try to interpret it in a different manner.

How exactly does ONL try to parse this file?  What code does the parsing?

Pushed to master:
- 97811ddc312c gen-config: remove comments from machine.conf

orent:
This code:

https://github.com/opencomputeproject/OpenNetworkLinux/blob/master/packages/base/all/vendor-config-onl/src/python/onl/install/ConfUtils.py#LC24

This [Jan 4 commit](https://github.com/opencomputeproject/OpenNetworkLinux/commit/6aadd03984d505e8403f99ff44efe50f92e8ac16) appears to fix it - it can execute the file and dump the environment instead of trying to parse it directly. 

There is still a range of onie+onl version combinations that fails to install. It will take a while to flush them out.

PR #739, opened 2018-04-20
Update to Kernel 4.9.95

cbrune:


cbrune:
The NXP armv8 machines do not compile with this change.

It looks like some MTD related kernel patches no longer apply cleanly.

Perhaps these folks can help: @ZhaoQiang-b45475 @shengzhou

cbrune:
Thanks to @ZhaoQiang-b45475 for the quick fix for the NXP armv machines.  See f2cc2828e254.

All pushed to master now.

PR #745, opened 2018-05-02
update uboot and kernel for ls series platforms of nxp

ZhaoQiang-b45475:


cbrune:
Thanks, pushed to master:
- ec87e872d46b nxp: ls: update kernel patches from LSDK 1709 to LSDK 1803
- 6b3f5d3790bc nxp: ls: update uboot from 2017.07 to 2017.11

PR #750, opened 2018-05-21
Accton platforms: update kernel version to 4.9.95

david56:


cbrune:
This looks like a really great clean-up.  Thanks.

All pushed to master.

PR #756, opened 2018-06-20
Accton AS5916_54XKS: add to support ONIE

david56:


cbrune:
Looks good.  Thanks!

Pushed to master:
- 4854512216ea Accton AS5916_54XKS: add to support ONIE

PR #758, opened 2018-06-22
machine.conf: add onie_grub_image_name for old NOS installers

david56:
In ONIE v2018.05, the variable is renamed to `onie_uefi_boot_loader`.

The previous version of demo OS installers or some old NOS installers
cannot register correct EFI boot entry without `onie_grub_image_name`.
With the incorrect boot entry, it will fall back to ONIE environment
directly.

The patch addes `onie_grub_image_name` for legacy NOS/demo OS installers
to create correct boot entry.

The patch has been tested on Accton AS5712_54X and AS5912_54X.

Fixes: 0e56d23


cbrune:
Sounds reasonable.  Thanks for the fix.
Pushed to master:
- 5c4e0dfcf6d6 machine.conf: add onie_grub_image_name for old NOS installers

PR #760, opened 2018-06-26
Fixing a newline issue in busybox conf of delta/ag9064.

KrishnaKolakaluri:
Absence of a newline in busybox config file is causing compilation issue for delta/ag9064.


cbrune:
Thanks for the fix.  Pushed it to master.

PR #761, opened 2018-07-04
fix NXP arm64 updater issues

Ed-Swarthout-NXP:
Fix three issues using onie_update for arm64 on NXP ls1088ardb


installer/install.sh
```diff
@@ -139,7 +139,7 @@ check_machine_image()
 update_syseeprom()
 {
     # Set platform name and ONIE version to sys_eeprom.
-    if [ -x /usr/bin/onie-syseeprom ] ; then
+    if [ -x /usr/bin/onie-syseeprom -a -f /dev/mtd-board_eeprom ] ; then
```
cbrune: Why is this needed?

As written here, this will break all of x86, none of which have /dev/mtd-board_eeprom.  All of x86 uses i2c to access the onie syseeprom.

If you want to skip update_syseeprom() for a particular platform there are ways to do that.

EdSwarthout: nxp_ls1088ardb update was getting:

ERROR: Problems accessing sys_eeprom
Can't open /dev/mtd-board_eeprom: No such file or directory
Notice:  Invalid TLV header found.  Using default contents.
Notice:  Invalid TLV checksum found.  Using default contents.
Failure: Unable to install image: tftp://192.168.2.1/onie-updater-arm64-nxp_ls1088ardb-r0

But it looks like the correct fix is to remove the busybox onie-syseeprom config until correct i2c eeprom access is enabled.  (./machine/nxp/nxp_ls1088ardb/busybox/conf/config)

cbrune:
Hi @Ed-Swarthout-NXP -

Great news to hear the TLV info is now supported.  Very cool.

One thing though -- I would use the name `mtd-board_eeprom` as that name is already used by several other boards that store the TLV info in mtd.  Less confusing.  An example:
https://github.com/opencomputeproject/onie/blob/master/machine/accton/accton_as5610_52x/kernel/platform-accton-as5610_52x.patch#L86

Other than that it all looks good to me.  Thanks for adding the TLV support!

Ed-Swarthout-NXP:
OK, Updated MTD TLV info patch to use "hw-info" for partition name.
Also moved it to end of the 64MB NOR/QSPI so a full SDK install will not overlay it.

cbrune:
Looks great, pushed to master.  Thanks for all your time and efforts!

PR #762, opened 2018-07-20
Update default kernel to Linux 4.9.113

cbrune:
Update the default kernel to 4.9.113.

Also update all the machines using the previous default kernel.  As this is a pain in the neck, also add a script that help automate this task.

Side note -- it feels like it would be better for machines to specify "use the default", instead of hard coding a kernel version and having to the update that version all the time.  We should have a PR to add that feature.  Of course not all machine owner will want to do that, sticking with a specific kernel version has advantages in some circumstances.


cbrune:
That is a pretty good idea. I like it.  Let's cancel this PR and try something else.

cbrune:
Re-opening after remove nxp_ machines whose kernel patches no longer apply with 4.9.113.

PR #763, opened 2018-07-23
grub-arch: show BIOS mode during booting up

david56:
Showing a message to indicate that ONIE is booted up via which BIOS
mode (either legacy or UEFI).  The message would look like:

    Info: Mounting ONIE-BOOT on /mnt/onie-boot ...
    Info: BIOS mode: legacy

or:

    Info: Mounting ONIE-BOOT on /mnt/onie-boot ...
    Info: Mounting EFI System on /boot/efi ...
    Info: BIOS mode: UEFI

The patch has been tested on Accton AS7716_32X.


cbrune:
Sounds reasonable.  Pushed to master.

PR #765, opened 2018-08-01
Clarify big-endian values

scott-emery:
The IANA enterprise number and the CRC-32 are stored in big-endian format, but this was not previously specified.


cbrune:
Looks reasonable, push to docs.

This change will be reflected the next time the web pages are generated, which will happen in a couple of days corresponding with the 2018.08-rc1 release.

PR #766, opened 2018-08-02
Linux 4.9.113

cbrune:
Update the default kernel to 4.9.113.

Also update all the machines using the previous default kernel. As this is a pain in the neck, also add a script that help automate this task.

Side note -- it feels like it would be better for machines to specify "use the default", instead of hard coding a kernel version and having to the update that version all the time. We should have a PR to add that feature. Of course not all machine owner will want to do that, sticking with a specific kernel version has advantages in some circumstances.  Some interesting ideas are mentioned in #762 .


cbrune:
canceling this PR.  The NXP arm machines do not apply their kernel patches cleanly to 4.9.113.

I thought I could just leave them off of this kernel update, but that leads to other grossness.  We would need to preserve the common `patches/kernel/4.9.95` and `build-config/conf/kernel/4.9.95` even though they are identical to the 4.9.113 versions.

Leaving the default kernel version where it is until we come up with a better mechanism for maintaining the notion of "default kernel".

lakesh-sharma:
Does current default kernel 4.9.95 has all the security related patches/support for meltdown/spectre
issue , reported few months back.

cbrune:
Yes.  Those were addressed with 4.9.80.

Issue #767, opened 2018-08-06
Multiple vendor extension fields in system EEPROM


cbrune:
Regarding TLV type 0xFD, vendor extension, the spec declares that multiple instances of the field can be added if more space is required.

It does mean you can define multiple 0xFD TLVs, but it is not supported very well by the current tool.  I believe that statement was aspirational :)

onie-syseeprom does not *exactly* support multiple vendor-extension tlv codes.  The code will dump the multiple TLVs correctly, however it will not allow you to set multiple codes.

I am unaware of any vendors setting multiple 0xFD TLVs today.  If there are any, then they are using a custom tool to generate the TLV data.

Options on the table:

1. update the spec and remove the bit about multiple fields being possible.

2. update the onie-syseeprom tool to allow setting multiple 0xFD TLVs.

PR #768, opened 2018-08-14
ONIE-SYSEEPROM: Allow Multiple Vendor Extension TLVs

scott-emery:
Issue #767 

This patch modified the behavior of the onie-syseeprom -s (--set) command. If a TLV type code is specified which is allowed to appear multiple times in the EEPROM, then the TLV, if present, is not removed from the EEPROM before the new TLV value is added to the EEPROM. Currently, the only TLV type code which can appear multiple times within the EEPROM is the Vendor Extension (0xFD). Like all other TLVs added with the -s option, vendor extension TLVs are added at the end of the EEPROM, just before the CRC TLV. Deleting a vendor extension TLV with the "onie-syseeprom -s 0xfd" (no value) command will remove the first vendor extension TLV in the EEPROM. This command can be used multiple times to remove all vendor extension TLVs.


scott-emery:
Whoops. This is not the correct code.

PR #769, opened 2018-08-14
onie-syseeprom: Allow multiple vendor extension TLVs

scott-emery:
Issue #768 

This patch modifies the behavior of the onie-syseeprom -s (--set) command. If a TLV type code is specified which is allowed to appear multiple times in the EEPROM, then the TLV, if present, is not removed from the EEPROM before the new TLV value is added to the EEPROM. Currently, the only TLV type code which can appear multiple times within the EEPROM is the Vendor Extension (0xFD). Like all other TLVs added with the -s option, vendor extension TLVs are added at the end of the EEPROM, just before the CRC TLV. Deleting a vendor extension TLV with the "onie-syseeprom -s 0xfd" (no value) command will remove the first vendor extension TLV in the EEPROM. This command can be used multiple times to remove all vendor extension TLVs.

Testing has been done on an Accton 5712 platform by adding/deleting/replacing both vendor extension and MAC address TLVs.

Signed-off-by: Scott Emery <scotte@cumulusnetworks.com>

### DIFF
--- a/patches/busybox/feature-onie-syseeprom-command.patch
+++ b/patches/busybox/feature-onie-syseeprom-command.patch
@@ -8,7 +8,7 @@ SPDX-License-Identifier:     GPL-2.0
 
 diff --git a/include/24cXX.h b/include/24cXX.h
 new file mode 100644
-index 000000000000..5fb4641379ea
+index 0000000..5fb4641
 --- /dev/null
 +++ b/include/24cXX.h
 @@ -0,0 +1,66 @@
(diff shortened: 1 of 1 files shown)


cbrune:
This all seems very reasonable.

A couple of minor things:

1. please update the "-h" help output to document the special treatment of TLV 0xFD.

2. in the commit message please add "Closes:  #767", which is the issue this commit is fixing.  When the commit is ultimately pushed, the github system will automatically close #767.

scott-emery:
I've updated the -h help output to include verbiage about the vendor extension TLV being allowed to be included multiple times. Also "Closes: #767" is not included in the commit message.

cbrune:

- 72aee53783e1 onie-syseeprom: Allow multiple vendor extension TLVs

PR #773, opened 2018-09-05
Platform support for the DELL EMC S5100

sharma-lakesh:
It uses Rangeley C2538 as CPU module.  NPU used is  Cavium Xp70 switch ASIC.  It has 48x10G/25G and 6x40G/100G ports.


cbrune:
This looks reasonable to me.

```
```

ehdoyle:
Pushed to master 
d2088ba Platform support for the DELL EMC S5100 with the Rangeley C2538 CPU module. It uses Cavium Xp70 switch ASIC as NPU. It has 48x10G/25G and 6x40G/100G ports.

PR #775, opened 2018-09-11
machine_accton_wedge_16x:  Allow using m.2 storage

ggiamarchi:
Accton Wedge 16x micro server have both mSATA and m.2 slots. Currently, ONIE works only with mSATA devices because of the storage device detection function.

This patch allows to use either mSATA or m.2 storage.


cbrune:
This looks reasonable to me.

```
```

ehdoyle:
Commit looks good. Thanks!
- adfb83e machine_accton_wedge_16x: Allow using m.2 storage

PR #780, opened 2018-10-16
Prevent halt from bricking switch on NE2572 and NE10032

mrhpearson:
We have an issue on our NE2572 and NE10032 switches where if you do halt or poweroff the BIOS shuts down and then can't be recovered even by a power cycle. 
This is a fairly ugly, but simple, workaround to prevent that happening by preventing the ACPI message being sent to the BIOS that causes the problem. 
Let me know if any questions. 
Thanks
Mark

### DIFF
--- a/machine/lenovo/lenovo_ne10032/kernel/prevent_halt_issue.patch
+++ b/machine/lenovo/lenovo_ne10032/kernel/prevent_halt_issue.patch
@@ -1,0 +1,19 @@
+diff --git a/drivers/acpi/acpica/hwxfsleep.c b/drivers/acpi/acpica/hwxfsleep.c
+index 3b37676..4bbdf19 100644
+--- a/drivers/acpi/acpica/hwxfsleep.c
++++ b/drivers/acpi/acpica/hwxfsleep.c
+@@ -367,6 +367,14 @@ acpi_status acpi_enter_sleep_state(u8 sleep_state)
+ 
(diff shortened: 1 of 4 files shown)


cbrune:
At Cumulus we have seen this type of problem before from other vendors.  It is a nasty support call to take.

I believe the real fix is to update the BIOS -- ask someone from Accton, they had this problem several years ago.

What you have here works for ONIE, but it is a time bomb waiting for all OS vendors on these platforms....  Can't expect every OS vendor to patch their kernels.

mrhpearson:
Thanks Curt,

Interesting to know others are seeing it too - I will follow up with our BIOS vendor. If it is OK to add this to ONIE that would be great - I'm finding that getting BIOS updates done can be really tough sometimes.

Mark

cbrune:
I think several hw vendors hit this the first time they made x86 white box switches.

They transitioned from a proprietary NOS that never offered `halt` or `poweroff` to a NOS running Linux where `halt` and `poweroff` are pretty normal.  Also moving from PowerPC or ARM to x86 uncovers all kinds of wack-o x86 baggage.

cbrune:
This looks reasonable to me.

```
```

ehdoyle:
- d25cf93 Prevent halt from bricking switch on NE2572 and NE10032
Nice work. Thanks!

PR #786, opened 2018-11-08
Efi sed fix

mrhpearson:
We have a switch and the output of the 'efibootmgr -v' command has extra chars on the end of the line.
e.g.
Boot0003* ONIE: Open Network Install Environment	HD(1,GPT,15d105fb-a59a-43d3-be68-c849f9132c4b,0x800,0x80000)/File(\EFI\ONIE\GRUBX64.EFI)..BO

The ..BO after the file name thows the sed command used in onie-blkdev-common so we get error messages like;
ERROR: EFI boot variable not found: ONIE:

I've changed the regexp to just discard everything from the ')' onwards so shouldn't impact regular items (and I did check on one of our other platforms that doesn't have this problem).

Thanks
Mark


rootconf/grub-arch/sysroot-lib-onie/onie-blkdev-common
```diff
@@ -289,7 +289,7 @@ uefi_rm_stale_bootnum()
 {
     efibootmgr -v | grep ')/File(' | awk '{ print $1 }' | while read boot_num ; do
         local boot_num=${boot_num%\*}
-        local boot_path=$(efibootmgr -v | grep $boot_num | sed -e 's/.*)\/File(//' -e 's/)//' -e 's/\\/\//g')
+        local boot_path=$(efibootmgr -v | grep $boot_num | sed -e 's/.*)\/File(//' -e 's/).*//')
```
cbrune: I think zapping everything after the ')' is very reasonable.

However, this patch dropped the last sed statement `-e 's/\\/\//g'`, which provides a valuable service on some systems I have seen.  In the file path, this statement translates MS-DOS style back-slashes `\` into UNIX style forward-slashes `/`.  I think we need to keep that part.

mrhpearson: Sure - let me test it out with that part back in (I'd misunderstood what it did) and confirm it works as expected.

mrhpearson: I put it back in and everything is working great. I think the change shows up below - let me know if it doesn't :)
Mark

ehdoyle:
Looks good. I'm going to apply it as two commits just to keep awareness about the backslash conversion.

Signed-off-by Alex Doyle adoyle@cumulusnetworks.com

PR #790, opened 2018-11-27
pesign: use comment method to bypass -Werror=implicit-fallthrough

cbrune:
The gcc attribute 'fallthrough' was introduced with gcc version 7 and
breaks compilation using older compilers.  To achieve portable code,
use the "comment method" described in the gcc documentation for the
-Wimplicit-fallthrough=3 gcc option.

Fixes: bae571195b87 ("pesign: update for compiling with gcc-7")
Signed-off-by: Curt Brune <curt@brune.net>

### DIFF
--- a/patches/pesign/implicit-fallthrough.patch
+++ b/patches/pesign/implicit-fallthrough.patch
@@ -1,12 +1,12 @@
-Summary: From: Curt Brune <curt@brune.net>
-
 implicit-fallthrough.patch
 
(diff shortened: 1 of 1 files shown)


ehdoyle:
I successfully built a whole bunch of different targets with GCC 6.3,  and the couple of targets I tested with Ubuntu 18 using  GCC 7 worked as well, so this all looks good now.
Thanks!
Signed-off-by: Alex Doyle adoyle@cumulusnetworks.com

PR #791, opened 2018-11-29
Lenovo ne0152t

mrhpearson:
Support for Lenovo NE0152t platform


cbrune:
This PR is a little weird - it has 6 commits in it.

Can you rebase this branch on master first?  That way all we see is the `Lenovo NE0152T platform support` commit.

PR #794, opened 2018-11-30
Add Alpha Networks SNH61A0-640F: 64*QSFP28 with Intel Rangeley C2558.

YuChang9565:
SNH61A0-640F have 64x100 Gigabit Ethernet QSFP28 ports with Broadcom BCM56970 switch ASIC. The CPU is Intel Rangeley C2558 Quad Cores/2.4GHz and the default console device is ttyS1.


ehdoyle:
Looks good and built without issue. Thanks!

PR #796, opened 2018-12-06
Add copyright notices to machine.make files

mrhpearson:
Hi - I know this is a tedious one to review, but whilst internally reviewing our open source code for compliance it was pointed out to me that the GPL license is missing from the machine.make files.

I'm not 100% convinced the files deserve or need a GPL licence, but it was easy to add.

Let me know if any objections :)

Thanks
Mark


ehdoyle:
Looks good. I've asked around to be sure but don't think that should limit ONIE at all from a legal standpoint as we've got GPLv2 code all over the place.  Thanks!

PR #797, opened 2018-12-10
   Support the different platform name with the same design CPU board

hans-tseng:
1. In order to share the same onie code base with different platforms name but have
       the same design on CPU board. We add the build option,ONIE_BUILD_MACHINE, to build the different platform name. Please see the attached file Supported_platform.md for more detail.

Signed-off-by: hans <hans.tseng@deltaww.com>

### DIFF
--- a/machine/delta/delta_ag9032v2a/Supported_platform.md
+++ b/machine/delta/delta_ag9032v2a/Supported_platform.md
@@ -1,0 +1,14 @@
+There are some of platforms which use the same design of CPU board.
+So we share the same onie's source with defferent model name. 
+if using the default way to build onie and we will get the ag9032v2a platform.
+
+    $ cd build-config
+    $ make -j4 MACHINEROOT=../machine/delta MACHINE=delta_ag9032v2a all
+
+We use the delta_ag9032v2a as base to extend other model and list the externed model in the following.
+ 
+1. delta_agc7648sv1 
+    $ cd build-config
+    $ make -j4 MACHINEROOT=../machine/delta MACHINE=delta_ag9032v2a ONIE_BUILD_MACHINE=delta_agc7648sv1 all
+
+


hans-tseng:
Hi ehdoyle:
   please see my reply  as following
1 - Can you provide an example of how this additional argument would be used ? 
As a general rule I'd like to keep the builds as similar as possible, 
and I'd like to understand the implications of adding something new.

    A: We add the option "ONIE_BUILD_MACHINE" to change the platform,s and it will modify the 
         "onie_build_machine" & "onie_build_machine" on the machine.conf and machine-build.conf.

2 - To address a common code base for multiple products, is it feasible to define an 
ONIE machine target that is shared by all the different versions of the same hardware,
 and is deployed during manufacture on a per-product basis ?

    A:Yes,using the common code might be a proper way but we somtimes have the extend model 
       and it may not all fit the common code.  
  
3 - If there are differences that need to be present in the code, then that would seem
 to imply each build target would require it's own directory, even if that's mostly a copy 
 of another one, but to be fair these directories aren't terribly big. Would that work? 
 This also leads to the question:

    A: For example of delta_agc7648sv1, it uses the same CPU board design but MAC ASIC. If it use its
        directory and that would creat the new platform not extended model. So we figure this way out to 
        add option value "ONIE_BUILD_MACHINE" which base on the ag9032v2a directory to build the 
        agc7648sv1
 
4- How many product variants are expected? Adding a few more directories is not a big
 problem, but if it's a hundred, then a more dynamic solution would be required.
Thoughts?
      A : I think we do not have  hundred exteded model would be commited.

cbrune:
This topic has come up before the mailing list:
http://lists.opencompute.org/pipermail/opencompute-onie/2017-July/001475.html

Also, support for common CPU modules was added here:
https://github.com/opencomputeproject/onie/pull/592

This commit in particular:
https://github.com/opencomputeproject/onie/pull/592/commits/5b3756e2b496b3d64f406d39b0e7cc7c65a9d32f

The commit comment explains what to do.

hans-tseng:
Hi, cbrune:
     Thanks your time to explain how to share the ONIE with common CPU board designed.
   And I think we did that the way that you mentioned but we do not want to change the ONIE 
   platform's name in run-time but did it while building ONIE image. So we add the "ONIE_BUILD_MACHINE" option parameter and here is the command to build make -j4 
   MACHINEROOT=../machine/delta MACHINE=delta_ag9032v2a 
   ONIE_BUILD_MACHINE=delta_agc7648sv1 all. There are example of the two platforme's onie-builf.conf in 
   ag9032v21 and agc7648sv1   

    ag9032v2a:
     	        onie_version=ag9032v2a-201808211203onie_version_1.0
		onie_vendor_id=2254
		onie_build_machine=delta_ag9032v2a
		onie_machine_rev=0
		onie_arch=x86_64
		onie_build_platform=x86_64-delta_ag9032v2a-r0
		onie_config_version=1
		onie_build_date="2018-08-21T12:03+08:00"
		onie_partition_type=gpt
		onie_kernel_version=4.1.38
		onie_firmware=auto
		onie_switch_asic=bcm
		onie_skip_ethmgmt_macs=no
		onie_uefi_boot_loader=grubx64.efi
		onie_uefi_arch=x64

	agc7648sv1:
		onie_version=agc7648sv1-master-201808281123onie_version_1.0
		onie_vendor_id=2254
		onie_build_machine=delta_agc7648sv1
		onie_machine_rev=0
		onie_arch=x86_64
		onie_build_platform=x86_64-delta_agc7648sv1-r0
		onie_config_version=1
		onie_build_date="2018-08-28T11:23+08:00"
		onie_partition_type=gpt
		onie_kernel_version=4.1.38
		onie_firmware=auto
		onie_switch_asic=bcm
		onie_skip_ethmgmt_macs=no
		onie_uefi_boot_loader=grubx64.efi
		onie_uefi_arch=x64

     You can see that the platform name is changed by form delta_ag9032v2a to delta_agc7648sv1, even the image name was changed. We hope to add this text file to tell our customers how to build the other platforms with same CPU board designed without change it run-time. 

Hans

ehdoyle:
Hi - so if I understand the request correctly, you'd have one common code base for ONIE, then separately built binaries that are platform particular, which would have to be installed on their matching switch?
I did a bit of digging after Curt had pointed out the existing method for addressing this, which is to
have a common build of ONIE that can detect what hardware configuration it is using, and take action off of that.  This way only one build is required to handle all cases.  I'd suggest taking that approach since there are already other systems already doing it this way, ( plus if any of the other platforms find bugs with this approach, everybody benefits from the fix )

In addition to the commits Curt referenced in his comment I can add that the feature is in the documentation here:
https://opencomputeproject.github.io/onie/developers/multiple_cpu.html?highlight=onie_machine

And an example of a system that does this can be found under the quanta directory:
https://github.com/opencomputeproject/onie/blob/master/machine/quanta/quanta_common_rglbmc/rootconf/sysroot-lib-onie/gen-config-platform

Hope this helps,
-Alex

PR #798, opened 2018-12-11
update uboot and kernel

ZhaoQiang-b45475:
update uboot from 2017.11 to 2018.03 and kernel from LSDK1803 to LSDK1809


ehdoyle:
Hi.  Any chance this commit can be broken down into a series of smaller, more focused commits? Otherwise it's going to be a while before I can properly review all 653 files in this change set.
Thanks.

ZhaoQiang-b45475:
Hi Alex,
I am sorry for that, but it is hard to break down them. Actually, most of the modification are uboot patches and kernel patches needed for NXP machine, and they are added under directory machine/nxp/.
In the first patch, I updated the uboot to 2018.03 and rebased the uboot patches for the new uboot, it is hard to break down them, otherwise compile issue will be caused. And so do the kernel part.

ZhaoQiang-b45475:
Thank you!

PR #801, opened 2018-12-27
Accton AS7316_54XS: add to support ONIE

david56:
The machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on Broadcom
Tomahawk+ switch ASIC.  CPU module is Intel Xeon D-1518.


ehdoyle:
Looks good, and builds fine. Thanks!

PR #802, opened 2018-12-27
Accton AS7716_24XC: add to support ONIE

david56:
The machine has 16x100GbE QSFP28+ ports and 8 slots supporting
CFP2 modules on Broadcom Tomahawk switch ASIC.  CPU module is Intel
Xeon D-1518.


ehdoyle:
Builds fine. Thanks!

PR #809, opened 2019-01-25
Add CIG CS6436-56P device into ONIE system

michealylj1:
It's new device  CS6436-56P from CIG :
ASIC Vendor: Nephos
Swich ASIC: NP8366
Port Configuration: 25GX48,100GX8


ehdoyle:
Code looks good and builds clean - I squashed the two commits into one. 
In case I'm looking at this in the future, I did double check the new code for the address for the 3rd serial comm port being passed to Grub, and while it may be particular to this hardware, the serial interface is standard so I was OK with it applying to other X86 builds as well.
Thanks!

PR #810, opened 2019-01-29
Add Alpha Networks SNJ60B0-320F: 32*QSFP-DD+2*SFPP/Intel Broadwell-DE D-1518.

Mingxikao:
SNJ60B0-320F has 32 x 400 Gigabit Ethernet QSFP-DD ports and 2 x 10 Gigabit Ethernet SFPP ports with Broadcom BCM56980(Tomahawk3) switch ASIC. The CPU is Intel Broadwell DE D-1518 Quad cores/2.3GHz and the default console device is ttyS0.


ehdoyle:
Hi. Looks like everything built fine. I did reformat the commit message to get it into 80 column format, but that was it.
Thanks for the commit!

PR #811, opened 2019-02-01
Increase I2C write delay to fix eeprom corruption issues

mrhpearson:
We found on these boards the i2c write delay needed increasing to 5ms to avoid corrupt writes.


ehdoyle:
Both platforms built with no issues (which is no surprise given this was a minor change, but I've learned to trust nothing ;).
Thanks for the commit!

PR #813, opened 2019-02-19
fixed the DUT which can not access the syseeprom

hans-tseng:
Signed-off-by: hans <hans.tseng@deltaww.com>
Fix the VUT can't access the syseeprom because of the default value of CPLD reg was changed


ehdoyle:
Changes looked pretty straightforward, and the delta_ag9032v2a rebuilt without issue.
Thanks for the fix!
-Alex

PR #825, opened 2019-03-18
Use available interface for arping instead of default eth0

mrhpearson:
The NE0152T board has eth2 as it's mgmt port  instead of eth0

The arping command is by default using eth0 and should be using the available $intf interface. Use the -I option to make sure the active and available interface option is used.

Tested on NE0152T (uses eth2) and NE2572 (uses eth0)


ehdoyle:
I double checked that intf should always be set by the calling function so I'd say it's good to go.
Thanks!

PR #827, opened 2019-04-03
add to support new machines of Accton Wedge100BF series

david56:


ehdoyle:
Everything builds and looks good. Thanks for the commit!

PR #833, opened 2019-05-02
Fix "make clean" for gnu-efi, pesign and shim

kashuwong:
The "make clean" target misses gnu-efi-clean, pesign-clean and shim-clean.

This patch adds these clean targets to user-clean or machine-clean.


ehdoyle:
Nice catch. Thanks for the patch. Checking the code base I found another spot where targets were assigned to the USERSPACE_CLEAN as well.  Thanks!

PR #834, opened 2019-05-02
Accton AS4620_54PE: add to support ONIE

david56:
The machine has 48x1/2.5GbE RJ45, 4x25G SFP28 and 2x40GbE QSFP+ ports on
Broadcom Hurricane3 switch ASIC.  CPU module is Intel Rangeley C2538.


ehdoyle:
The commit looks good and built without issue. Thanks!

PR #835, opened 2019-05-02
Accton AS7326_56X: add to support ONIE

david56:
The machine has 48x25GbE SFP28, 8x100GbE QSFP28 and 2x10GbE SFP+ ports
on Broadcom Trident3 switch ASIC.  CPU module is Intel Xeon D-1518.


ehdoyle:
Commit built without issues. Thanks!

PR #836, opened 2019-05-02
Accton platforms: support to install on mSATA SSD

david56:


ehdoyle:
Builds clean. Thanks for the commit!

PR #839, opened 2019-05-15
Accton AS7715_32X: add to support ONIE

david56:
The machine has 32x100GbE QSFP ports on Broadcom Tomahawk switch
ASIC.  CPU module is Intel Denverton C3508.


ehdoyle:
Pull request looks good - built without issue. Thanks!

PR #840, opened 2019-05-15
Accton AS5816_54X_TD2: add to support ONIE

david56:
The machine has 48x1/10GbE SFP+ and 6x40GbE QSFP ports on
Broadcom Trident2+ switch ASIC.  CPU module is Intel Xeon D-1518.


ehdoyle:
Compiled without issue. Looks good. Thanks!

PR #842, opened 2019-05-23
Accton AS7916_54XK: add to support ONIE

david56:
The machine has 48x1GbE/10GbE/25GbE SFP28 and 6x40GbE/100GbE QSFP28
ports on Broadcom Jericho switch ASIC.  CPU module is Intel Xeon D-1518.


ehdoyle:
Built with no issues. Looks good. Thanks!

PR #843, opened 2019-05-23
Accton AS7316_26XB: add to support ONIE

david56:
The machine has 16x10GbE SFP+, 8x25GbE SFP28 and 2x100GbE QSFP28
ports on Broadcom Qumran switch ASIC.  CPU module is Intel Xeon
D-1518.


ehdoyle:
Code builds fine. Thanks for the commit!

PR #846, opened 2019-06-06
changes for AS5916 series platforms

david56:


ehdoyle:
Tested with rebuild of AS5916_54XK and AS5915_54XKS - both looked good. Thanks!

PR #847, opened 2019-06-06
Accton AS5916_54XL: add to support ONIE

david56:


ehdoyle:
New platform built fine. Thanks!

PR #850, opened 2019-06-14
The machine has 48x25GbE SFP28 and 8x100GbE QSFP28 ports …

tingjack:
…with BCM56873 and the CPU module is Intel Broadwell-DE.

Signed-off-by: Ting.Jack <ting.jack@inventec.com>


ehdoyle:
Confirmed that the code built cleanly. Thanks for the commit!

PR #851, opened 2019-07-17
Inventec D10064: add to support ONIE

WillieYangInventec:
The machine has 64x100GbE QSFP28 ports
The CPU module is Intel Broadwell-DE

Signed-off-by: Willie <yang.willie@inventec.com>


ehdoyle:
The platform built cleanly, and has been added to ONIE.
Thanks for the commit!
-Alex

PR #854, opened 2019-07-24
Fixed typo

YannickFricke:


ehdoyle:
Nice catch. Thanks for the fix!

PR #855, opened 2019-07-26
Juniper QFX5210-64C-S: add to support ONIE

ciju-juniper:
This switch has 64 QSFP28 (40G/100G) ports, 2 SFP+ (1G/10G) ports on Broadcom Tomahawk II chipset. CPU used in QFX5210-64C-S is Intel Broadwell-DE. The machine has Redundant and hot-swappable Power Supply (1+1) and also has Redundant and hot swappable fans (3+1).


ehdoyle:
The commits have been squashed for this pull request as it is the introduction of a new platform. ONIE builds without issue, although the demo target build is in the process of being implemented. My rationale for taking the pull request is that it provides a clean starting point for resolving the demo target build and most importantly ONIE builds cleanly so what is coming in definitely adds value.
Thanks for the commit!

PR #856, opened 2019-07-31
Pegatron FN_6254_DN_F: add to support ONIE

PegaSwitch:
Hi, greeting from Pegatron.

This it first pull request from Pegatron, to support new manchine "FN_6254_DN_F".

This machine has 48x25GbE SFP28 and 6x100GbE QSFP28 ports on Nephos switch ASIC.
CPU module is Intel Denverton C3538.

This patch is very common, just add only new machine "FN_6254_DN_F" stuff with extract busybox and Linux patches which may also use in other machines as well.

In this machine, we used "coreboot with payload GRUB2" instead of normal BIOS, so there are many GRUB menu need to be handled, this is why a lot of files existed in "rootconf" folder, major for

 1. GRUB munu preparation
 2. Pegatron DIAG checking, and
 3. mutliple Disk scan for GRUB config checking and NOS install destination selection.


ehdoyle:
All code built fine and it looks good.
Thanks for the commit!

PR #857, opened 2019-08-04
busybox: enable xmodem

xdel:
This brings ability do download images to the ONIE environment
when management network is inaccessible by some reason.


xdel:
@ehdoyle well, I've had recently hands on a perfectly functioning switch whose physical management port got some invisible damage and PHY was not able to detect a link anymore, though device still has some worth as a completely standalone setup. To install the OS with ONIE wrapper, I've had not much options aside making FAT partition on a flash and uploading the installer over U-Boot (it was an ancient one with controlplane based on mpc850). Honestly I could not imagine developer's device where OS needs to be installed over ONIE while network interface haven't been brought up yet, so let it be only a poor sysadmin's usage case.

ehdoyle:
Interesting - I'd not considered the possibility of physical damage that left the switch mostly functional 
I'm assuming the FAT partition you'd mentioned was on a USB drive? That's usually the option of last resort to get data onto a switch when the network is not an option.

billie-alsup:
Is there a reason to turn it on globally, versus letting a platform turn it on themselves within their own busybox config?  i.e. machine/<vendor>/<platform>/busybox/conf/config

xdel:
@ehdoyle that was a qct's t1048-lb9, so no other options was available. The footprint increase for the bb binary is pretty small, the min-entropy increase is about 1100b and (rounded) binary file size isn't changed at all.

billie-alsup:
I don’t quite follow your response.  There are no busybox patches for that machine today, but the busybox.make is setup to handle such fragments as optional per machine. So theoretically it is simply a matter of adding a machine/quanta/quanta_lb9/busybox/conf/config file enabling the feature, right?

I don’t really have a problem with it as globally enabled; it just seems a very specific fix for a specific problem that is not really applicable to the majority of devices.   I would suspect most businesses would RMA or otherwise replace such a broken device.

I suppose  there could also be some concerns over generically allowing a download over the serial port, as that might not be something easily traced/tracked in most environments (whereas network monitoring might be present). Console access does not imply physical access, so it does open another attack vector. That said, it appears most vendor implementations do not appear to support a SecureBoot environment anyways (although perhaps vendors offer SecureBoot builds to customers directly, since they obviously wouldn’t commit their private key into the ONIE repository!)

xdel:
@billie-alsup I could agree with the statement that it could be a machine-dependent conditional, but certainly disagree that it poses additional security risk for the hardened installation - malicious operator could use all richness of text mangling tools being included already in ONIE to transfer whatever binary s/he wants over a regular console session. Same applies to an extreme case where someone is using a data-diode-style proxy for operator-device interaction, passing xmodem transfer through this kind of device may reveal already existing hole in the security provided by that diode but by no mean availability of the 'rx' command on a target device extends an *actual* attack surface in this case.

billie-alsup:
Yes, I agree that the generic ONIE is not secure to begin with.  Again, I don't have a problem with it being generically enabled; just trying to understand why.  If a particular vendor or machine doesn't want it, seems like they can still turn it off in their own busybox config file afaik.

xdel:
Funny but it turned out to be an act of (in)voluntarily sabotage from datacenter personnel. In other words, physical connector was fine but somehow not connected by different patchcords on multiple occasions. I also was a hard believer in a situation where port shows no signs of physical damage and PHY happily responds to configuration requests, think we may close this request then.

PR #860, opened 2019-08-20
Adding the 'demo' target for Juniper QFX5210 platform

ciju-juniper:
This patch set does the following:
 - Defining the ONIE version tag to a fixed string
 - Adding the demo target 
 - Updating the documentation


ciju-juniper:
Attaching the test logs
[Onie_demo_installer_boot.txt](https://github.com/opencomputeproject/onie/files/3522076/Onie_demo_installer_boot.txt)

ehdoyle:
Demo target built successfully. Thanks!
I've incorporated this pull request. One thing you may want to check is the commit to define the ONIE version to a fixed string - in there it's 2019.08-rc2, which is the release candidate build. There's a 2019.08 that is the official release. Apart from the version the two branches should be the same, but it might avoid some confusion in the future if you're using the 2019.08.
Thanks again!

ciju-juniper:
Thanks Alex.

The reason why I chose 'rc2' was that patches are not part of 2019.08 release of onie. I can make it as the next onie release string. Please let me know if that is fine. In that case, please let me know the next release month as well.

Thanks
Ciju
________________________________

PR #867, opened 2019-10-01
Alphanetworks STX-60D0-062F: Use acpi --no-ebda in ISO recovery menu

talisych:
In the STX-60D0-062F, grub's "acpi" command cannot copy intact EBDA.
(Extended BIOS Data Area) if the EBDA size of BIOS is over 1K.
This command causes system hang.
Therefore the "acpi" command can only be executed with the
"--no-ebda" parameter.

This patch modifies the grub configuration of the recovery ISO prior to the
ISO file's creation, and the original grub-iso.cfg will be restored
after the ISO file is created.

### DIFF
--- a/machine/alphanetworks/alphanetworks_stx60d0_062f/machine.make
+++ b/machine/alphanetworks/alphanetworks_stx60d0_062f/machine.make
@@ -39,6 +39,7 @@ LINUX_MINOR_VERSION	= 95
 # Older GCC required for older 3.14.27 kernel
 #GCC_VERSION = 4.9.2
 
+include $(MACHINEDIR)/rootconf/grub-machine.make
 
 #-------------------------------------------------------------------------------
 #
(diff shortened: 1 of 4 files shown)


ehdoyle:
Hi - thanks for the commit. I'm updating the git commit to have most of the text from your description at the top of the pull request so anyone browsing the git history will have a better sense of what the patch does. 
I wasn't initially familiar with the --no-ebda option, so after looking it up, I reworded the commit message a bit to make it a bit clearer what was happening here, but wanted to make sure I was being accurate about the changes. Does this revised wording look correct?
Thanks!
------------------------
Alphanetworks STX-60D0-062F: Use acpi --no-ebda in ISO recovery menu       
                                                                           
In the STX-60D0-062F, grub's "acpi" command cannot copy intact EBDA.       
(Extended BIOS Data Area) if the EBDA size of BIOS is over 1K.             
This command causes system hang.                                           
Therefore the "acpi" command can only be executed with the                 
"--no-ebda" parameter.                                                     
                                                                           
This patch modifies the grub configuration of the recovery ISO prior to the
ISO file's creation, and the original grub-iso.cfg will be restored        
after the ISO file is created.

talisych:
Hi, thanks for your comment. Yes, the revised wording is correct and I will update the commit message based on your suggestions. Thanks!

PR #869, opened 2019-10-09
Accton AS4630_54PE: add to support ONIE

david56:


ehdoyle:
Hi David - the code for disabling grub configurations looks OK, but I'd like to add some context to the commit as to why we'd want to disable this ( like, hardware doesn't support it, or it looks like GRUB changes can be applied when they won't for some reason, etc. ).
Thanks!

david56:
@ehdoyle ,
It was my bad.  I've removed the commit of disabling grub configurations because it works only in UEFI boot mode.  The commit just intended to prevent grub booting in UEFI mode from showing messages of unsupported configurations.  To make AS4630_54PE boot in either legacy or UEFI mode, we would like to keep the grub configurations now.  Thank you.

ehdoyle:
Merged and built clean. Thanks!

Issue #870, opened 2019-10-30
ONIE build process missing packages


konstcranky:
Today i've tried to build onie 2019.02br
It seems some packages that ONIE uses is absent ? 

make: *** Waiting for unfinished jobs....
failed: Connection refused.
failed: Connection refused.
failed: Connection refused.
failed: Connection refused.
http://zlib.net/zlib-1.2.8.tar.gz:
2019-10-30 12:52:02 ERROR 404: Not Found.
wget: unable to resolve host address 'softlayer-dal.dl.sourceforge.net'
ERROR:  Unable to download zlib-1.2.8.tar.gz
make/zlib.make:39: recipe for target '/users/kostya/iss/onie/build/download/zlib-download' failed
make: *** [/users/kostya/iss/onie/build/download/zlib-download] Error 1
2019-10-30 12:52:02 URL:http://savannah-nongnu-org.ip-connect.vn.ua/dmidecode/dmidecode-3.1.tar.xz [61204/61204] -> "/users/kostya/iss/onie/build/download/dmidecode-3.1.tar.xz" [1]
dmidecode-3.1.tar.xz: OK
2019-10-30 12:52:04 URL:https://sourceware.org/git/?p=lvm2.git [35142] -> "/users/kostya/iss/onie/build/download/lvm2-2_02_105.tar.xz" [1]
lvm2-2_02_105.tar.xz: FAILED
sha1sum: WARNING: 1 computed checksum did NOT match
ERROR: Unable to verify sha1 for lvm2-2_02_105.tar.xz
make: *** [/users/kostya/iss/onie/build/download/lvm2-2_02_105-download] Error 1
make/lvm2.make:87: recipe for target '/users/kostya/iss/onie/build/download/lvm2-2_02_105-download' failed
2019-10-30 12:52:05 URL:https://matt.ucc.asn.au/dropbear/releases/dropbear-2016.74.tar.bz2 [1622234/1622234] -> "/users/kostya/iss/onie/build/download/dropbear-2016.74.tar.bz2" [1]
dropbear-2016.74.tar.bz2: OK

....

==== Getting upstream flashrom ====
failed: Connection refused.
https://web.daniel-baumann.ch/files/software/dosfstools/dosfstools-3.0.26.tar.xz:
2019-10-30 13:34:56 ERROR 404: Not Found.
ERROR:  Unable to download dosfstools-3.0.26.tar.xz
[...]

PR #871, opened 2019-11-01
Support system wide download cache at /var/cache/onie/download

ehdoyle:
The problem:

ONIE downloads specific versions of packages (most as compressed
tarfiles ) that get used in build. These packages are checksummed
against values stored in git, and are stored in onie/build/download.
These files are not removed on clean operations, although the compiled
versions of them are, so once created, onie/build/downloads is
fairly static, unless another ONIE build target is built, in which case
it may have to download additional packages, or different versions
of ones that it already has.

On a first-time checkout of the ONIE source code, this directory
gets populated by makefiles using ./scripts/fetch-package.

This script will iterate through a list of URLs to download the
source code. It starts with mirror.opencompute.org/onie, and
will then try other URLS, if any.

As some of the alternate URLs are old, they are no longer valid,
and mirror.opencompute.org/onie becomes the only reliable way to get them.

If something happens to this website, or the user has limited
Internet access, building ONIE becomes pretty much impossible.

...and it was the website going offline for a few days prompted this fix.

The solution:

The scripts/fetch-package script, as a last resort, will now
look for a system wide cache at /var/cache/onie/download.

Setting up this cache is entirely the responsibility of the
user, as it is not needed under normal circumstances. However if:

1 - Internet connectivity is, slow, intermittent, or non existent.
2 - or something happens to the mirror server.

...it might be nice to have a local copy of these files that
never change.

As of 10/31/19, the entire mirror takes up about 2GB of space, and
most users will only need a subset of these packages, so it
may be trimmed for individual use.

To set up the local copy:

One can either build all ONIE targets that one is interested in,
and copy the resulting onie/build/download folder to
/var/cache/onie/download

Or:

mkdir /var/cache/onie/download

...and poplulate it with the entire mirror using:

wget --recursive --cut-dirs=2 --no-host-directories \
 --no-parent --reject "index.html" http://mirror.opencompute.org/onie/

This was originally brought up in:
Bug URL: https://github.com/opencomputeproject/onie/issues/870

Fixes:  #870   ONIE build process missing packages

### DIFF
--- a/build-config/scripts/fetch-package
+++ b/build-config/scripts/fetch-package
@@ -2,6 +2,7 @@
 
 #  Copyright (C) 2013-2014 Curt Brune <curt@cumulusnetworks.com>
 #  Copyright (C) 2016 Pankaj Bansal <pankajbansal3073@gmail.com>
+#  Copyright (C) 2019 Alex Doyle <adoyle@cumulusnetworks.com>
 #
 #  SPDX-License-Identifier:     GPL-2.0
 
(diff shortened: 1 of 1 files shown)


ehdoyle:
Merged.

PR #872, opened 2019-11-02
Update FAQ to explain /var/cache/onie/downloads local package mirror dir

ehdoyle:
The commit for #870 goes into more depth. The short of it is that if
network downloads fail, ONIE will look for local packages under
/var/cache/onie/downloads. The trick is that the user has to have
had the foresight to set up this cache while the network is still
functional.

Fixes: #870


ehdoyle:
Merged.

PR #873, opened 2019-11-06
Alphanetworks STX-60D0-062F: Make --no-ebda applied on each build.

talisych:
Remove iso-grub-machine-complete stamp to make sure the customized grub-iso.cfg will be applied on each build.


ehdoyle:
Hi - I'd like to put a bit more context in the commit message. 
So the patch here is basically to make sure that the rescue iso grub menu has 
    acpi --no-ebda
if the platform has efi?
Apart from that it looks like everything built fine, so if you can confirm I've interpreted the patch correctly I'll just note that in the commit and get it pulled in.
Thanks!

talisych:
Hi Alex,
Yes, the purpose of the patch is to fix the compilation procedure. 
For example, currently the grub-iso.cfg file will be restored to the original file after compilation finish and the stamp(iso-grub-machine-complete) file would not be deleted.
If the stamp still existed, the apci --no-ebda would not be applied to rescue iso grub menu in the next compiling, thus I modify the building script to delete the stamp.

ehdoyle:
Excellent. I updated the commit and just pushed. Thanks for the pull request and the clarification!

Issue #82, opened 2014-11-10
Local filesystem discovery does not work for UBIFS (NAND Flash)


cbrune:
ONIE has a method where it mounts the local block device partitions looking for an installer.

Currently this method does not include UBI volumes.  The current method looks in /proc/partitions for block devices.

The method should also consult /sys/class/ubi/ubi0

``` bash
    for dev in /sys/class/ubi/ubi0/ubi0_?; do
        ubiname=`cat $dev/name`
        mount -t ubifs ubi:$ubiname /mnt
        look for image ....
    done
```

PR #878, opened 2019-11-15
UBIFS Support for Local Filesystems

scott-emery:
This patch adds support for installing an OS from a locally attached UBIFS file system, closing an issue which has been open since 2014. An additional discovery function, called sd_localubifs(), has been added which is similar to the sd_localfs() function. It uses the /sys/class/ubi/ubi?/ubi?_? path to search for UBIFS file system partitions, instead of the output from blkid, and then looks on those file systems for the default image filenames.

In a similar manner to sd_localfs(), skip_parts_arch() can be used to skip partitions on an architecture by architecture basis.

Testing has been done on a Dell n3048ep platform by copying over the 3 modified files and starting ONIE discovery.

Signed-off-by: Scott Emery <scotte@cumulusnetworks.com>
Closes: #82


ehdoyle:
Code looks clean and doesn't seem to have impacted build of other platforms at all. Thanks for the commit and the bug closeout.

PR #883, opened 2019-12-06
Accton AS7916_80XK: add to support ONIE

david56:
The machine has 80x40GbE/100GbE QSFP28 ports on Broadcom Jericho+ switch


ehdoyle:
Everything built clean and the commit message looks good. Thanks for the commit!

PR #884, opened 2019-12-06
Accton AS9716_32D: add to support ONIE

david56:
The machine has 32x400GbE QSFP-DD ports on Broadcom Tomahawk3 switch


ehdoyle:
Code built all targets without errors. Thanks for the commit!

PR #886, opened 2019-12-13
[Juniper][QFX5210] Adding platform uninstall hook

ciju-juniper:
There are primary and secondary bios in qfx5210 platform.
There is a problem with bios which prevents the OS booting
from the secondary bios when the OS was installed using
primary bios. Secondary bios fails to detect the UEFI
partition. Right now the workaround is to have a folder
structure /EFI/BOOT/BOOT64x.efi. While uninstalling the NOS,
onie should be aware of this new UEFI entry created.
This patch adds the logic to delete this extra UEFI entry.

Signed-off-by: Ciju Rajan K <crajank@juniper.net>
Signed-off-by: Balasubramanyam A <balasubramaa@juniper.net>


ehdoyle:
Everything built cleanly. Thanks for the fix!

PR #889, opened 2019-12-23
Accton AS4222_28PE: add to support ONIE

david56:
The machine has 48x1GbE SFP ports on Broadcom Hurricane2 switch
ASIC.  CPU module is Intel Rangeley C2538.


ehdoyle:
Built without issues. Thanks for the commit!

PR #890, opened 2019-12-23
Accton MiniPack: add to support ONIE

david56:
The machine has 128x100GbE QSFP28 ports on Broadcom Tomahawk3 switch
ASIC.  CPU module is Intel Xeon D-1527.


ehdoyle:
Built all targets without any problems. Thanks for the commit!

PR #892, opened 2020-01-02
Accton AS5916_54XM: revise to support EEPROM at 0x57

david56:


ehdoyle:
Looks good. Thanks for the commit!

Issue #893, opened 2020-01-05
Y2.02K bug - master branch builds broken for 2020


ehdoyle:
#  The problem  
A clean checkout of ONIE's master branch  will not build in 2020, as the date generated for LSB_RELEASE_TAG in the top level Makefile causes link failures in the build of the xtools target, specifically Crosstool-ng.
#  The workaround  
1.  From the build-config directory
2.  echo "mybuildversion" > conf/onie-release
This will define ONIE_RELEASE_TAG to not have a timestamp in it
Release branches do not require this workaround as they already come with a conf/onie-release file
#  Error output:
INFO ]  =================================================================
[INFO ]  Installing pass-2 core C gcc compiler
[INFO ]  Installing pass-2 core C gcc compiler: done in 278.98s (at 16:03)
[INFO ]  =================================================================
[INFO ]  Installing C library
[INFO ]    =================================================================
[INFO ]    Building for multilib 1/1: ''
[ERROR]      collect2: error: ld returned 1 exit status
[ERROR]      make[2]: *** [libc/Makefile.in:77: lib/libc.so] Error 1
[ERROR]
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Building for multilib 1/1: '''
[ERROR]  >>        called in step 'Installing C library'
[ERROR]  >>        called in step '(top-level)'
[ERROR]  >>
[ERROR]  >>  Error happened in: CT_DoExecLog[scripts/functions@338]
[ERROR]  >>        called from: do_libc_backend_once[scripts/build/libc/uClibc.sh@222]
[ERROR]  >>        called from: CT_IterateMultilibs[scripts/functions@1760]
[ERROR]  >>        called from: do_libc_backend[scripts/build/libc/uClibc.sh@82]
[ERROR]  >>        called from: do_libc[scripts/build/libc/uClibc.sh@63]
[ERROR]  >>        called from: main[scripts/crosstool-NG.sh@653]
[ERROR]  >>
[ERROR]  >>  For more info on this error, look at the file: 'build.log'
[ERROR]  >>  There is a list of known issues, some with workarounds, in:
[ERROR]  >>      '/home/adoyle/ONIE/badonie/build/crosstool-ng/crosstool-ng-1.23.0/docs/B - Known issues.txt'
[ERROR]  >>
[ERROR]  >>  If you feel this is a bug in crosstool-NG, report it at:
[ERROR]  >>      https://github.com/crosstool-ng/crosstool-ng/issues/
[...]

PR #894, opened 2020-01-06
Remove year from LSB_RELEASE_TAG to avoid Y2.02K bug

ehdoyle:
Ticket 893 https://github.com/opencomputeproject/onie/issues/893
has a more thorough write up on this.

The summary is that the date format:  +%Y%m%d%H%M
Puts a 2020 in for the year (as one would expect) but when this string
is used to name the tools generated by crosstools-ng, configuration
changes and linker failures occur.

The fix is to remove the date from the year, as the second pair
of bytes will never be more than '12', while still providing enough
time related context to create unique values.

And really, in practice, that's probably not really needed at all.

Branched code will have a build-config/conf/onie-release file that
has a shorter version string in it ( 2019.11br, for example ) that
does not suffer from this issue. However, if a historical build is
checked out and the onie-release file removed, the LSB_RELEASE_TAG
gets auto generated for master, and old builds will fail the same
way the latest ones do - so this bug has been around for forever,
and it's just being hit in 2020.

Tested enabling/disabling of the fix with kvm_x86_64 build
Closes: #893


ehdoyle:
Built master branch kvm without issue.

Issue #697, opened 2018-02-06
Enhancement request: DHCP option 61 support 


basilsaji-rtb:
We have a requirement where we want ONIE's dhcp request to contain option 61(with serial number of the device) so that we can use the serial number in the configuration of the dhcp server for provisioning the NOS images instead of the mac address that is currently available. Kindly help with the same.

cbrune:
DHCP requests are made by the busybox `udhcpc` client:
```
ONIE:/ # udhcpc --help                                                                                                       
BusyBox v1.25.1 (2018-01-30 16:43:57 PST) multi-call binary.                                                                 
                                                                                                                             
Usage: udhcpc [-fbqvRB] [-a[MSEC]] [-t N] [-T SEC] [-A SEC/-n]                                                               
        [-i IFACE] [-s PROG] [-p PIDFILE]                                                                                    
        [-oC] [-r IP] [-V VENDOR] [-F NAME] [-x OPT:VAL]... [-O OPT]...                                                      
                                                                                                                             
        -i,--interface IFACE    Interface to use (default eth0)                                                              
        -s,--script PROG        Run PROG at DHCP events (default /usr/share/udhcpc/default.script)                           
        -p,--pidfile FILE       Create pidfile                                                                               
        -B,--broadcast          Request broadcast replies                                                                    
        -t,--retries N          Send up to N discover packets (default 3)                                                    
        -T,--timeout SEC        Pause between packets (default 3)                                                            
        -A,--tryagain SEC       Wait if lease is not obtained (default 20)                                                   
        -n,--now                Exit if lease is not obtained                                                                
        -q,--quit               Exit after obtaining lease                                                                   
        -R,--release            Release IP on exit                                                                           
        -f,--foreground         Run in foreground                                                                            
        -b,--background         Background if lease is not obtained                                                          
        -S,--syslog             Log to syslog too                                                                            
        -a[MSEC],--arping[=MSEC] Validate offered address with ARP ping                                                      
        -r,--request IP         Request this IP address                                                                      
        -o,--no-default-options Don't request any options (unless -O is given)                                               
        -O,--request-option OPT Request option OPT from server (cumulative)                                                  
        -x OPT:VAL              Include option OPT in sent packets (cumulative)                                              
                                Examples of string, numeric, and hex byte opts:                                              
                                -x hostname:bbox - option 12                                                                 
                                -x lease:3600 - option 51 (lease time)                                                       
[...]

PR #895, opened 2020-01-14
Add client ID to udhcp parameters.

xfbs:
Sends the serial number of the device in the DHCP packet as option 61 (Client ID).

Untested. Addresses issue #697.

### DIFF
--- a/rootconf/default/lib/onie/functions
+++ b/rootconf/default/lib/onie/functions
@@ -401,6 +401,8 @@ udhcpc_args()
     local udhcp_args="-q -S -V onie_vendor:${onie_platform}"
     # user_class - option 77 - string in hex
     local udhcp_user_class="-x 77:$(str2hex onie_dhcp_user_class)"
+    # client id - option 61 - serial number in hex
+    local udhcp_client_id="-x 61:00$(str2hex $(onie-sysinfo -s))"
     # vendor specific options - option 125 - string in hex
     # code 3 - machine
     # code 4 - CPU architecture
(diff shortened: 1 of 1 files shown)


ehdoyle:
Hi - I just finally had a chance to review this. So, two questions:
1 - I appreciate the clarity of the 'untested', could you elaborate on the extent to which this has been verified, or software used to check that the requests were formatted as expected?
2 - The other thing I'm not quite clear on, is the setting of the length byte (and not immediately having access to a setup, I'm parsing the code rather than just running it and seeing what it does )
To elaborate:
It looks like str2hex does a straight ascii-hex encoding, so serial number "foo" would encode as 
666f6f
The string passed to udhcpc then becomes:
-x 61:00666f6f
The spec ( https://www.ietf.org/rfc/rfc2132.txt , section 9.14 ) has a length parameter (n )
Code   Len   Type  Client-Identifier
   +-----+-----+-----+-----+-----+---
   |  61 |  **n**  |  t1 |  i1 |  i2 | ...
   +-----+-----+-----+-----+-----+---
...and the documentation on how udhcpc handles things is a bit thin, as it's part of busybox.
( https://manpages.debian.org/stretch/udhcpc/udhcpc.8 )
However, I'd suspect  then that udhcpc inserts the length byte after parsing the argument?

Thanks!
-Alex

ehdoyle:
Confirmed during ONIE status call that the length issue is handled. Code merged cleanly. Thanks for the clarification and commit !

Vincent-Chang-Ufi:
> Sends the serial number of the device in the DHCP packet as option 61 (Client ID).
> 
> Untested. Addresses issue #697.

Hi I am wondering that if ONIE has multiple interface (ex : eth0,eth1) , and both request the ip from DHCP server through same serial number. eth0 and eth1 will get the same ip, it exist the conflict. Anyway can workaround ? Please see the attached Picture below :
eth0_eth1_same_ip

PR #896, opened 2020-01-16
Remove 'realpath' transitional debian package.

xfbs:
realpath is a coreutils transitional package. According to the package
description, it can be safely removed and exists only to facilitate
upgrades.


ehdoyle:
Just checked and it looks like the realpath binary is supplied by the coreutils package in Debian 8/9/10, so that can be  cleaned up - I'll add this PR shortly.
Thanks!

ehdoyle:
Confirmed there was (as expected) no effect on build.
Thanks for the commit!

PR #897, opened 2020-01-16
Fixes formatting of RFC3004 user class data

xfbs:
The RFC mandates that individual user class data items are prefixed with their length (in addition to the option length field).

See https://tools.ietf.org/html/rfc3004 for reference.

This causes issues with the DHCP packet, as seen in Wireshark:

Screenshot 2020-01-16 at 20 42 14

### DIFF
--- a/rootconf/default/lib/onie/functions
+++ b/rootconf/default/lib/onie/functions
@@ -388,6 +388,13 @@ str2hex()
     echo -n "$1" | hexdump -ve '1/1 "%.2x"'
 }
 
+# turn string into hex with prefixed length (as hex byte)
+str2lenhex()
+{
+    local str="$1"
+    printf "%02X%s" ${#str} $(str2hex $str)
+}
+
 make_str_opt()
 {
     local code="$1"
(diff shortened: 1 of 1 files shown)


ehdoyle:
Hi - the fix looks correct to me - thanks for including the RFC for context.
Question: Is a hex dump of the wireshark capture reflecting that the length byte isn't set ?
I'd just like to confirm that udhcpc doesn't try to put a length byte in that field on transmission ( see my question for ONIE PR 895 ( https://github.com/opencomputeproject/onie/pull/895 ) )
The reason I ask is that there was an issue with older versions of Wireshark handling formatting that did not conform to the spec.
( https://osqa-ask.wireshark.org/questions/35332/dhcp-option-77-malformed-option )  and I'd just like to make sure the malformed error shows it parsing the first character of the string rather than the length byte that should be there  ( and also to provide some additional history to provide context should this issue ever get revisited )
Thanks!
-Alex

ehdoyle:
Confirmed formatting validation on the ONIE status call. Nice catch!
Thanks for the fix.

PR #898, opened 2020-01-22
Alphanetworks STX-60D0-062F: Modify the console setting for new BIOS version.

talisych:
Modify the console setting for new BIOS (Version 20191003 and later version).


ehdoyle:
Confirmed that the platform rebuilt without problems. (Hardly surprising given the scope of the changes, but one never knows...)
Thanks for the commit!

PR #902, opened 2020-02-27
v08 for S9700_53DX, S9700_23D, and S9705_48D

jaytclin:
1. Add IPMI support 2. Add firmware update tool support 3. Modify kernel I/O delay type


ehdoyle:
All platforms built cleanly. Thanks for the update!

PR #908, opened 2020-04-16
Marvell a7020-comexp - add support for common CPU module

gilt-marvell:
Signed-off-by: Gil Tabibian <gilt@marvell.com>


ehdoyle:
Hi Gil - the pull request looks good apart from one issue I had a question about.
In onie/rootconf/default/etc/init.d/gen-config.sh 
The line:
-    local onie_platform="${onie_arch}-${onie_machine}-r${onie_machine_rev}"
got changed to:                
+    local onie_platform=${onie_platform:-${onie_arch}-${onie_machine}-r${onie_machine_rev}}
I was curious if this was absolutely needed, since that code is shared by other platform builds.
Thanks,
-Alex

gilt-marvell:
Hi Alex,
thanks for the quick review.
with current implementation, there is no way to update 'onie_platform' by generic rootconf script ('gen-config-platform'), because 'gen-config.sh' will always run over any local 'onie_platform' setting.

for example in my case, I have common CPU for several boards, so I want to update 'onie_platform' 
with this script - rootconf/sysroot-lib-onie/gen-config-platform.
it works fine, but then generic rootconf script (rootconf/default/etc/init.d/gen-config.sh) not taking in consideration if this setting already initialized before.

therefore, I did this check to verify if there was any configuration for 'onie_platform' before, if not - the default will be set.
let me know what do you think.

ehdoyle:
Hi Gil
 Ok - read through that a few times along with the code, and got it. I'll add a comment noting that the update is to set onie_platform if it is not already set, in the same way that onie_machine is set above it. Thanks!

PR #913, opened 2020-05-05
Netberg nba610: Fixed config file instead of erroneous junk

andrewsapronov:
Hello, 

Please merge for fix erroneous config file. It was some binary junk in it :(


ehdoyle:
Well the new version is certainly human readable via hexdump whereas the prior version wasn't, so although I can't test it I have to imagine it works better :)
Thanks for the commit!

PR #914, opened 2020-05-06
Accton EBOF: add to support ONIE

david56:
The machine has 24x200GbE Fabric ports on Marvell Falcon 98CX8520
switch ASIC.  CPU module is Intel Denverton C3558.


ehdoyle:
Everything built cleanly. I've also added it to the onie-build-targets.json file to 
 - confirm that it does build
 - it built with the Debian 9 DUE container
 - and to create an automated build entry when the ONIE Jenkins I'm using gets configured.
 Thanks for the commit!

PR #917, opened 2020-06-05
Add a generic ONIE for BROADWELL-DE CPU board.

JohnsonYJLu:
Add a generic ONIE for BROADWELL-DE CPU board.

Signed-off-by: johnson <JOHNSON.LU@deltaww.com>


ehdoyle:
The platform built 'all' and 'demo' targets cleanly. I've added it to the onie-build-targets.json file so it will be included in future automated builds.
Thanks for the commit!

PR #920, opened 2020-06-10
New Platform is created for AGCV208S

graceyliu:
AGCV208S is created and check in.


ehdoyle:
Looks good! Platform built all and demo targets without errors.
Thanks for the commit!

PR #923, opened 2020-07-07
Add KEXEC_FILE to x86_64 kernel config.

mshych:
This config allows calling kexec tool with option -s (call to kexec_file_load
systemcall). It can be required in NOS install from UEFI booted ONIE.

Signed-off-by: Michael Shych <michaelsh@mellanox.com>


billie-alsup:
Does this allow one to bypass secure boot (and boot an unsigned kernel)?  If so, why put it in the common options?  Shouldn't it be a platform specific decision?

mshych:
The change is not related to secure boot.
Enabling KEXEC_FILE adds “CONFIG_KEXEC_VERIFY_SIG is not set”.
This is a reason why it’s appeared in the commit.
Currently, disabled KEXEC_VERIFY_SIG option can be used in the future in secure boot.
Change to CONFIG_KEXEC_VERIFY_SIG=y  should be done on machine level

CONFIG_KEXEC_FILE allows to use another mode of calling user space kexec tool
Through kexec -s option. Kexec tool can be used also in old mode if new system call isn’t required.
SMBIOS tables, ACPI tables as well as EFI data isn’t passed to new kernel after
call of kexec with old system call (kexec_load)  on ONIE installed in UEFI mode.
It was indeed done initially at the kernel config change at the specific machine level
However, I think it’s common option that doesn’t affect old mode of kexec usage but
can be useful also on other platforms.

Thank you,
    Michael.

billie-alsup:
thanks for the info

ehdoyle:
Hi - since the kernel has upgraded in the master branch, and changing this here would affect all platforms, I'm going to close this out and leave it as an option for the machine targets to enable in their kernel configs. Thanks!

ehdoyle:
Closing out after comment period.

PR #935, opened 2020-12-04
[DELTA] Add new platform delta agc7008s

JohnsonYJLu:
Add new platform delta agc7008s

Signed-off-by: johnson <JOHNSON.LU@deltaww.com>


ehdoyle:
Code built all targets without any problems. Thanks for the commit!

PR #936, opened 2020-12-18
Add new platform: delta_tx4810

vincent0083:
The machine has 48x10GbE SFP+ ports on a Mellanox Spectrum switch ASIC.
    The CPU module is a Marvell amada 88F7040.


ehdoyle:
Confirmed that the new platform built clean using a DUE Debian 9 ( Stretch ) build environment.
Thanks for the commit!

PR #938, opened 2021-01-11
Delta tx4810: disable unused ethernet interfaces in kernel device tree

vincent0083:
Signed-off-by: os-vincent.liu <vincent.liu@deltaww.com>


paulmenzel:
Looks good to me. Thank you. Two nits:

1.  Instead of *os-vincent.liu* could you use *firstname lastname*?
2.  Does Linux currently print an error?

ehdoyle:
Looks good to me. I adjusted the username in the commit.
Note that there is no demo os build for arm64, (which is not a requirement at this stage).
Thanks for the commit!

PR #941, opened 2021-01-20
Accton X86_64: update management port driver

david56:


ehdoyle:
Hi - can you suggest a platform (or platforms) to build with the update so I can verify everything compiles cleanly? Thanks!

david56:
@ehdoyle - The platforms are accton_as5916_54xks and accton_as5916_54xl.  Thank you very much.

paulmenzel:
Can you please add a commit message describing the problem, and the implementation of the fix?

david56:
@paulmenzel , the commit message had been upgraded.  Please check it.  Thank you.

paulmenzel:
(Please note, that Open Network Linux currently seems unmaintained, and the last merge/pull requests or patches were accepted several months ago. No idea, if this pull/merge request is going to be submitted. (I do *not* have the corresponding permissions.))

Thank you for updating the commit message. As you list two items (two different drivers), could you please split the commit into two? (`git reset HEAD^` and then `git commit -p -c 5327f151a6e127cb49e72c7bec758da835f7db79` for example (two times).)

Also, are these changes in the upstream Linux kernel?

david56:
> could you please split the commit into two?

Done.



> Also, are these changes in the upstream Linux kernel?

Nope.  These changes are not in the upstream.

ehdoyle:
Hi - apologies for the delay in getting to this. The holiday break has given me some breathing room to start looking at the pull request backlogs.
**Summary** 
If it is ok to upgrade the compiler version, I can pull this in, as it will build. However as I have no way to run the image, I can't know that it worked, so  @david56 or someone else from Accton will have to verify that it works, and sign off on it.
The compiler upgrade fix may work for the other platforms mentioned for test build:
accton_as5916_54xks and accton_as5916_54xl

**Additional information**
The as5916_26xb doesn't build on the master branch after component upgrades were introduced.
However, if the GCC_VERSION is set to 8.3.0 in the [machine.make](https://github.com/opencomputeproject/onie/blob/8409e9543b784037ec0b468386c35ff9d3d7f6f4/machine/accton/accton_as5916_26xb/machine.make#L38)
Then it will build all and demo targets on the master branch
I can't tell you if those actually run, as I don't have any way of testing it, but it looks promising.
The pull request applies cleanly to master with the above changes, and I built the as5916_26xb without any problems, so if the compiler upgrade still produces a working image I think that would sort this out.

PR #949, opened 2021-04-01
Strip trailing whitespace from `onie-build-targets.json`

paulmenzel:


ehdoyle:
Cleaned up. Thanks for the commit!

paulmenzel:
Thank you for the quick action.

Out of curiosity, why aren’t you just accepting the merge/pull request but manually cherry-pick the commit?

ehdoyle:
Hi - it's a workflow consistency thing. Sometimes submitted commits need to be squashed to present as one functional change, or I'll do some cleanup on the comments for clarity, and in those cases I can just handle it without having to ask for a resubmission, and it'll have me down as having signed off on it.  If it doesn't need any of that, then the same workflow applies -it just gets rubber stamped and committed.

paulmenzel:
I split the discussion into the separate issue https://github.com/opencomputeproject/onie/issues/952. Sorry for mixing this in the first place.

PR #950, opened 2021-04-01
Fix build of delta platforms

paulmenzel:
Building *delta_tn48m* and *delta_tx4810* fails on a GNU/Linux system with GNU patch 2.7.5 and StGit 1.0.


paulmenzel:
Please accept the merge/pull request, so it’s tagged as *Merged* and not cherry-pick the commits yourself and close it.

ehdoyle:
Hi - The commit for each platform makes sense to me.  For sanity checks I buiilt both with and without the pull requests. Both built fine in the DUE Debian 9 container (I was afraid I'd missed something there). I can't verify the build environment you're using, but if this fixes them there, I'm fine with the change since it does not seem to have a negative impact on existing builds.
I also appended an update to the onie-build-targets.json file to indicate that I'd rebuilt both platforms with the latest release.
Thanks for the commit!

Issue #946, opened 2021-03-09
dropbear version is very outdated


user318:
Hello,

I suggest to update the version of dropbear. The version in ONIE is 2016.74, which is almost 5 years old. Current version 2020.81 have better support for ciphers, etc. For example it has support for ed25515 keys and chacha20-poly1305 cipher.
I have attached some patch, that I made for myself.
[onie-dropbear-update.patch.txt](https://github.com/opencomputeproject/onie/files/6111603/onie-dropbear-update.patch.txt)
Couple of words about what is there.
It works for me, but it needs some additional work, I think. Because I have disabled there some ciphers and dss, but those may be required by others.
Dropbear now have different structure of included config file, so you do not need to copy all the defaults there and it needs to be placed as localoptions.h. The defines also have 0/1 values now instead of being set or not.
I have added "--disable-harden" to configure, because otherwise the build procces was breaking with an error about zlib was missing. But in config.log I saw it because ld complaint about some ssp library. Do not know exactly what it is. I suggested it is related to stack protection and found out the "--disable-harden" option and it worked out.

paulmenzel:
If you have a patch, it’d be great if you submitted a merge/pull request.

user318:
Ok, I'll do it. Just, as I said, I am not sure my solution is good for general use, so I decided just to show what I mean.

ehdoyle:
Closing out this bug as the pull request has been merged and seems to be working well.
[here](https://github.com/opencomputeproject/onie/commit/00a3761fdd891ce8ac90b07abdfa66bdac302dfc)

PR #955, opened 2021-04-09
Update dropbear to 2020.81, change its config: remove dss, add ed2551…

user318:
…9 keys, harden ciphers set

Changes applied it this patch:
- use modern dropbear version
- new version has separate config header file for customizations
- parameters in config are now set with 0/1 value instead undefined/defined
- update sets of ciphers, hashes, sighature algorithms, etc. to the stronge ones
- remove ssh host dss key and add ed25512 key in config and scripts
- add "--disable-harden" to dropbear compile options for it to succeed

### DIFF
--- a/build-config/conf/dropbear.config.h
+++ b/build-config/conf/dropbear.config.h
@@ -1,356 +1,69 @@
-/* Dropbear SSH
- * Copyright (c) 2002,2003 Matt Johnston
- * All rights reserved. See LICENSE for the license. */
+#ifndef DROPBEAR_LOCAL_OPTIONS_H_
+#define DROPBEAR_LOCAL_OPTIONS_H_
 
-#ifndef DROPBEAR_OPTIONS_H_
-#define DROPBEAR_OPTIONS_H_
+#define NON_INETD_MODE 1
+#define INETD_MODE 0 /* -- ONIE not set */
 
-/* Define compile-time options below - the "#ifndef DROPBEAR_XXX .... #endif"
- * parts are to allow for commandline -DDROPBEAR_XXX options etc. */
+#define DROPBEAR_SMALL_CODE 0
 
-/* IMPORTANT: Many options will require "make clean" after changes */
+#define ENABLE_X11FWD 0 /* -- ONIE not set */
 
-#ifndef DROPBEAR_DEFPORT
-#define DROPBEAR_DEFPORT "22"
-#endif
+#define DROPBEAR_CLI_LOCALTCPFWD 1
+#define DROPBEAR_CLI_REMOTETCPFWD 1
 
-#ifndef DROPBEAR_DEFADDRESS
-/* Listen on all interfaces */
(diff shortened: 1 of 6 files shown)


user318:
This pull request is for the issue #946

ehdoyle:
Thanks! It's going to be a bit for this pull request as I'll be doing platform rebuilds to make sure nothing breaks, and work has been very successful in keeping me from being bored these days :)

ehdoyle:
Hi - just tried running this using the kvm build.  The code built fine, but it hangs after image install is selected from Grub and  dropbear starts.
 Any thoughts?
Here's the output:
`ONIE: OS Install Mode ...
Platform  : x86_64-kvm_x86_64-r0
Version   : HEAD-05142357
Build Date: 2021-05-14T23:57+00:00
Info: Mounting ONIE-BOOT on /mnt/onie-boot ...
Info: Mounting EFI System on /boot/efi ...
Info: BIOS mode: UEFI
Running demonstration platform init pre_arch routines...
Running demonstration platform init post_arch routines...
Info: Making NOS install boot mode persistent.
network_driver: Running demonstration pre_init routines...
network_driver: Running ASIC/SDK init routines...
network_driver: Running demonstration post_init routines...
Info: Using eth0 MAC address: 52:54:00:13:34:21
Info: eth0:  Checking link... up.
Info: Trying DHCPv4 on interface: eth0
ONIE: Using DHCPv4 addr: eth0: 10.0.2.15 / 255.255.255.0
Starting: klogd... done.
Starting: dropbear ssh daemon...

`

user318:
I can guess that the cause is the lack of entropy. Have you tried to run qemu with option: `-device virtio-rng-pci`? It it boots faster with it, than it freezes on accessing /dev/random when there is little entropy in the kernel, which is the case for the virtual machine.
I have looked through the source of dropbear, and through my changes. I see that I have dropped the `#define DROPBEAR_DELAY_HOSTKEY` from `dropbear.config.h`, most probably it it the source of the problem. I've also found that libtomcrypt in dropbear can use `/dev/random` if `LTC_TRY_URANDOM_FIRST` is not set. And I do not find where it could be set with current config. So the other option is to try to add `#define LTC_TRY_URANDOM_FIRST` to `dropbear.config.h`.

ehdoyle:
Hi - sorry about the delay here. With the master branch finally upgraded I finally got back to this, and I'm not seeing the issue I'd seen previously. It built fine for the kvm_x86_64 and newly updated qemu_armv8a virtual machines.
I tested it by ssh'ing to a remote host and to 127.0.0.1.
The x86 target had no issues
The ARM64 target doesn't recognize the user when connecting to the loopback address and wants a password, but a build with the previous version of Dropbear had exactly the same problem, so I do not believe upgrading Dropbear breaks anything. and that the problem lies elsewhere in the configuration of the arm64 virtual machine.
Thanks for the commit!

PR #960, opened 2021-05-13
Fix the wrong link

nienjen:
The original link "https://www.kernel.org/doc/Documentation/email-clients.txt" is not work now, so the new link is updated in this pull request.


ehdoyle:
Thanks for the fix - documentation is one of those things that always seems to get the least amout of attention.
Since it is a markdown file, I took the opportunity to turn the updated URL you tracked down into a hyperlink.
Thanks for the commit!

PR #956, opened 2021-04-26
[armv8a] fix bug for compile qemu armv8a image for issues 940

LuiSzee:
Fix bug for compile qemu armv8a image for issues 940

build:
cd onie/build-config
make -j4 MACHINE=qemu_armv8a all demo

run:
qemu-img create -f qcow2 onie-aarch64-demo.img 1G
wget http://releases.linaro.org/components/kernel/uefi-linaro/16.02/release/qemu64/QEMU_EFI.fd
cat QEMU_EFI.fd /dev/zero | dd iflag=fullblock bs=1M count=64 of=flash0.img
dd if=/dev/zero of=flash1.img bs=1M count=64
qemu-system-aarch64 -machine virt -m 1024M -cpu cortex-a57 -nographic -drive if=none,file=onie-aarch64-demo.img,id=hd0 -device virtio-blk-device,drive=hd0 -cdrom onie-recovery-arm64-qemu_armv8a-r0.iso -pflash flash0.img -pflash flash1.img

PR #970, opened 2021-08-19
Emulation support for ARM64 with qemu_armv8a

### LINKED ISSUE #940: ONIE 2020.11 fails to build with `MACHINE=qemu_armv8a`
Attempting to build ONIE 2020.11 in a Debian Stretch build container for the `qemu_armv8a` machine results in build failure. Build fails during kernel build as it fails to find `qemu_armv8a.dtb`:

```
make[2]: *** No rule to make target 'arch/arm64/boot/dts/qemu_armv8a.dtb'. Stop.
arch/arm64/Makefile:94 recipe for target 'qemu_armv8a.dtb' failed
```

Those are the most relevant sections of the output, and that's where things get interesting. Taking a look through [onie/machine/qemu_armv8a](https://github.com/opencomputeproject/onie/tree/master/machine/qemu_armv8a) gives the impression this is not actually a supported machine and is incompletely implemented. The kernel config is empty as is the u-boot config, which I suspect should at least both have _something_ in them as they're populated with content for [qemu_armv7a](https://github.com/opencomputeproject/onie/tree/master/machine/qemu_armv7a).

I also perused kernel sources to attempt to locate a device tree file for `qemu_armv8a` and struck out. It seems like there never has been nor is a `qemu_armv8a.dts` in the kernel tree. The device tree isn't included in a patch like it is for [qemu_armv7a](https://github.com/opencomputeproject/onie/blob/master/machine/qemu_armv7a/kernel/platform-qemu_armv7a.patch) again leading me to believe that support for this machine is not really implemented.

Can anyone confirm whether ONIE should in fact build for machine type `qemu_armv8a`?

ehdoyle:
This fixes build problems with the qemu_armv8a virtual machine for ARM64, and adds it to the onie/emulation/onie-vm.sh script so that it can be run through the same interface as the kvm_x86_64 target.

This also upgrades a number of components for compatibility reasons.

It does have the side effect of ARM building with parted version 3.4 and AMD64 using version 3.1, which will only cause an error if the ARM64 target is built in the same ONIE checkout as teh AMD64 one. ( In which case, cleaning for the existing version will allow the other to download )

Fixes #940 
Built off of code sumbitted in PR #956 by Shi Lei <shil@centecnetworks.com>


ehdoyle:
Self approved.

Issue #972, opened 2021-10-06
/home/build/src/onie/build/crosstool-ng/crosstool-ng-1.24.0/ct-ng:261: recipe for target 'build' failed


squeeb:
When building the `all` target for `MACHINE=kvm_x86_64` within the dockerised Debian build environment, the build eventually fails with the following errors: 

```
[INFO ]  =================================================================
[INFO ]  Installing C library headers & start files
[INFO ]    =================================================================
[INFO ]    Building for multilib 1/1: ''
[ERROR]      make[3]: *** [Makefile.in:164: ../..//extra/locale/c8tables.h] Error 1
[ERROR]      make[2]: *** [Makefile.in:185: headers] Error 2
[ERROR]
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Building for multilib 1/1: '''
[ERROR]  >>        called in step 'Installing C library headers & start files'
[ERROR]  >>        called in step '(top-level)'
[ERROR]  >>
[ERROR]  >>  Error happened in: CT_DoExecLog[scripts/functions@376]
[ERROR]  >>        called from: uClibc_backend_once[scripts/build/libc/uClibc.sh@141]
[ERROR]  >>        called from: CT_IterateMultilibs[scripts/functions@1586]
[ERROR]  >>        called from: uClibc_backend[scripts/build/libc/uClibc.sh@36]
[ERROR]  >>        called from: uClibc_start_files[scripts/build/libc/uClibc.sh@10]
[ERROR]  >>        called from: do_libc_start_files[scripts/build/libc.sh@28]
[ERROR]  >>        called from: main[scripts/crosstool-NG.sh@696]
[ERROR]  >>
[ERROR]  >>  For more info on this error, look at the file: 'build.log'
[ERROR]  >>  There is a list of known issues, some with workarounds, in:
[ERROR]  >>      '/home/build/src/onie/build/crosstool-ng/crosstool-ng-1.24.0/docs/manual/B_Known_issues.md'
[ERROR]  >>
[ERROR]  >> NOTE: You configuration uses non-default patch sets. Please
[ERROR]  >> select 'bundled' as the set of patches applied and attempt
[ERROR]  >> to reproduce this issue. Issues reported with other patch
[ERROR]  >> set selections (none, local, bundled+local) are going to be
[ERROR]  >> closed without explanation.
[ERROR]  >>
[ERROR]  >>  If you feel this is a bug in crosstool-NG, report it at:
[...]

PR #973, opened 2021-12-02
fix: add unzip to build tools

cummings-chris:
While toying around with getting the recovery-iso to build, I found that the make file build process depends on `unzip` however, this is not included in the docker image for the onie build environment and is not installed by the debian-prepare-build-host make target. This PR has two tiny commits to add the unzip package when using those tools.

### DIFF
--- a/build-config/Makefile
+++ b/build-config/Makefile
@@ -597,7 +597,7 @@ DEBIAN_BUILD_HOST_PACKAGES	= build-essential stgit u-boot-tools util-linux \
 				  libexpat1-dev fakeroot python-sphinx rst2pdf \
 				  libefivar-dev libnss3-tools libnss3-dev libpopt-dev \
 				  libssl-dev sbsigntool uuid-runtime uuid-dev cpio \
-				  bsdmainutils
+				  bsdmainutils unzip
 
 PHONY += debian-prepare-build-host
 debian-prepare-build-host:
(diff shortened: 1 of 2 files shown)


cummings-chris:
Additionally, to get this to build, I had to follow the steps outlined in #972. I'd be glad to add some updates to the docs for the build procedure if that would be helpful, but it looks like there might already be an effort underway to update the docs, so I don't want to duplicate effort unnecessarily. 

Cheers!

ehdoyle:
Hi Chris. thanks for the pull request and I'm good with  adding it. For tracking purposes can you give me a replication scenario for where the fix was needed and worked, like
Debian release version you're building in
Machine target you're trying to build.
That'll make it easier to draw a line between where and why it wasn't working, and demonstrate that the fix resolves it.

As an aside, I've been building ONIE in Docker containers using [Dedicated User Environment](https://github.com/CumulusNetworks/DUE/tree/master/templates/onie) for the past few years. The container approach is really handy if you have to use a build environment that doesn't match your host system's OS version ( Ex: the master branch builds for Debian 10, but currently all machine targets build in a Debian 9 environment with the 2021:08 ONIE release)
Thanks!
Alex Doyle - ONIE Project lead.

cummings-chris:
Hi Alex! Thanks for the response. Specifically I ran into this both on my Debian 10 workstation and using the onie/crontrib/build-env docker build container (which I believe is based on Debian 9). 

The specific build target that I was trying were `make MACHINE=kvm_x86_64 all` and `make -j4 MACHINE=kvm_x86_64 all recovery-iso`. I did eventually get `make MACHINE=kvm_x86_64 all` to work in the docker container with `unzip` installed. 

Thanks for the pointer on DUE, I'll give this a shot!

ehdoyle:
Hi - just merged this in with both commits squashed and a bit of additional context in the commit message. Thanks for the commit!

PR #975, opened 2021-12-14
Add Alpha Networks SNJ62D0-640F:

SeanTai-ALPHA:
64 x 400G QSFP DD ports, Intel Denverton C3538 4 Core/2.1GHz.

Spec: 64 x 400G QSFP DD ports.
      Broadcom BCM56990(Tomahawk4) switch ASIC.
      Intel Denverton C3538 4 Core/2.1GHz.
      The default console device is ttyS0.

Signed-off-by: SeanTai-ALPHA <Sean_Tai@alphanetworks.com>


ehdoyle:
Built on the master branch in a Debian 10 container with zero issues. Beautiful, and this is the first new platform to work with the master code (even though it's using a 4.x kernel) since I broke older builds with component upgrade. 
Thanks for the commit!

PR #982, opened 2022-01-09
image.make: Add libnns_files.so to sysroot image on ARM64

felixradensky:
This library is required for ssh localhost login to work, otherwise
local users are not recognized.

Signed-off-by: Felix Radensky <fradensky@nvidia.com>

### DIFF
--- a/build-config/make/images.make
+++ b/build-config/make/images.make
@@ -180,7 +180,8 @@ SYSROOT_LIBS	= ld-$(XTOOLS_LIBC_VERSION).so \
 		  libutil.so.1 libutil-$(XTOOLS_LIBC_VERSION).so \
 		  libdl.so.2 libdl-$(XTOOLS_LIBC_VERSION).so \
 		  libpthread.so.0 libpthread-$(XTOOLS_LIBC_VERSION).so \
-		  librt.so.1 librt-$(XTOOLS_LIBC_VERSION).so
+		  librt.so.1 librt-$(XTOOLS_LIBC_VERSION).so \
+		  libnss_files.so.2 libnss_files-$(XTOOLS_LIBC_VERSION).so
   ifeq ($(ARCH),arm64)
     SYSROOT_LIBS	+= ld-linux-aarch64.so.1
   endif


ehdoyle:
Hi Felix,
 the code definitely works on arm64 for ONIE and the DemoOS for qemu_armv8a, and does not break anything for kvm_x86_64 ONIE or DemoOS, so it is good to add.
 A couple of questions though - 
 1 - can you elaborate a bit on what the library is doing so that can be added to the commit message to provide context for developers that aren't as familiar.
 2 - Any idea how the x86 version of the code was able to handle login without this library? I'd tested that when processing the pull request for the DropBear upgrade a few weeks back.  See comments at the end here:
 [DropBear Upgrade Pull Request](https://github.com/opencomputeproject/onie/pull/955)
 Thanks!

felixradensky:
Hi Alex,
libnss_files.so is part of glibc NSS implementation. For more details on NSS see https://en.wikipedia.org/wiki/Name_Service_Switch.
This particular shared object provides access to plain text database files like /etc/passwd and /etc/hosts. On ARM64, dropbear failed to identify local users since the only way to access /etc/passwd on glibc based systems is via libnss_files.so. On X86 the login worked since uclibc does not provide NSS implementation, as can be seen in https://uclibc.org/downloads/Glibc_vs_uClibc_Differences.txt

felixradensky:
Hi Alex,
There's a typo in the patch subject, it should be libnss_files.so instead of libnns_files.so. Feel free to modify the commit message as you see fit.
Felix.

ehdoyle:
Excellent - that's exactly what I was looking for.  One final thing - is there any reason this library should be part of the general inclusion and not ARM64 specific? Moving it one line down to be part of the ARM64 logic seems to work fine, and I think, makes it stand out a bit more as an arm64 specific change:
Ex:
`
	  libnss_files.so.2 libnss_files-$(XTOOLS_LIBC_VERSION).so
  ifeq ($(ARCH),arm64)
    SYSROOT_LIBS	+= ld-linux-aarch64.so.1
  endif
 `
becomes
`  ifeq ($(ARCH),arm64)
    SYSROOT_LIBS	+= ld-linux-aarch64.so.1   libnss_files.so.2 libnss_files-$(XTOOLS_LIBC_VERSION).so
  endif`
  
If that makes sense I can do the edit and push to resolve the pull request.  
Thanks again for solving this issue.

felixradensky:
Hi Alex,
Unlike ld-linux-aarch64.so.1 this library is not ARM64 specific, so I don't think it should be moved.
I was under impression that glibc is used only on ARM64, that's why I've mentioned ARM64 in the patch subject. Maybe it should be modified to mention glibc instead of ARM64.
Felix.

ehdoyle:
Ah, so the distinction is the library working with glbc (which is used on ARM64, and not AMD64) rather than just being for ARM64. That makes sense. Thanks!

PR #983, opened 2022-01-11
Add platform of WNC QSA63_OAM8_48

willchao6790:
WNC QSA63-OAM8-48 switch
* Solution: Marvell AC5
* Configuration: 48x1G + 4x10G


ehdoyle:
Everything built perfectly, and I've added the 4.14.76 kernel to the OpenCompute mirror in case something happens to the original site.
[linux-4.14.76.tar.xz](http://mirror.opencompute.org/onie/linux-4.14.76.tar.xz)
Thanks for the follow up and the commit!

PR #989, opened 2022-03-01
Update build targets for QuantaMesh T1048-LB9

danielkauffman:
`make MACHINEROOT=../machine/quanta MACHINE=quanta_lb9 all demo` succeeds with Debian 9 tools on branch 2021.08br but fails with Debian 10 tools on branch 2022.02br


ehdoyle:
Thanks for the commit  - I'd added a 'Signed-off-by' in the commit message, which is what GitHub was complaining about.

PR #1019, opened 2022-11-15
Accton AS7326_56X: revise to support EEPROM at 0x57

david56:


mshych (approved):
Approved.
Thank you for your commit.

PR #1032, opened 2023-04-25
CI for our ONIE builds

mheese:


mheese:
For anybody who is wondering what that was: Github by default unfortunately _always_ creates a PR on the original repository instead of on the fork. I did not intend to open the PR here.

PR #1033, opened 2023-04-26
Accton AS4625: add to support ONIE

david56:


mshych:
Merged to master.
Thank you for commit.

PR #1066, opened 2024-09-04
Append warning message to GRUB menu "ONIE: Install OS"

david56:
If you go into "ONIE: Install OS" option it will screw up the boot order of the box, and automatically always boot into ONIE instead of the pre-installed NOS. This is a problem and not user friendly, and there should be a warning message in the window, that alerts you that clicking that menu option will reset the boot order. The boot order will not boot into the NOS, until you install a new OS.

### DIFF
--- a/installer/grub-arch/grub/grub-common.cfg
+++ b/installer/grub-arch/grub/grub-common.cfg
@@ -10,7 +10,7 @@ set timeout=%%GRUB_TIMEOUT%%
 
 onie_submenu="ONIE (Version: $onie_version)"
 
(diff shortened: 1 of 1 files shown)


mshych:
Hi,
It can be not so good to make globally for all systems.
1. It's possible that some users already have automation and expect specific output.
2. This warning is not so relevant for systems with UEFI boot, where boot order is defined by efiboot entries.
You can overwrite grub config for your specific system, e.g. ./machine/<manufacturer>/<system>/rootconf/grub/grub-common.cfg

PR #1067, opened 2024-10-02
Fix  issue in the onie-release causing possible install error because it can't fit in the EEPROM

hannescam:


mshych:
Thank you for noting this.
It was corrected and pushed.

PR #1098, opened 2025-07-30
crosstool-NG: Add upstream patch to allow building with docker

cpackham-atlnz:
crosstool-NG wants to know the username but it isn't critical. Bring in an upstream patch that allows using docker containers started with the -u option to provide a numeric UID.


patches/crosstool-NG/Ignore-a-bad-id-command-to-allow-docker.patch
```diff
@@ -0,0 +1,26 @@
+From aa6a6c3a34b3e2c5b094855e8d2b57a698457561 Mon Sep 17 00:00:00 2001
+From: Quentin Boswank <qubos@outlook.de>
+Date: Mon, 5 Feb 2024 16:20:18 +0100
+Subject: [PATCH] Ignore a bad id command to allow docker
+
```
paulmenzel, line 5: Could you add the URL to the upstream patch in the commit message body, please?

mshych:
Pushed to master. Thank you.

PR #1110, opened 2026-02-24
Handle missing management MAC address in syseeprom

Benjamin-Davies:
When the management MAC address was missing from the syseeprom, /etc/init.d/networking.sh would attempt to parse the error message as a MAC address. This would result in a fairly cryptic error message:
```
/etc/rcS.d/S30networking.sh: line 1: arithmetic syntax error
```

This change explicitly handles that case by showing the user a warning and skipping steps that rely on this value.

### DIFF
--- a/rootconf/default/etc/init.d/networking.sh
+++ b/rootconf/default/etc/init.d/networking.sh
@@ -198,6 +198,10 @@ if [ "$1" = "start" ] ; then
 
     # Set MAC addr for all interfaces, but leave the interfaces down.
     base_mac=$(onie-sysinfo -e)
+    if [ "$base_mac" = "ERROR:" ] ; then
+        log_warning_msg "Unable to retrieve management Ethernet MAC address"
+        onie_skip_ethmgmt_macs=yes
+    fi
     for intf in $intf_list ; do
         if [ "$onie_skip_ethmgmt_macs" = "no" ] ; then
             mac="$(mac_add $base_mac $intf_counter)"


mshych:
It looks like this change is trying to solve a problem in system customization in production.
According to the current behaviour, onie-sysinfo -e should return the eth0 MAC address when 
the machine is defined with onie_skip_ethmgmt_macs=yes. The current processing is incorrect,
because onie-sysinfo -e does not return "ERROR" in this case, but "unknown".
Please also see the comments in onie-sysinfo:
"
# 1. is it set in the environment?  highest priority                                  
# 2. platform function provided                                     
# 3. architecture function provided                                 
# 4. use the contents of /sys/class/net/eth0/address                
# 5. return 'unknown'
"

Benjamin-Davies:
In our case, `machine-build.conf` specifies `onie_skip_ethmgmt_macs=no`. Because of this, `onie-sysinfo -e` calls `get_ethaddr_arch` which calls `onie-syseeprom -g 0x24` which returns an error as the value is missing from the syseeprom. This error is then passed to `mac_add` in [networking.sh line 207](https://github.com/Benjamin-Davies/onie/blob/6533201cd74b5b2e318783597f59c5c891b37695/rootconf/default/etc/init.d/networking.sh#L207), which fails with the arithmetic syntax error.

The point of this patch is to handle this error case. Even if `onie-sysinfo -e` did return "unknown", networking.sh line 207 would still fail as "unknown" is not a valid mac address.

Would you rather `onie-sysinfo -e` return "unknown" in this case?

mshych:
> In our case, `machine-build.conf` specifies `onie_skip_ethmgmt_macs=no`. Because of this, `onie-sysinfo -e` calls `get_ethaddr_arch` which calls `onie-syseeprom -g 0x24` which returns an error as the value is missing from the syseeprom. This error is then passed to `mac_add` in [networking.sh line 207](https://github.com/Benjamin-Davies/onie/blob/6533201cd74b5b2e318783597f59c5c891b37695/rootconf/default/etc/init.d/networking.sh#L207), which fails with the arithmetic syntax error.
> 
> The point of this patch is to handle this error case. Even if `onie-sysinfo -e` did return "unknown", networking.sh line 207 would still fail as "unknown" is not a valid mac address.
> 
> Would you rather `onie-sysinfo -e` return "unknown" in this case?

If this is a common issue affecting all systems of the same machine type, please resolve it by setting onie_skip_ethmgmt_macs=no for that machine.

If the issue occurs only on specific systems (for example, due to incorrect production or customization), please fix it by setting the BASE_MAC using:
onie-syseeprom -s 0x24=...

PR #1114, opened 2026-06-01
ci: add GitHub Action to build and boot-test the kvm_x86_64 ONIE image

bhouse-nexthop:
## Summary

Adds a GitHub Actions workflow that **builds and boot-tests** the generic
amd64 ONIE target (`kvm_x86_64`, intended to run as a VM under KVM/QEMU) on
every push and pull request. The intent is to make reviewing other PRs easier:
a green check confirms the change still produces an ONIE image that **actually
boots**, so "it builds and boots" becomes part of acceptance testing rather
than something a reviewer verifies by hand.

## `build` job

Containerized build following the upstream-recommended flow (`contrib/build-env`,
`machine/kvm_x86_64/INSTALL`):
- `.github/onie-build/Dockerfile` provides a Debian 11 build environment
  (required because `build-config` still pulls in python2-era tooling), run as a
  `build` user matching the host UID/GID.
- Builds with secure boot enabled (kvm_x86_64's default): `signing-keys-generate`
  → `shim-self-sign` → `make MACHINE=kvm_x86_64 all recovery-iso`.
- The crosstool-NG cross toolchain (the ~25-min long pole) is built as a discrete
  `xtools` target and cached, keyed on every toolchain input. On a cache hit
  `make xtools` is a no-op.
- Uploads the recovery ISO as a workflow artifact for the boot-test job.

## `boot-test` job (`needs: build`)

Downloads the recovery ISO and boots it **headlessly under QEMU + OVMF (UEFI)**,
capturing the serial console and asserting ONIE gets **past GRUB into the OS** —
the GRUB menu, the ONIE banner, and the rescue console all appear, with **no
kernel panic**.
- ONIE routes GRUB, the kernel, and userspace all to `ttyS0`, so a serial
  capture is sufficient — no screen-scraping/OCR.
- GitHub's Linux runners expose `/dev/kvm` (after a udev rule grants access), so
  QEMU uses KVM acceleration, falling back to TCG otherwise.
- Secure-Boot-relaxed: it validates *does it boot*; enforcing the SB signature
  chain is left to a later change.
- Harness: `emulation/ci-boot-test.sh`.

## Notes
- apt `Acquire::Retries` so transient Debian mirror resets don't fail the build.
- Concurrency group (workflow + ref, cancel-in-progress) so a newer push cancels
  the prior run.

### DIFF
--- a/.github/onie-build/Dockerfile
+++ b/.github/onie-build/Dockerfile
@@ -1,0 +1,103 @@
+# ONIE build environment for CI.
+#
+# Debian 11 is required: onie/build-config/Makefile still pulls in
+# python2-era tooling (python-all-dev) that is unavailable on newer
+# Debian releases.
+#
+# This image builds the generic kvm_x86_64 target with secure boot
+# enabled.  It includes the secure-boot key-generation and EFI signing
+# tooling (gnupg2, efitools, sbsigntool), the NSS command-line tools the
+# shim signing step shells out to (libnss3-tools: pk12util/certutil), and
+# the build dependencies for the pesign that ONIE compiles from source
+# (NSS/NSPR dev packages).
+FROM debian:11
+WORKDIR /onie
+
+# Tolerate transient Debian mirror hiccups (connection resets mid-fetch).
+RUN echo 'Acquire::Retries "5";' > /etc/apt/apt.conf.d/80-retries
+
+# Build dependencies for crosstool-NG and the ONIE image.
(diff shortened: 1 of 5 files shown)


bhouse-nexthop:
Closing: this change has been incorporated into the new `onie-modernization-2026` branch by the maintainers (applied as commit `1a24708f`). No further action needed here — tracking the modernization work on that branch going forward.

PR #1124, opened 2026-06-15
Accton AS9817_32O/AS9817_32D: add to support ONIE

laura-sychang:
Accton AS9817_32O/AS9817_32D: add to support ONIE


mshych:
Pushed to the master. Thank you.

PR #1116, opened 2026-06-11
python2-to-3: port python to python3; remove unused gitdm git-stats tooling

bhouse-nexthop:
## Summary

Two related cleanups that make the tree Python-3-only:

1. **Port all retained Python from Python 2 to Python 3** — the build-path `mk-part-table.py`, the `$(PYTHON)` plumbing, the `contrib/onie-server` / `contrib/oce` tooling, and the per-machine test/DUT scripts.
2. **Remove the unused gitdm-based git-stats tooling** (`contrib/git-stats/`) instead of porting it (see rationale below).

## Commits in this PR
- **python2-to-3: port all python from python2 to python3**
- **contrib: remove unused gitdm-based git-stats tooling**

## Why remove gitdm rather than port it

`contrib/git-stats/` vendored Jonathan Corbet's LWN "git data miner" (gitdm) to generate ONIE release contribution statistics. The evidence says retire it:

- **Dead weight:** imported once in 2018 (`052c3f91`) and untouched for ~8 years — the only later change would have been the mechanical py2→py3 port in this series.
- **Unused / unreferenced:** nothing in the tree references it — no build, no CI, no docs; it is a self-contained island run by hand, if ever. No generated reports were ever committed.
- **No upstream to track:** canonical gitdm (`git.lwn.net`) is *itself still Python 2* and effectively unmaintained, so there is no newer version to sync — only third-party forks to adopt. Porting it would mean carrying ~10 files of someone else's stalled tool.
- **Nothing lost:** the affiliation maps remain in git history, and equivalent statistics are trivially regenerable (upstream gitdm, or `git shortlog -sne`).

## Stacked series

Cross-repo stacked PR: prerequisite work is squashed into the baseline commit `stacked PR #1115` — **not for review here**; review only the commits above.

Depends on (merge first, in order):
1. #1114 — ci: add GitHub Action
2. #1115 — build: generate per-image SBOMs

PR #1125, opened 2026-07-09
python2-to-3: port python to python3; auto-install a python3 gitdm for git-stats

bhouse-nexthop:
Rebased onto `onie-modernization-2026` (replaces #1116, which targeted `master` and stacked on the SBOM work; the SBOM content is **not** included here).

### 1. Port all python from python2 to python3
`python2-to-3: port all python from python2 to python3` — ports the remaining python2 code (build-config Makefile/scripts, `contrib/oce`, `contrib/onie-server`, machine DUT test helpers) to python3. `$(PYTHON)` now defaults to `python3`. All ported files compile under python3.

### 2. git-stats: drop the vendored python2 gitdm, auto-install a python3 gitdm
The `contrib/git-stats` tooling vendored a full copy of gitdm (from `git.lwn.net`), which is python2 and effectively unmaintained. Per maintainer request, rather than port/maintain a copy in-tree:

- **Removed** the vendored `contrib/git-stats/gitdm/` engine.
- **Kept** the ONIE-specific `onie-git-stats` wrapper, the `gitdm-config/` maps, and the README.
- `onie-git-stats` now **auto-installs a python3 gitdm on demand**: on first run it clones a pinned python3 port ([OSSystems/gitdm](https://github.com/OSSystems/gitdm), a straight python3 port of Corbet's gitdm) into a git-ignored cache (`contrib/git-stats/.gitdm`) and reuses it thereafter. So `onie-git-stats` runs as-is with only `git` + `python3` present.
- Source/revision are **pinned and overridable** via `GITDM_REPO` / `GITDM_REF`.
- A user who already has a python3 gitdm can set **`GITDM`** to point at it; the wrapper then uses it directly and **skips the clone** entirely.
- gitdm is now invoked with `python3` (was `python`), and the pinned upstream's harmless `SyntaxWarning`s are silenced.
- `README.md` documents the auto-install, the pinning/override knobs, and the pre-existing-gitdm path.

## Testing
- `onie-git-stats <range> <prefix>` exercised end-to-end: fresh auto-clone → python3 gitdm → valid `results.txt`/`.html`.
- Verified the `GITDM` override skips the clone.
- `python3 -m py_compile` passes on all ported `.py` files.


bhouse-nexthop:
Added a small, path-scoped CI check (`ci: validate contrib/git-stats gitdm tooling`) so this doesn't rely on manual testing: it runs `onie-git-stats` end-to-end and asserts a report is actually generated, covering **both** the auto-install path (clones the pinned python3 gitdm) and the `GITDM` override path (uses a pre-existing gitdm, skips the clone). It only triggers when `contrib/git-stats/**` (or the workflow itself) changes.

mshych:
Thank you, pushed to onie-modernization-2026 branch.

PR #1127, opened 2026-07-09
packages: modernize packages that build on the existing toolchain

bhouse-nexthop:
Targets `onie-modernization-2026`. **Independent** — not stacked on any other modernization PR. It applies cleanly on `onie-modernization-2026` and every package builds on the **existing** cross-toolchain (no dependency on the toolchain milestone, the python3 port, SBOM, or OCE).

Each package is now its **own commit** (one bump per commit, bisectable and reviewable in isolation).

## New dependency libraries
Added to unblock consumers:

| Package | Version | For |
| --- | --- | --- |
| libmnl | 1.0.5 (new) | ethtool's netlink interface |
| libaio | 0.3.113 (new) | lvm2 ≥ 2.03's bcache I/O engine |

## Package upgrades

| Package | From | To | Notes |
| --- | --- | --- | --- |
| ethtool | 3.14 | 6.15 | now uses libmnl for netlink |
| zlib | 1.2.11 | 1.3.2 | source → madler GitHub release (old SF mirror 404s) |
| lzo | 2.09 | 2.10 | |
| libpcap | 1.10.4 | 1.10.6 | |
| tcpdump | 4.99.4 | 4.99.6 | needs libpcap ≥ 1.10.5 |
| dropbear | 2020.81 | 2026.91 | |
| keyutils | 1.5.10 | 1.6.3 | source → kernel.org snapshot (.tar.bz2 → .tar.gz) |
| dmidecode | 3.1 | 3.7 | cross-compile patch refreshed |
| acpica-tools | 20150410 | 20260408 | source → GitHub release assets (acpica.org retired) |
| ipmitool | 1.8.18 | 1.8.19 | GitHub source archive (./bootstrap) |
| kexec-tools | 2.0.22 | 2.0.32 | drop upstreamed aarch64 / dt-sorting patches |
| btrfs-progs | v4.9.1 | v7.0 | |
| pciutils | 3.2.1 | 3.15.0 | ONIE Makefile patch refreshed |
| gptfdisk | 0.8.8 | 1.0.10 | drop obsolete add-onie-partition-types patch |
| popt | 1.16 | 1.19 | drop obsolete aarch64 / uClibc-glob patches |
| openssl | 1.1.1g | 3.5.7 (LTS) | drops legacy 1.1.x support |
| mtd-utils | 1.5.2 | 2.3.1 | |
| i2c-tools | 3.1.1 | 4.4 | |
| dosfstools | 3.0.26 | 4.2 | |
| busybox | 1.25.1 | 1.38.0 | see DHCP VIVSO note below |
| lvm2 | 2.02.105 | 2.03.41 | requires libaio |

All bumps target current upstream releases from canonical sources (stable hosts preferred over flaky mirrors); patch series were refreshed and dead/flaky source URLs fixed as part of each bump.

## busybox 1.38 + DHCP VIVSO
busybox 1.38 dropped the `vivso` (opt-125) and `tftpsiaddr` (opt-150) names from udhcpc's stock table. The busybox commit restores them (extended `patches/busybox/dhcp-additional-options.patch`) and makes ONIE's discovery decode DHCP options type-aware in `rootconf/default/lib/onie/udhcp4_sd` / `udhcp4_net`, so VIVSO-based OCP discovery keeps working. (The OCE compliance PR is the gate that exercises this end-to-end.)

## Testing
Validated by full local cross-builds producing a green recovery ISO, plus CI on the prior stack. Commits are individually bisectable (each builds in order).


build-config/conf/busybox.config
```diff
@@ -551,45 +603,27 @@ CONFIG_DEFAULT_DEPMOD_FILE="modules.dep"
 #
 # Linux System Utilities
 #
+CONFIG_ACPID=y
+CONFIG_FEATURE_ACPID_COMPAT=y
 # CONFIG_BLKDISCARD is not set
+CONFIG_BLKID=y
+# CONFIG_FEATURE_BLKID_TYPE is not set
```
mshych: Change to y

build-config/conf/busybox.config
```diff
@@ -721,65 +782,99 @@ CONFIG_FEATURE_BEEP_LENGTH_MS=30
...
 # CONFIG_SYS_EEPROM_DEVICE_MTD is not set
 # CONFIG_SYS_EEPROM_DEVICE_DISK is not set
 # CONFIG_SYS_EEPROM_SYSFS_FILE is not set
 CONFIG_SYS_EEPROM_I2C_DEVICE=""
 CONFIG_SYS_EEPROM_I2C_ADDR=
-CONFIG_SYS_EEPROM_I2C_MEM_ADDR_BITS=8
-CONFIG_SYS_EEPROM_I2C_WRITE_1B_DELAY=10
-CONFIG_SYS_EEPROM_I2C_WRITE_2B_DELAY=1000
-CONFIG_SYS_EEPROM_I2C_WRITE_3B_DELAY=10
+CONFIG_SYS_EEPROM_I2C_MEM_ADDR_BITS=0
```
mshych: Change this and the following 3 configs to the same default values

patches/busybox/dhcp-additional-options.patch
mshych, line 1: DHCP option 7 (log server) is used in multiple machine forks.
Add a patch just for this option.

patches/busybox/dhcp-logsrv-option.patch
dhcp-logsrv-option patch

SPDX-License-Identifier:     GPL-2.0

Re-add DHCP option 7 (log server) to udhcpc's built-in option table.

ONIE requests logsrv by name in config_ethmgmt_dhcp4() and consumes
$logsrv in the udhcp4_net action script, including in the per-machine
copies of both, so the name has to exist in the table rather than be
decoded by a single caller.

diff --git a/networking/udhcp/common.c b/networking/udhcp/common.c
index ad580f3..1cc6da6 100644
--- a/networking/udhcp/common.c
+++ b/networking/udhcp/common.c
@@ -27,7 +27,7 @@ const struct dhcp_optflag dhcp_optflags[] ALIGN2 = {
 //     { OPTION_IP | OPTION_LIST                 , 0x04 }, /* DHCP_TIME_SERVER   */
 //     { OPTION_IP | OPTION_LIST                 , 0x05 }, /* DHCP_NAME_SERVER   */
        { OPTION_IP | OPTION_LIST     | OPTION_REQ, 0x06 }, /* DHCP_DNS_SERVER    */
-//     { OPTION_IP | OPTION_LIST                 , 0x07 }, /* DHCP_LOG_SERVER    */
+       { OPTION_IP | OPTION_LIST                 , 0x07 }, /* DHCP_LOG_SERVER    */
 //     { OPTION_IP | OPTION_LIST                 , 0x08 }, /* DHCP_COOKIE_SERVER */
        { OPTION_IP | OPTION_LIST                 , 0x09 }, /* DHCP_LPR_SERVER    */
        { OPTION_STRING_HOST          | OPTION_REQ, 0x0c }, /* DHCP_HOST_NAME     */
@@ -101,7 +101,7 @@ const char dhcp_option_strings[] ALIGN1 =
 //     "timesrv" "\0"          /* DHCP_TIME_SERVER     */
 //     "namesrv" "\0"          /* DHCP_NAME_SERVER     */
        "dns" "\0"              /* DHCP_DNS_SERVER      */
-//     "logsrv" "\0"           /* DHCP_LOG_SERVER      */
+       "logsrv" "\0"           /* DHCP_LOG_SERVER      */
 //     "cookiesrv" "\0"        /* DHCP_COOKIE_SERVER   */
        "lprsrv" "\0"           /* DHCP_LPR_SERVER      */
        "hostname" "\0"         /* DHCP_HOST_NAME       */

And add it to patches/busybox/series

rootconf/default/lib/onie/udhcp4_sd
```diff
@@ -15,10 +15,58 @@ PROG="DHCP"
...
+        # busybox's udhcpc exports a RECEIVED option as a friendly typed
+        # "<name>=<value>" env var only if that option is in its built-in table;
+        # options it doesn't know are exported generically as "opt<code>=<hex>"
+        # -- RAW HEX, untyped.  busybox trims rarely-used options from that table
+        # for size (it dropped vivso/125 after ~1.25 and never carried
+        # tftpsiaddr/150), so several options ONIE's discovery consumes by name
+        # arrive as opt<code> in hex.  Rather than carry a busybox source patch
+        # to re-add them (which upstream is unlikely to take back), decode those
+        # here -- the one place that turns the udhcpc env into onie_disco_* --
+        # to the typed value and name ONIE expects:
```
mshych: +        #    43 -> vendoropts  (VENDOR_OPTS, binary -- left as hex)

rootconf/default/lib/onie/udhcp4_sd
```diff
@@ -15,10 +15,58 @@ PROG="DHCP"
...
+                local d=$((0x${h%${h#??}})); h=${h#??}
+                out="$out $a.$b.$c.$d"
+            done
+            printf '%s' "${out# }"
+        }
+        [ -n "$opt114" ] && export url="$(hex2str "$opt114")"
+        [ -n "$opt72"  ] && export wwwsrv="$(hex2ip "$opt72")"
+        [ -n "$opt150" ] && export tftpsiaddr="$(hex2ip "$opt150")"
+        [ -n "$opt125" ] && export vivso="$opt125"
+
```
mshych, line 66: +        [ -n "$opt43"  ] && export vendoropts="$opt43"

rootconf/default/lib/onie/udhcp4_sd
mshych, line 1: Option 43 is still in rootconf/default/bin/discover
and can be used in NOS installer

build-config/conf/busybox.config
bhouse-nexthop: Done, `CONFIG_FEATURE_BLKID_TYPE=y`.

For the record, since it affects how you may want to read it: this one was not a regression from the bump. The pre-bump config also had `# CONFIG_FEATURE_BLKID_TYPE is not set` — it has been off since the initial commit on both `master` and `onie-modernization-2026`, and it only shows up as an added line here because the file was wholesale-regenerated and the surrounding lines moved.

I applied it anyway because I think your reasoning holds independently: busybox declares it `default y` (`util-linux/blkid.c:18`), so ONIE has been overriding the upstream default off, and that matches the standard you applied in the next comment.

Checked it is safe for existing callers: `TYPE="..."` is printed after `LABEL` and `UUID` (`util-linux/volume_id/get_devname.c:244`), so the UUID extraction in `onie-blkdev-common`, the `s/:.*$//` device-list parsing in `discover`, and the `ONIE-BOOT|-DIAG` greps are all unaffected.

build-config/conf/busybox.config
bhouse-nexthop: Done — restored to 8 / 10 / 1000 / 10.

You were right, and it is worse than a wrong default: **it breaks the build**. Those four are `range 8 16` and `range 10 1000000`, so `0` is outside every one of them. kconfig discards an out-of-range value and treats the symbol as new, which makes `silentoldconfig` prompt — and a prompt with redirected console input aborts:

```
delay time for i2c writing 1 byte (SYS_EEPROM_I2C_WRITE_1B_DELAY) [10] (NEW) aborted!
Console input/output is redirected. Run 'make oldconfig' to update configuration.
```

Reproduced against real busybox 1.38.0 with this branch's patch series and a real machine config (`accton_833435_b21`, which leaves the 1B/2B delays to the base config):

- as submitted → `silentoldconfig` rc=2, no `include/autoconf.h`, build fails
- with these four restored → rc=0, and the machine's own overrides win (16 address bits, 4000 for 3B) with the base defaults filling the rest

**167 of the 247** machines that set `CONFIG_SYS_EEPROM_DEVICE_I2C=y` leave at least one of the four to the base config, so all of them would have failed to build. CI missed it because `kvm_x86_64` uses `SYS_EEPROM_DEVICE_DISK`, which leaves the i2c symbols invisible.

I then scanned the whole regenerated config for the same class of bug — any int/hex value outside its declared Kconfig range — and found two more, so I fixed those here too:

- `CONFIG_SYS_EEPROM_MAX_SIZE=0` (`range 256 65536`, default 2048). Note this is out of range in the **pre-bump** config as well, so it is not from this PR; it is inert only because all 283 `SYS_EEPROM=y` machines set it explicitly. Happy to split it out if you would rather not carry an unrelated fix here.
- `CONFIG_FEATURE_HTTPD_PORT_DEFAULT=0` (`range 1 65535`, default 80). New symbol in 1.38, inert today because no machine enables `HTTPD`.

Verified afterwards by running `silentoldconfig` for **all 289** machine busybox configs in the tree against the fixed base config: 0 aborts.

patches/busybox/dhcp-additional-options.patch
bhouse-nexthop: Done — added `patches/busybox/dhcp-logsrv-option.patch` and listed it in `patches/busybox/series`.

I regenerated the diff from a pristine 1.38.0 tree rather than pasting yours, purely so I could be sure it applies in series order; it came out with the same blob hashes you had (`ad580f3..1cc6da6`), i.e. byte-identical to your version. Verified all 9 patches now apply cleanly in order, and `logsrv` is back in both `dhcp_optflags[]` and `dhcp_option_strings[]`.

You were right that this cannot be handled in one caller, and the consequence is more severe than losing the log server. `-O <name>` with an unknown name is fatal — `bb_error_msg_and_die("unknown option '%s', known options: %s")` at `networking/udhcp/common.c:229`. So without the table entry:

- **19** machines carry their own `rootconf/sysroot-etc/init.d/networking.sh` that still requests `-O logsrv` by name (18 ufispace, 1 stordis) → udhcpc dies → **DHCP fails entirely** on those platforms, not just syslog.
- **29 of 29** machine copies of `udhcp4_net` reference `$logsrv` without the `opt7` aliasing I had added to the default copy → silent loss of the DHCP-supplied log server.

Since the option is back in the table, the script-side workarounds are unnecessary, so I reverted them: `networking.sh` goes back to requesting `logsrv` by name and `udhcp4_net` drops the hex decode. Both files are now byte-identical to the pre-bump tree, so this PR no longer touches either one — which also means the machine copies stay consistent with the default ones.

rootconf/default/lib/onie/udhcp4_sd
bhouse-nexthop: Added, along with the rest of the option 43 handling — see the reply on the `[ -n "$opt43" ]` comment below for the details.

rootconf/default/lib/onie/udhcp4_sd
bhouse-nexthop: Added, `[ -n "$opt43"  ] && export vendoropts="$opt43"`, with the matching `43 -> vendoropts` line in the comment block above.

Confirmed the regression is real: `sd_dhcp4()` in `rootconf/default/bin/discover:154` requests `-O 7 43 54 66 67 72 114 125` and this PR does not change that line. Before the bump, the dropped `dhcp-additional-options.patch` put option 43 in udhcpc's table, so it arrived as `vendoropts=<hex>` and `printenv` carried it into `onie_disco_vendoropts`. After the bump it arrives as `$opt43`, and the new `sed -e '/^opt[0-9]/d'` I added to drop the raw `opt<code>` vars was throwing it away — so ONIE was requesting the option and then silently discarding the answer.

While checking this I confirmed the other six options `sd_dhcp4()` requests are all covered: 54/66/67 are `serverid`/`tftp`/`bootfile` in busybox's stock table, 72/114/125 are decoded here, and 7 is back in the table via the new patch.

rootconf/default/lib/onie/udhcp4_sd
bhouse-nexthop: Confirmed, and fixed — thanks, this is exactly where the breakage was.

`rootconf/default/bin/discover:154` requests option 43 and is unchanged by this PR, and three pegatron machines carry their own copy of `discover` with the same request list. `import_parms()` in `lib/onie/functions` exports every `onie_disco_*` pair into the environment, so `onie_disco_vendoropts` is visible to the NOS installer even though ONIE itself only decodes `onie_disco_vivso` (option 125) internally. Dropping option 43 therefore broke a real interface with no in-tree consumer to fail loudly and show it.

mshych:
Busybox additional notes:
1. busybox dropped timeout -t in 1.30 (coreutils syntax now), so 38 call sites in ufispace machine trees stop working — link_phy.sh PHY init on 6 platforms, and the NTP block in 29 copies of udhcp4_net (live, since those machines set CONFIG_NTPD=y). Same bucket: 30 ufispace busybox configs still set CONFIG_FEATURE_WGET_SSL_HELPER=y, a symbol gone since 1.26, which apply-config-patch now silently drops. 
As we discussed before: These are vendor-specific platform changes, and the question is what is preferable: pushing untested changes that may or may not work on a vendor platform, or letting vendors validate and opt into the new compiler/toolchain themselves. A vendor that wants to adopt the new toolchain should verify that its platform builds and functions correctly and then update its machine configuration as needed. Until that validation is done, we should not switch vendor-specific platforms to an unverified configuration. 
Please just flag it in the commit message / release notes so the affected vendors know what to rework.

2. oldconfig picked up four new upstream defaults not mentioned in the commit message: CONFIG_TLS, CONFIG_SSL_CLIENT, CONFIG_SSL_SERVER, CONFIG_FEATURE_WGET_HTTPS. Pre-bump ONIE had no TLS, so https:// installer URLs failed; now exec_installer will fetch them via code that prints note: TLS certificate validation not implemented. FEATURE_WGET_OPENSSL isn't a substitute (we ship no CA store, so -verify_return_error fails everything). Please set all four to "not set" and keep this PR behaviour-neutral; verified HTTPS deserves its own PR.

3. The config fix isn't stable: feeding the fixed busybox.config back through make oldconfig re-zeroes the four SYS_EEPROM_I2C_* values, SYS_EEPROM_MAX_SIZE and FEATURE_HTTPD_PORT_DEFAULT, because those symbols are invisible until a machine override makes them visible. So the next regeneration reintroduces the exact bug, and CI won't catch it (kvm_x86_64 uses SYS_EEPROM_DEVICE_DISK). Worth adding the sweep you ran by hand as a make target plus a CI step — all 285 machines take ~10s once busybox is unpacked.

mshych:
i2c-tools bump breaks the build for every machine that uses i2c-tools for syseeprom, and it's invisible to CI. Those are exactly the machines you're asking about: build-config/Makefile only pulls in make/i2ctools.make when I2CTOOLS_SYSEEPROM = yes, which is 6 machines, and none of them override busybox.config — the base config has # CONFIG_SYS_EEPROM is not set, so they have no busybox syseeprom applet and depend entirely on the i2c-tools-built /usr/bin/onie-syseeprom. If i2c-tools doesn't build, those machines have no way to read the ONIE TLV EEPROM at all.
E.g. alphanetworks and celestica

mshych:
OpenSSL bump.
OpenSSL 3.x installs ENGINESDIR=$(libdir)/engines-{SHLIB_VERSION}, i.e. /usr/lib/engines-3. OPENSSL_LIBS still says plain engines:


openssl.make
Ln 45–48
OPENSSL_LIBS	= \
	engines \
	libcrypto.so libcrypto.so.3 \
	libssl.so libssl.so.3
Both the chmod u+w -R loop in openssl-build and the cp -av loop in openssl-install therefore fail on that entry. They don't fail the build, because engines is the first item and a shell for loop exits with the status of its last iteration — I confirmed this:

chmod: cannot access '/tmp/enginetest/usr/lib/engines': No such file or directory
loop exit status = 0
So the effect is two error messages scrolling past and engines-3/ silently never reaching $(SYSROOTDIR). This is newly reachable in this PR: the base branch defaulted to OPENSSL_VERSION ?= 1.1.1g, whose OPENSSL_LIBS correctly said engines-1.1; the engines spelling came from the never-default 3.4.0 branch. install_sw does depend on install_engines and install_modules, so engines-3/ and ossl-modules/ both really exist in DEV_SYSROOT and are just not copied. The new comment # ONIE standardizes on OpenSSL 3.x (SONAME major 3, "engines" dir) is also wrong on that last point.

mshych:
acpica-tools bump.

acpinames was deleted upstream, and asking for it recurses forever
changes.txt in 20260408 says "Removed support for the acpinames utility", generate/unix/acpinames/ is gone, and upstream PROGS is now acpibin acpidump acpiexamples acpiexec acpihelp acpisrc acpixtract iasl. ONIE still asks for it:


acpica-tools.make
Ln 30–30
ACPIBINS = acpibin acpidump acpiexec acpihelp acpinames acpisrc acpixtract
The upstream build rule is @cd $(BUILD_DIRECTORY_PATH)/$@; mkdir -p obj; $(MAKE) || exit "$$?" — separated by ;, not &&. So the failed cd leaves make in the top directory, where it re-invokes itself with the same PROGS, forever. Reproduced:

/bin/sh: 1: cd: can't cd to generate/unix/acpinames
make[1]: Entering directory '/tmp/dl/x/acpica-unix-20260408'
... (still climbing at make[15] when I killed it at 25s)
Severity is limited because ACPI_ENABLE is never set to yes anywhere in the tree — build-config/Makefile:432 is its only consumer — so no machine builds acpica today. That's exactly why local builds and CI missed it, but it means the acpica bump is unbuildable as written for anyone who opts in.

mshych:
Other minor items:
1. The coverage gap.
Three of the bumps are not compiled by any default build, which is worth flagging explicitly since it's the same blind spot that hid the acpica bug:

ipmitool 1.8.19 — IPMITOOL_ENABLE ?= no in build-config/arch/x86_64.make:159, so the x86_64 default and kvm_x86_64 skip it. But 106 machines under machine/ set IPMITOOL_ENABLE = yes. This is the bump with the most new machinery in it (GitHub tag archive instead of a dist tarball, a directory rename in the source step, and ./bootstrap/autoreconf at configure time), and it reaches 106 platforms without any build in the repo exercising it. I'd ask him to confirm he built this one specifically.
tcpdump 4.99.6 and libpcap 1.10.6 — TCPDUMP_ENABLE is never assigned a default anywhere, and zero machines set it to yes. libpcap.make is only ever included from tcpdump.make:12, so both are effectively dead code today. Low risk, but it means neither bump has been compiled.
For contrast, the ones a kvm_x86_64 build does cover are mtd-utils, gptfdisk, btrfs-progs, lvm2, ethtool + libmnl, dosfstools, kexec-tools, dmidecode, keyutils, dropbear, zlib, lzo, popt, openssl, mokutil and libaio — so the bulk of the PR is genuinely exercised. Related: only the DCO check runs on this PR, so the "CI on the prior stack" claim in the description isn't visible from here.

2. Small things to clean up.
mtdutils-configure-help looks like debugging scaffolding that got committed — it's a real target wired into PHONY that just runs configure --help. Probably meant to be local.

libaio is the only package in the PR that receives neither ONIE_CFLAGS nor ONIE_LDFLAGS, so it builds with upstream's -g -fomit-frame-pointer -O2 and, more notably, without --sysroot=$(DEV_SYSROOT). It only needs libc so it should link against the toolchain's own sysroot, and -g gets stripped later, but it's an odd exception to the convention every other fragment follows. (libmnl passes CFLAGS but not LDFLAGS; that one is genuinely harmless, since automake puts CFLAGS on the link line and --sysroot rides along.)

The stale .sha1 cleanup is inconsistent: the PR deletes six (dosfstools 3.0.26, i2c-tools 3.1.1, mtd-utils e4c8885, openssl 1.0.2l/1.1.1g/3.4.0, pciutils 3.2.1) and leaves fifteen behind (acpica 20150410, btrfs-progs v4.9.1, dmidecode 3.1, dropbear 2020.81, ethtool 3.14, gptfdisk 0.8.8, ipmitool 1.8.18, kexec-tools 2.0.22, keyutils 1.5.10, libpcap 1.10.4, lzo 2.09, popt 1.16, tcpdump 4.99.4, zlib 1.2.11, busybox 1.25.1). Since the repo already keeps unused ones like lvm2-2_02_155.tar.xz.sha1, the deletions are arguably the deviation — either way it should be one rule.

3. The new packages will be uploaded to ONIE mirror.
Two of the new sources are generated on demand, not uploaded artifacts. keyutils now comes from a git.kernel.org cgit /snapshot/ URL and ipmitool from https://github.com/ipmitool/ipmitool/archive/refs/tags. Neither host guarantees the bytes stay identical across cgit/git version changes (GitHub has broken auto-generated archive checksums before), so the pinned sha1s in upstream/ can start failing without anything in the PR changing. Both verified correct today; no action needed on your side — I'll mirror the tarballs so $(ONIE_MIRROR) is a real fallback.

PR #1134, opened 2026-07-16
Accton AS1817_64O: add to support ONIE

laura-sychang:


mshych:
Please add a new platform to build-config/scripts/onie-build-targets.json

PR #1137, opened 2026-08-18
ci: harden the kvm_x86_64 build/boot workflow and boot harness

bhouse-nexthop:
Follow-up fixes to the CI added in `1a24708f`, all found in review by @mshych on #1126. They are here rather than in that PR because they fix already-merged code, which is how he asked for them to be split.

Every item was re-verified against the tree before being applied; the notes are his numbering.

**Workflow**

1. **Cross-toolchain cache key omitted inputs that change the toolchain.** `xtools.make:25` derives `XTOOLS_VERSION = $(ONIE_ARCH)-g$(GCC_VERSION)-lnx$(LINUX_RELEASE)-...`, but `LINUX_RELEASE` is defined in `build-config/make/kernel-download.make:22` and `ONIE_ARCH` in `machine/kvm_x86_64/machine.make:10` — neither was hashed. A kernel bump therefore keeps *hitting* the key while the `build/x-tools/<XTOOLS_VERSION>` directory name changes underneath it: the restored toolchain is the wrong one, `make xtools` rebuilds from scratch, and the save step (gated on a cache miss) skips it, so every later run pays the same rebuild. Silent and self-perpetuating. `GCC_VERSION`/`XTOOLS_LIBC` default in `crosstool-ng.make`, which is already hashed, so those two were the only gaps. Also adds his guard: a `::warning` when a cache hit still had to build a toolchain.

   This is the one worth landing before the kernel PR (#1130) does exactly this bump.

2. **`push:` and `pull_request:` both unfiltered** ran the pipeline twice for every push to a branch with an open PR; the concurrency group can't collapse them because the refs differ. Fixed, but **not** with the branch filter that was suggested — see the note at the bottom. A small `gate` job now skips the push run when an open PR in the same repository already covers the commit.

3. **`--privileged` dropped** from both `docker run` invocations. The Dockerfile ends with `USER build`, so the build is unprivileged and can't use anything `--privileged` grants. Confirmed the build needs no loop mounts — it goes through `fakeroot` and `mtools`, and `onie-mk-iso.sh` hardcodes `/sbin/mkdosfs`.

4. **`permissions: contents: read`** at workflow level plus `timeout-minutes` on each job (180/45/60).

**Dockerfile**

5. **`groupadd -g $GID` failed when the host GID already exists.** Confirmed in `debian:11`: `groupadd -g 30 build` → `rc=4, GID '30' already exists`. The low range is full of system groups (30 dip, 50 staff, 100 users), which is common with central/NFS accounts and the norm on macOS (GID 20 = dialout there). CI never sees it because the runner's GID is free. Reuses the group when the GID is taken, and chowns by numeric id since `build` then isn't a group name. Test-built at GID 30 and GID 1001 — both work now.

10. **The `/sbin:/usr/sbin` PATH line never took effect.** It was appended to `~/.bashrc`, and Debian's default `.bashrc` returns early for non-interactive shells, so `bash -lc` never reaches it. As he noted, `ENV` alone isn't enough either: `/etc/profile` lines 5–9 *replace* PATH for non-root users with one that omits both sbin dirs. Verified in the image — `bash -lc` as non-root gets `/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games` and `blkid` is off PATH. Fixed with `ENV` plus an `/etc/profile.d` snippet (profile.d is sourced at line 27, after the reset). Verified both login and non-login shells now resolve `/sbin/blkid`.

**emulation/ci-boot-test.sh**

9. **The "ONIE userspace" milestone was matched by GRUB output.** `build-config/recovery/grub-iso.cfg:32` echoes `"ONIE: Rescue Mode ..."` and `:24` echoes `"Version   : ..."`, both before the kernel loads — so the check duplicated "GRUB reached". Now asserts `Info: BIOS mode:`, printed unconditionally from `rootconf/grub-arch/sysroot-lib-onie/init-arch:130`.

6. **secureboot mode never checked that Secure Boot was enforced.** Now asserts the guest's own `Info: Secure Boot: Active.` (`init-arch:121`, read from the SecureBoot EFI variable via efivar), and asserts its absence on the relaxed run.

7. **A failed `virt-fw-vars` was silently ignored** — piped through `| sed | grep ... || true`, which dropped its status twice over. Enrolment moved into a helper that fails the run.

8. **A QEMU that never started could be scored as a successful SB rejection.** QEMU's stderr is kept and an immediate exit is now a hard error. Verified with a stub QEMU that exits 1: the harness prints the stderr and exits 2, where before it reported `FAIL no serial output captured` / `PASS ONIE did not boot without db`. Also captures the OVMF debug console when `isa-debugcon` is present — belt-and-braces, since as he corrected himself the release OVMF does put its `BdsDxe: ... Access Denied` line on ttyS0.

11. **One vCPU when `/dev/kvm` is unusable**, so the documented `accel=kvm:tcg` fallback works for local runs. CI has `/dev/kvm` and still gets two.

Smoke-tested locally against real OVMF: logs are written as expected, the firmware console is empty on Ubuntu's release OVMF (as he predicted), and the milestone assertions behave correctly on a non-ONIE ISO.

---

### Note on the duplicate-run fix

@mshych suggested restricting `push:` to the long-lived branches, which is the usual answer. I took a different route because the filter has a blind spot that bites this exact patch series.

A fork's pull requests are opened against the *upstream* repository, so they raise no `pull_request` event in the fork; the fork's only trigger is `push`. Filtering `push` by branch name therefore leaves every topic branch in a fork with **no CI at all**, and `workflow_dispatch` cannot fill the gap unless the workflow also exists on the fork's default branch (it does not — this workflow lives on `onie-modernization-2026`, and forks default to `master`).

That is not hypothetical: with the branch filter applied, pushing this very branch to my fork produced zero runs, which is how I noticed.

The `gate` job targets the actual condition instead — "an open PR here already covers this commit" — so it drops the duplicate upstream while a fork, where the query finds nothing, keeps full push CI. Verified on the fork: `No open pull request in bhouse-nexthop/onie for bhouse.ci-review-fixes; running.` Only `build` depends on the gate; the other jobs chain from it and skip automatically.

If you would rather have the simple branch filter, say so and I will switch it — the tradeoff is that fork-based development loses pre-submit CI.

### DIFF
--- a/.github/onie-build/Dockerfile
+++ b/.github/onie-build/Dockerfile
@@ -87,14 +87,35 @@ RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
 #       (https://github.com/moby/moby/issues/5419#issuecomment-41478290)
 ARG UID=1000
 ARG GID=1000
-RUN groupadd -g $GID build && \
-    useradd -l -m -u $UID -g $GID -s /bin/bash build && \
-    chown -R build:build /onie
+# Reuse the host's primary group when that GID is already taken in the
+# image.  Debian fills the low GID range with system groups (30 dip, 50
+# staff, 100 users), so insisting on creating our own group makes the image
+# unbuildable for anyone whose primary group lands there -- common with
+# central/NFS accounts, and the norm on macOS hosts (GID 20 = dialout here).
+# When the group is reused, "build" no longer exists as a group name, so
+# chown by numeric id.
+RUN if getent group "$GID" >/dev/null; then \
+        echo "GID $GID is $(getent group "$GID" | cut -d: -f1); reusing it"; \
+    else \
+        groupadd -g "$GID" build; \
+    fi && \
+    useradd -l -m -u "$UID" -g "$GID" -s /bin/bash build && \
+    chown -R "$UID:$GID" /onie
+
+# /sbin and /usr/sbin hold tools the build invokes.  Set this in two places
+# so it holds however the image is invoked:
+#   ENV        -- non-login shells (docker run <cmd>, bash -c)
+#   profile.d  -- login shells.  Debian's /etc/profile *replaces* PATH for
+#                 non-root users with one that omits both sbin dirs, and it
+#                 does so before sourcing profile.d, so the snippet wins.
+# ~/.bashrc, which this used to use, works for neither: Debian's default
+# .bashrc returns early when the shell is not interactive, so under the
+# workflow's "bash -lc" the PATH line was never reached.
+ENV PATH="/sbin:/usr/sbin:${PATH}"
+RUN printf '%s\n' 'export PATH="/sbin:/usr/sbin:$PATH"' \
+        > /etc/profile.d/00-onie-sbin.sh
 
 USER build
-
-# /sbin and /usr/sbin hold tools the build invokes.
-RUN echo 'export PATH="/sbin:/usr/sbin:$PATH"' >> ~/.bashrc
 
 # The build runs git commands; give it a default identity.
 RUN git config --global user.email "build@example.com" && \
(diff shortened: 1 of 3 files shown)


mshych (approved):
Approved

